PR #14, opened 2025-09-16
[NVIDIA] Update B200 vLLM image for 70B

kedarpotdar-nv:
Only slurm for now.


kimbochen:
[This](https://github.com/InferenceMAX/InferenceMAX/actions/runs/17774405454) is a successful run right?
lgtm

PR #15, opened 2025-09-17
[NVIDIA] Add Llama 70B FP4 using vLLM on B200

kedarpotdar-nv:
Added 70b_fp4_b200_slurm.sh and modified image in 70b-tmpl.yml. Successfully tested in CI https://github.com/InferenceMAX/InferenceMAX/actions/runs/17786819400


kimbochen:
lfggg

PR #17, opened 2025-09-17
[NVIDIA] Add GPT-OSS B200 vLLM

kedarpotdar-nv:
1. updated gptoss-tmpl.yml to add bmk-b200
2. Created b200 vllm slurm benchmark script.


kimbochen:
lgtm

PR #20, opened 2025-09-18
[NVIDIA] Update 70B TRT To use latest container and flags

kedarpotdar-nv:


kimbochen:
- Updated TRT-LLM images
- Updated server configs for B200 and H200 TRT-LLM
- Increased concurrencies for B200 and H200

PR #23, opened 2025-09-18
Update max-num-batched-tokens parameter to 131072 to improve throughp…

japarada:
…ut and ttft.
Will provide an expected performance boost. For example for the 8K1K con 4 we are seeing ~29% boost on Mi300x.
Here are the runs: before https://github.com/InferenceMAX/InferenceMAX/actions/runs/17838687561/job/50722323101#step:4:575
and after
https://github.com/InferenceMAX/InferenceMAX/actions/runs/17838980169/job/50723297124#step:4:578


kimbochen:
lgtm

PR #41, opened 2025-09-23
[AMD] amd gpt-oss mi355x update

seungrokj:
Hi @japarada  @kmbochen 

This is mi355x + gpt-oss with a proper setting. For the verbose compilation sizes + graph capture sizes, it is to cache the compiled artifacts & expand the cuda graph size (default is 1k). 
It may increase a slight compilation time overhead but we verified the overhead is minimal. 

Regards, 
Seungrok


japarada (changes requested):
please do not remove PORT variable.

kimbochen:
lgtm
@japarada removing `PORT` is fine, given it's defined and set in the docker launch script.  
That said, it will also need to be defined if we have a slurm launch script in the future.  
[Validated run](https://github.com/InferenceMAX/InferenceMAX/actions/runs/17949632205)

PR #44, opened 2025-09-23
[NVIDIA] Remove Triton Dependency in GPT-OSS H200 TRT

kedarpotdar-nv:
1. revert to older container which has triton pre-installed
2. update image in YML files


kedarpotdar-nv:
closing . need to run more tests

PR #47, opened 2025-09-24
vllm cache reuse

seungrokj:
Hi @kimbochen

Please ignore if this is not helpful. Just to speed up overall tests.

Regards, 
Seungrok


kimbochen:
Yes, ignoring because vLLM cache reuse causes errors when the number of runs increases

PR #48, opened 2025-09-24
Typo fix: vLLM 70b launch cmd had a typo. 

kedarpotdar-nv:
two changes:
1. 70b fp8 b200 docker had an = sign that didn't work for config.yaml
2. adding env var that i missed for b200 gptoss vllm


kimbochen:
What that's crazy. Can you share where you found having `=` stops config from loading?

kedarpotdar-nv:
The perf from 70b docker runner wasn't matching 70b slurm runner. Turns out the only difference is in = sign. Other vllm docker scripts don't have this . only the fp8 70b b200 vllm. 

More details on usage here: https://docs.vllm.ai/en/latest/configuration/serve_args.html#cli-arguments

kimbochen:
Yeah I added those, my bad.  
Thanks for catching that

PR #53, opened 2025-09-25
[NVIDIA] Add DSR1 MTP

kedarpotdar-nv:
As per guidance we've recvd, wanted to add MTP ON configs for single node after GB200 is merged. 
Only TRT MTP for now.

Changes:
1. Added benchmarks/dsr1_*_mtp_slurm.sh files for fp8 and fp4
2. modified dsr1-tmpl.yml to account for new runs.

### DIFF
--- a/.github/workflows/benchmark-tmpl.yml
+++ b/.github/workflows/benchmark-tmpl.yml
@@ -101,7 +101,7 @@ jobs:
       - name: Launch job script
         env:
           RUNNER_NAME: ${{ runner.name }}
-          RESULT_FILENAME: ${{ env.EXP_NAME }}_${{ env.PRECISION }}_${{ env.FRAMEWORK }}_tp${{ env.TP }}_conc${{ env.CONC }}_${{ runner.name }}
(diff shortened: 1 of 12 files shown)


kedarpotdar-nv:
I'm not sure the mtp-mode: on actually triggers the dsr1_b200_fp4/fp8_trt_mtp.sh scripts.

kedarpotdar-nv:
@kimbochen can you please guide how to add logic in runner/launch b200 to account for mtp flag?

csahithi:
Can we add a variable called `mtp-mode` (default set to off) to `benchmark-tmpl.yaml`? We can pass this as an env variable which will be used by the runner script. If set as `on`, then we attach `_mtp` as a suffix to the filename so that the correct script gets picked up.

kedarpotdar-nv:
Thanks @csahithi  . i have made changes accordingly. ran several sample sweeps:

https://github.com/InferenceMAX/InferenceMAX/actions/runs/18050118322
https://github.com/InferenceMAX/InferenceMAX/actions/runs/18050505472
https://github.com/InferenceMAX/InferenceMAX/actions/runs/18051017127 
https://github.com/InferenceMAX/InferenceMAX/actions/runs/18049715475

functionstackx:
stale

PR #69, opened 2025-09-28
Add a comment on the use of float16 and set some EVs explicitly

qcolombet:


JArnoldAMD:
Fixed up a couple of envvar settings for Llama 70B FP4 in the 8192/1024 configs

PR #75, opened 2025-09-30
improve NVIDIA CI stability

kedarpotdar-nv:
### Changes

- Disable NCCL Graph Registration

NCCL_GRAPH_REGISTER tries to automatically enable user buffer registration with CUDA Graphs. Disabling it can reduce our vLLM and SGLang perf but will improve CI stability.

- remove enroot --container-mount-home

This causes conflicts with pre-installed packages inside user directory and caused last night's runs to fail 

See successful run: https://github.com/InferenceMAX/InferenceMAX/actions/runs/18146263616


kedarpotdar-nv:
Also modified docker run to include --init to improve stability by better handling zombie processes

https://www.paolomainardi.com/posts/docker-run-init/

PR #104, opened 2025-10-14
[AMD] Use vLLM framework for DeepSeek R1 on MI325 and MI355 hardware

omirosh:
Please consider changing framework for DeepSeek R1 to vLLM, it shows better performance over SGLang. 
Here is also [documentation for running DeepSeek with vLLM](https://rocm.docs.amd.com/en/docs-7.0-docker/benchmark-docker/inference-vllm-deepseek-r1-fp8.html).


qcolombet:
@cquil11 , @functionstackx, I don't have the permission to assign a reviewer, so just tagging you both :).
I know you guys are figuring out the load on the CI before moving forward with the review.
Let me know what we can do to help.

functionstackx:
@qcolombet yes, we are looking into it

@cquil11 is just trying to land an massive refactor PR first to reduce tech debt and then we can look into this one

functionstackx:
@merrymercy

functionstackx:
@omirosh closing as this is stale PR. happy to take a look at amd deepseek vllm in addition to amd's sglang deepseek configs if single node is an community vllm image that supports fp4 and fp8

PR #106, opened 2025-10-15
[AMD] Adjust the threshold for when to have AITER MHA enabled

JArnoldAMD:
Internal tests have shown improved results with VLLM_ROCM_USE_AITER_MHA=1 for ISL=8k / OSL=1k / concurrency 16.


cquil11 (approved):
seems reasonable
thanks for the updat

PR #109, opened 2025-10-15
feat: output throughput per gpu 

functionstackx:
vllm `benchmark_serving.py` already benchmarks and collects the output throughput per sec but the middleware helper script for transforming the data didn't piping it through yet. 

added quick patch to `utils/process_result.py`

confirmed that no changes to `utils/collect_results.py` where needed

here is the validation run: https://github.com/InferenceMAX/InferenceMAX/actions/runs/18542476169/job/52853541173

## per benchmark job artifact output
image

## per workflow aggregration artifact
image

### DIFF
--- a/.github/workflows/benchmark-tmpl.yml
+++ b/.github/workflows/benchmark-tmpl.yml
@@ -137,12 +137,12 @@ jobs:
             exit 1
           fi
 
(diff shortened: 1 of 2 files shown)


kimbochen (approved):
lgtm

PR #113, opened 2025-10-17
feat: input throughput per gpu

functionstackx:
vllm `benchmark_serving.py` already benchmarks and collects the total & output throughput per sec but the middleware helper script for transforming the data didn't piping it through yet. (`input = total - output`)

added quick patch to `utils/process_result.py`

confirmed that no changes to `utils/collect_results.py` where needed

Validated changes here
https://github.com/InferenceMAX/InferenceMAX/actions/runs/18581713783


cquil11 (approved):
nice

kimbochen (approved):
lgtm

PR #145, opened 2025-10-24
feat: candidate search space (no perf changes for amd/nvidia)

cquil11:
# Candidate Search Space

> [!NOTE]
> No changes in the PR affect any performance for either AMD or NVIDIA benchmarks. It is purely a refactor.

Example Workflows:
- 1k1k Full Sweep Scheduler: https://github.com/InferenceMAX/InferenceMAX/actions/runs/18984029536
- Runner Sweep Test: https://github.com/InferenceMAX/InferenceMAX/actions/runs/18976786131
- See `.github/README.md` for more examples.

Follow up TODOs:
- #151 
- #152

## Problems with Current Approach
Currently, the InferenceMAX architecture uses what can best be described as a “trickle down” approach. More specifically, we start by separating the top-level workflows by sequence length (1k1k, 1k8k, 8k1k), which then invokes the “Full Sweep Template” which can be thought of as a function used by another workflow. Like any function, these templates specify certain variables that must be passed, specified in their “signature” or inputs. The Full Sweep Template then spawns a series of jobs conditioned on the input (i.e., the 1k1k scheduler only spawns jobs for ISL/OSL equal to 1024). The jobs spawned by the Full Sweep Template call another workflow function specific to each model (i.e., 70b Template). These workflow functions then invoke jobs across all hardware and precisions compatible with the particular model. Finally, these jobs call the bottom-level function Benchmark Template which encapsulates all of the logic for actually launching the benchmark script and scheduling it on a self-hosted runner. The “Current Architecture Diagram” entry in the Appendix A of this document gives a visual representation of this flow.

The problems with the current approach are as follows:
- Too many layers of abstraction, i.e., too many workflows, workflow functions, sub-workflows
     - Difficult to create new test workflows for specific cases, as this requires a separate job and/or workflow for each use case
     - Hard for contributors to make contributions – especially for the first time
- Vanilla GitHub Actions does not provide enough native support for the expressivity we want from configs. Put more simply, YAML + GHA primitives is not a programming language.

Let us elaborate on point 2 above. Our workflows rely on GitHub Action’s [matrix strategy](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations) to generate jobs for all parallelism and concurrency levels in the cartesian product tp-list x conc-list. For instance,

```yaml
 bmk-b200-trt-fp8:
   if: ${{ inputs.use_b200 }}
   uses: ./.github/workflows/benchmark-tmpl.yml
   with:
      runner: h200-trt
      image: 'nvcr.io#nvidia/tensorrt-llm/release:1.1.0rc2.post2'
      model: 'deepseek-ai/DeepSeek-R1-0528'
      framework: 'trt'
      precision: 'fp8'
      exp-name: ${{ inputs.exp-name }}
      isl: ${{ inputs.isl }}
      osl: ${{ inputs.osl }}
      max-model-len: ${{ inputs.max-model-len }}
      random-range-ratio: ${{ inputs.random-range-ratio }}
      conc-list: '[1, 2, 4]'
      tp-list: '[4, 8]'
```
would generate 6 jobs: tp4 conc1, tp4 conc2, tp4 conc4, tp8 conc1, tp8 conc2, tp8 conc4. There are some limitations with this approach:
1. There is no way to specify a different concurrency range for one level of parallelism and a different range for another. This is a problem, because first principles tells us that certain tp/conc configurations are not realistic for performance and will not fall on the Pareto frontier.
image
2. With the current approach, all points are swept across (even the ones that will not fall on the Pareto frontier) leading to wasted CI time and more importantly, wasted compute.
There is not a way to specify different parallelism techniques such as expert parallelism or DP attention. All of the logic for this is handled at the benchmark script level which is not ideal – it is preferable that this logic is lifted up out of Bash scripts so that these options can be more easily specified by users. Below is an example of how things like EP and DP attention are currently specified:

```bash
EP_SIZE="$TP"
MOE_BACKEND="CUTLASS"
DP_ATTENTION=false

if [[ "$ISL" == "1024" && "$OSL" == "1024" ]]; then
   if [[ $CONC -gt 64 ]]; then
       DP_ATTENTION=true
   fi
elif [[ "$ISL" == "1024" && "$OSL" == "8192" ]]; then
   if [[ $CONC -gt 64 ]]; then
       DP_ATTENTION=true
   fi
elif [[ "$ISL" == "8192" && "$OSL" == "1024" ]]; then
   if [[ $CONC -gt 32 ]]; then
       DP_ATTENTION=true
   fi
fi
```

## Proposed Solution
To reconcile all aforementioned issues, we propose a simpler design eliminating many of the layers of workflow functions by specifying all benchmark configurations in an external master configuration JSON, processing it at the top level workflow, and creating a flattened matrix to launch all of the jobs. This will effectively take away the need for the Full Sweep Template as well as the Model Templates. The diagram in Appendix B can be referenced to gain a high-level understanding of the proposed architecture.

In other words, all possible benchmark configurations are defined in one place, which will be considered the primary “source of truth.” Further, the logic for deciding which jobs run will be completely self-contained in a Python script.

### Workflow Files
The top-level schedulers will continue to be split up by sequence length – recall this is due to the limitation of approximately 500 jobs per workflow on GitHub actions, otherwise the UI fails to load and times out after 10 seconds. Furthermore, each scheduler workflow is split up by model, again due to a GitHub Actions limitation where [no more than 256 jobs may be generated via a matrix](https://docs.github.com/en/actions/reference/limits#:~:text=256%20jobs%20/%20workflow%20run). Each model has two jobs associated with it: one to retrieve the appropriate configs from the master config and dump the output JSON to an environment variable and another to consume that JSON to generate a matrix of jobs, for each parallelism-concurrency combination.

Consider the simplified example below:

```yaml
name: '1K/1K Sweep'

jobs:
  get-gptoss-configs:
    runs-on: ubuntu-latest
    outputs:
        search-space-config: ${{ steps.get-gptoss-configs.outputs.search-space-config }}
    steps:
      - name: Checkout code
        uses: actions/checkout@v4

      - id: get-gptoss-configs
        run: |
          CONFIG_JSON=$(python3 ${GITHUB_WORKSPACE}/utils/get_configs.py ${GITHUB_WORKSPACE}/.github/configs/master.yaml 1k1k gptoss)
          echo "search-space-config=$CONFIG_JSON" >> $GITHUB_OUTPUT

  benchmark-gptoss:
    needs: get-gptoss-configs
    uses: ./.github/workflows/benchmark-tmpl.yml
    name: gptoss 1k1k
    strategy:
      fail-fast: false
      matrix:
        config: ${{ fromJson(needs.get-gptoss-configs.outputs.search-space-config) }}
    secrets: inherit
    with:
      exp-name: "gptoss_1k1k"
      isl: 1024
      osl: 1024
      max-model-len: 2048
      runner: ${{ matrix.config.runner }}
      image: ${{ matrix.config.image }}
      model: ${{ matrix.config.model }}
      framework: ${{ matrix.config.framework }}
      precision: ${{ matrix.config.precision }}
      tp: ${{ matrix.config.tp }}
      conc: ${{ matrix.config.conc }}
```

The `get-gptoss-configs `job calls the `utils/get_configs.py` script with the appropriate input (the config file, sequence length, and model for which to receive configs). The `get_configs.py` script iterates through all configs in the master configuration and processes them such that they can be loaded via GitHub Action’s fromJson function. Upon success, a matrix of jobs is created for each model.

Screenshot 2025-10-31 at 11 28 31 AM

### Master Config Structure
The master config YAML is parsed by the `utils/get_configs.py` script and as such, must adhere to a strict structure. The structure for an entry is laid out below:

```yaml
<INFMAX_MODEL_NAME>-<PRECISION>-<RUNNER>:
  image: 
  model: 
  runner:
  precision:
  framework:
  seq-len-configs:
  - isl: 1024
    osl: 1024
    bmk-space:
    - { tp: 1, conc-start:, conc-end: }
    - { tp: 2, conc-start:, conc-end: }
	...
  - isl: 1024
    osl: 8192
    bmk-space:
	...
  - isl: 8192
    osl: 1024
    bmk-space:
    	...
```

Below is a concrete example:

```yaml
dsr1-fp8-h200-trt:
  image: nvcr.io#nvidia/tensorrt-llm/release:1.1.0rc2.post2
  model: deepseek-ai/DeepSeek-R1-0528
  runner: h200-trt
  precision: fp8
  framework: trt
  # For all sequence lengths, EP=TP
  seq-len-configs:
  - isl: 1024
    osl: 1024
    # If CONC > 64, then DP_ATTN=true
    bmk-space:
    - { tp: 8, ep: 8, conc-start: 4, conc-end: 64 }
  - isl: 1024
    osl: 8192
    # If CONC > 64, then DP_ATTN=true
    bmk-space:
    - { tp: 8, ep: 8, conc-start: 4, conc-end: 64 }
  - isl: 8192
    osl: 1024
    # If CONC > 32, then DP_ATTN=true
    bmk-space:
    - { tp: 8, ep: 8, conc-start: 4, conc-end: 32 }
    - { tp: 8, ep: 8, dp-attn: true, conc-start: 64, conc-end: 64 }
```

In order to keep the pipelines reproducible and reliable, the structure of the config as proposed above must adhere exactly to the format above. This will be both implicitly and explicitly enforced within the workflows.

### Test Workflows
Many of the existing workflows used for testing such as `runner-model-sweep-test.yml`, `runner-sweep-test.yml`, `runner-test.yml`, and more are useful and must be ported/refactored as part of these proposed changes. 

image

The above diagram depicts the general flow of any workflow using this proposed architecture. 
### Single Python Script
At a high level, the Python script (`utils/matrix-logic/generate_sweep_configs.py`) referenced in the above diagram works as follows:
The user enters a series of pre-defined inputs/arguments describing what configs they would like to run that come from the “source of truth” master configs.
These inputs trigger an appropriate function that invokes logic to fetch a list of all appropriate benchmarks to run, each represented as an individual JSON entry.
This final JSON object with all jobs to run is dumped to stdout such that the JSON can be loaded as using the GHA fromJson command and used to generate a matrix of jobs.

Here is the usage message for the script:

Screenshot 2025-10-31 at 11 12 10 AM

Let’s run through some of the existing test workflows and show we can get the equivalent functionality from this script:
#### `runner-test.yml`
Meant to: test 1 model on 1 runner node.
```bash
python3 utils/matrix-logic/generate_sweep_configs.py custom --config-files .github/configs/amd-master.yaml .github/configs/nvidia-master.yaml --runner-config  .github/configs/runners.yaml --runner-label h200-nv_1 --framework vllm --precision fp8 --exp-name 70b_test --image vllm/vllm-openai:v0.10.2 --model deepseek-ai/DeepSeek-R1-0528
```
Output:
Screenshot 2025-10-31 at 11 42 52 AM

#### `runner-model-sweep.yml`
Meant to:  test multiple models on multiple runner nodes/
```bash
python3 utils/matrix-logic/generate_sweep_configs.py runner-model-sweep --config-files .github/configs/nvidia-master.yaml .github/configs/amd-master.yaml --runner-config .github/configs/runners.yaml --runner-type h100
```
Output:
Screenshot 2025-10-31 at 11 44 01 AM

Notice identical output, and now even better since no one has to maintain all of the redundant values in runner-model-sweep.yml workflow – now they are all in one source of truth config.

#### `runner-sweep-test.yml`
Meant to: test 1 model on multiple runner nodes
- Never actually used (not even once) in the current repo.
- Slightly confusing for testers, since users are prompted to choose a runner type, Docker image, model, framework, and precision – it is hard for newer users / contributors to know what combinations of these inputs is actually meant to be run.
- This refactor eliminates this challenge to an extent. Instead, the user just inputs the model they want to sweep across runners, and the script finds all configs from the master configs that include the specified model.
- Also an option to filter by framework and precision. In the case that the user inputs an incorrect combination, the script fails immediately instead of failing at the benchmark script level, which takes longer to fail.

Screenshot 2025-10-31 at 11 47 06 AM

```bash
python3 utils/matrix-logic/generate_sweep_configs.py runner-sweep --config-files .github/configs/amd-master.yaml .github/configs/nvidia-master.yaml --runner-config .github/configs/runners.yaml  --model-prefix gptoss --runner-type h200
```	
In English: “Test all h200 runner nodes that are compatible with a gptoss configuration as described in the master configs.”

```bash
python3 utils/matrix-logic/generate_sweep_configs.py runner-sweep --config-files .github/configs/amd-master.yaml .github/configs/nvidia-master.yaml --runner-config .github/configs/runners.yaml  --model-prefix dsr1 --precision fp8 --runner-type b200
```
In English: “Test all b200 runner nodes that are compatible with a dsr1 fp8 configuration as described in the master configs.”

#### `full-sweep-test.yml`
Meant to: test full sweep functionality.
- Currently, user inputs boolean selections to run some combination of runners and ISL/OSL.
- To replicate this, `full-sweep-test.yml` is edited to use the Python script to run the selected configurations.
   - Quick and easy way to test, but still requires the “full” sweeps to run.
   - 
#### New: `e2e-tests.yml`
Provides even more control over what tests to run, as the user directly inputs the Python args to the workflow which allows maximum control over testing.
**Tradeoff**: doesn’t replicate a full sweep exactly i.e., no call to `collect.results.yml`, no integration for GB200 (yet) – however, this provides a great testing utility to test all cases quickly.
Examples:
```bash
python3 utils/matrix-logic/generate_sweep_configs.py full-sweep --model-prefix gptoss --config-files .github/configs/nvidia-master.yaml .github/configs/amd-master.yaml --runner-config .github/configs/runners.yaml | jq length
362
```
In English: “Run all gptoss configurations.”
```bash
python3 utils/matrix-logic/generate_sweep_configs.py full-sweep --model-prefix gptoss --runner-type b200 --config-files .github/configs/nvidia-master.yaml .github/configs/amd-master.yaml --runner-config .github/configs/runners.yaml | jq length
46
```
In English: “Run all gptoss configurations that run on B200.”

```bash
 python3 utils/matrix-logic/generate_sweep_configs.py full-sweep --model-prefix dsr1 --runner-type b200 mi300x --config-files .github/configs/nvidia-master.yaml .github/configs/amd-master.yaml --runner-config .github/configs/runners.yaml | jq length
63
```
In English: “Run all dsr1 configurations that run on b200 or mi300x.”

```bash
python3 utils/matrix-logic/generate_sweep_configs.py full-sweep --model-prefix dsr1 --runner-type b200 mi300x --precision fp8 --config-files .github/configs/nvidia-master.yaml .github/configs/amd-master.yaml --runner-config .github/configs/runners.yaml | jq length
30
```
In English: “Run all dsr1 configurations that run on b200 or mi300x and run fp4”

```bash
python3 utils/matrix-logic/generate_sweep_configs.py full-sweep --model-prefix dsr1 --runner-type b200 mi300x --precision fp8 --test-mode --config-files .github/configs/nvidia-master.yaml .github/configs/amd-master.yaml --runner-config .github/configs/runners.yaml | jq length
6
```
In English: “Run all dsr1 configurations that run on b200 or mi300x and run fp4, but only reduce the parallelism-concurrency search space to a single run (picks the highest parallelism and lowest concurrency levels available for that configuration)."

### Frontend Considerations
The proposed changes do not break anything on the frontend critical-path. In particular, the artifacts that get generated from the full sweeps have the same names, locations as artifacts, etc. Do note that a discovery of this PR is that the frontend critically depends on the name of the full sweep schedulers, as this is what is used to hash the runs and collect their information.

As a follow-up, artifacts now have the additional fields of ep and dp-attn, so should be added upon hover of data points accordingly.

### Tradeoffs
There are clear tradeoffs between the proposed architecture and the existing one. For one, while the proposed approach has far fewer levels of workflows (i.e., no more model templates, less workflow “functions”), it has more indirection. That is, most logic is self-contained within the workflow files themselves. This is good for reducing the room for error, but it means having to create a separate workflow file for each new type of functionality.

#### GB200 Integration
Currently, the GB200 multinode benchmarks run using a separate workflow call template than all other benchmarks (benchmark-multinode-tmpl.yaml). This causes some issues when trying to directly integrate with the proposed architecture. There are some ideas on how to integrate it to the master configuration, but for now the workaround is to just add the GB200 runs as a separate job to the XkYk-sweep.yml workflow files. See the example below from the 1k1k-sweep.yml file:
```
   # This is a workaround until we can integrate GB200 into master configs.
    benchmark-gb200:
        uses: ./.github/workflows/benchmark-multinode-tmpl.yml
        name: gb200 1k1k sweep
        strategy:
            fail-fast: false
            matrix:
                config:
                    - {
                          "image": "nvcr.io#nvidia/ai-dynamo/tensorrtllm-runtime:0.5.1-rc0.pre3",
                          "model": "deepseek-r1-fp4",
                          "model-prefix": "dsr1",
                          "precision": "fp4",
                          "framework": "dynamo-trtllm",
                          "mtp": "off",
                      }
                    - {
                          "image": "nvcr.io#nvidia/ai-dynamo/tensorrtllm-runtime:0.5.1-rc0.pre3",
                          "model": "deepseek-r1-fp4",
                          "model-prefix": "dsr1",
                          "precision": "fp4",
                          "framework": "dynamo-trtllm",
                          "mtp": "on",
                      }
                    - {
                          "image": "nvcr.io/nvidia/ai-dynamo/sglang-runtime:0.5.1-rc0.pre1",
                          "model": "deepseek-ai/DeepSeek-R1-0528",
                          "model-prefix": "dsr1",
                          "precision": "fp8",
                          "framework": "dynamo-sglang",
                          "mtp": "off",
                      }
        secrets: inherit
        with:
            runner: gb200
            image: ${{ matrix.config.image }}
            model: ${{ matrix.config.model }}
            framework: ${{ matrix.config.framework }}
            precision: ${{ matrix.config.precision }}
            exp-name: ${{ matrix.config.model-prefix }}_1k1k
            isl: 1024
            osl: 1024
            max-model-len: 2048
            mtp-mode: ${{ matrix.config.mtp }}
```

The plan is to integrate GB200 to the master configs at a later date. For now, this tradeoff is well worth it as it comes with faster development time allowing InferenceMAX to continue moving at the speed of light.

As far as tests for GB200 goes, the same logic as above applies. This refactor will provide a separate gb200-tests.yml workflow file that will allow the user to manually test GB200 nodes. Note that even with the current state of the repo, before refactor, there is very little support for testing GB200.

#### One Giant Python File
This refactor proposes one giant “god” Python file that encapsulates all logic for what benchmarks should run. This adds a new level of indirection and will make it more difficult to add new functionality. However, this is necessary to move away from in-workflow configurations to master configurations. This is also necessary to run a subset of concurrencies for each parallelism level, instead of running the entire Cartesian product of TP x CONC – this will allow CI time to be reduced by roughly 20%, which will allow more time for running more interesting, multinode tests.

Further, with this new level of indirection, developers may choose now to use it when creating their own testing workflows. Despite being used to generate the full sweeps, much of this script is intended to be used for input to e2e-tests.yml (testing) – if developers decide they would rather write their own logic/workflow files, they are free to do so.

if developers not like this god file for testing, we will go back to drawing board to refactor again,

it is hard for developers that wanna add 1 simple functionality to now add to understand an 900 line god file

Finally, one giant Python script encapsulating all of the logic used to generate tests will naturally be more error prone. In order to fend against bugs, any input from the config files are validated before being sent to the script, and all output is validated after being generated by the script and before being sent to benchmark-tmpl.yml.

#### Removed Graph Plotting
This PR removes graph plotting to reduce the amount of code that needs to be maintained and increase the speed of the CI pipeline. While this will increase development speed, it will prevent the ability to see different plotted results from different days as our frontend currently does not show historical comparison.

### Future considerations
For a while now, we have been thinking about ways to move away from a simple nightly cron job trigger to a code diff based trigger. I.e., only run CI pipelines for configs/files that have changed in a commit. This PR helps move towards this goal as there is now a single stateful representation of all configurations that could be run. Now it is just a matter of deciding what changes affect what configurations. 

Additionally, we are working towards pre-merge CI validation (based on labels, or something similar). The Python script will also help enable this, as it gives fine control over what tests should be run.

In order to further enforce against bugginess in the Python script, we intend to add an additional CI workflow that will be run upon any diff to the master configs to validate their structure before pushing to main.
### A Current Architecture Diagram
image

### B Proposed Architecture Diagram
image


.github/workflows/benchmark-tmpl.yml
```diff
@@ -127,7 +129,7 @@ jobs:
       - name: Launch job script
         env:
           RUNNER_NAME: ${{ runner.name }}
-          RESULT_FILENAME: ${{ env.EXP_NAME }}_${{ env.PRECISION }}_${{ env.FRAMEWORK }}_tp${{ env.TP }}_conc${{ env.CONC }}_${{ runner.name }}
```
functionstackx, line 130: can u hash on dpa too?

.github/workflows/benchmark-tmpl.yml
```diff
@@ -51,23 +60,16 @@ env:
...
-      matrix:
-        tp: ${{ fromJson(inputs.tp-list) }}
-        conc: ${{ fromJson(inputs.conc-list) }}
-    name: '${{ inputs.exp-name }} ${{ inputs.runner }} ${{ inputs.precision }} tp${{ matrix.tp }} conc${{ matrix.conc }}'
-
-    env:
-      TP: ${{ matrix.tp }}
-      CONC: ${{ matrix.conc }}
-
+    name: '${{ inputs.exp-name }} ${{ inputs.runner }} ${{ inputs.precision }} tp${{ inputs.tp }} ep${{ inputs.ep }} conc${{ inputs.conc }}'
```
functionstackx: hash on dpa too

utils/process_result.py
kimbochen, line 1: Can you change to reading inputs from env var, whenever possible?  
I think it's more descriptive and less error-prone.
e.g.
```python
import os
tp_size = os.environ['TP']
framework = osl.environ['FRAMEWORK']
...
```

.github/configs/amd-master.yaml
kimbochen, line 1: Change `bmk-space` to `search-space`

.github/configs/nvidia-master.yaml
kimbochen, line 1: Change `bmk-space` to `search-space`

utils/process_result.py
cquil11: i started doing this and then reverted everything because I realized how much shit is already in this PR

we can put a pin in this (also, it will affect the multi-node workflow)

utils/summarize.py
```diff
@@ -25,7 +25,9 @@
         f"| {framework.upper()} "
         f"| {precision.upper()} "
         f"| {result['tp']} "
+        f"| {result['ep']} "
         f"| {result['conc']} "
+        f"| {result['dp_attention']} "
```
functionstackx: nit: list tp, ep, dpa and then conc

doing tp, ep, conc and then dpa doesnt make sense

.github/workflows/1k8k-sweep.yml
```diff
@@ -0,0 +1,147 @@
...
+            search-space-config: ${{ steps.get-70b-configs.outputs.search-space-config }}
+        steps:
+            - name: Checkout code
+              uses: actions/checkout@v4
+
+            - id: get-70b-configs
+              run: |
+                  CONFIG_JSON=$(python3 ${GITHUB_WORKSPACE}/utils/matrix-logic/get_full_sweep_configs.py --config-files ${GITHUB_WORKSPACE}/.github/configs/nvidia-master.yaml ${GITHUB_WORKSPACE}/.github/configs/amd-master.yaml --seq-lens 1k8k --model-prefix 70b)
+                  echo "search-space-config=$CONFIG_JSON" >> $GITHUB_OUTPUT
+
```
functionstackx: since https://github.com/InferenceMAX/InferenceMAX/pull/149 , no need for llama 70b anymore

.github/configs/amd-master.yaml
```diff
@@ -0,0 +1,280 @@
...
+    - { tp: 2, conc-start: 32, conc-end: 64 }
+    - { tp: 4, conc-start: 4, conc-end: 64 }
+    - { tp: 8, conc-start: 4, conc-end: 64 }
+  - isl: 8192
+    osl: 1024
+    search-space:
+    - { tp: 1, conc-start: 32, conc-end: 64 }
+    - { tp: 2, conc-start: 32, conc-end: 64 }
+    - { tp: 4, conc-start: 4, conc-end: 64 }
+    - { tp: 8, conc-start: 4, conc-end: 64 }
```
functionstackx: since https://github.com/InferenceMAX/InferenceMAX/pull/149 , no need for llama 70b anymore

utils/process_result.py
```diff
@@ -5,31 +5,30 @@
...
-tput_per_gpu = float(bmk_result['total_token_throughput']) / tp_size
-output_tput_per_gpu = float(bmk_result['output_throughput']) / tp_size
-input_tput_per_gpu = tput_per_gpu - output_tput_per_gpu
-
 data = {
     'hw': hw,
     'tp': tp_size,
+    'ep': ep_size,
     'conc': int(bmk_result['max_concurrency']),
+    'dp_attention': dp_attention, # true or false
```
functionstackx, line 23: dp attention first before conc

utils/matrix-logic/get_test_sweep_configs.py
```diff
@@ -0,0 +1,151 @@
...
+                    'tp': tp,
+                    'conc': conc,
+                    'max-model-len': isl + osl,
+                }
+                
+                # Add optional fields if they exist
+                if ep is not None:
+                    entry['ep'] = ep
+                if dp_attn is not None:
+                    entry['dp-attn'] = dp_attn
```
functionstackx, line 137: shouldnt these be set to 1 and false if they dont exist

cquil11: this file shold not exist it got readded during rebase

.github/workflows/collect-results.yml
```diff
@@ -35,16 +35,3 @@ jobs:
...
-          pip install -q matplotlib
-          python3 utils/plot_perf.py results/ ${{ inputs.exp-name }}
-
-      - name: Upload performance graphs
-        uses: actions/upload-artifact@v4
-        with:
-          name: graphs_${{ inputs.exp-name }}
-          path: |
-            tput_vs_intvty_*_${{ inputs.exp-name }}.png
-            tput_vs_e2el_*_${{ inputs.exp-name }}.png
```
functionstackx, line 50: why r u deleting this?

utils/plot_perf.py
```diff
@@ -1,197 +0,0 @@
-import sys
-import json
-from pathlib import Path
-import matplotlib.pyplot as plt
-
-
```
functionstackx, line 6: why r u regressing functionality

.github/workflows/1k8k-sweep.yml
```diff
@@ -0,0 +1,147 @@
...
+            framework: ${{ matrix.config.framework }}
+            precision: ${{ matrix.config.precision }}
+            exp-name: ${{ matrix.config.model-prefix }}_1k8k
+            isl: 1024
+            osl: 8192
+            max-model-len: 2048
+            mtp-mode: ${{ matrix.config.mtp }}
+
+    collect-dsr1-results:
+        needs: benchmark-dsr1
```
functionstackx: can u fix this, doesnt colect dsr1 need to be dependent on `benchmark-gb200` too?

utils/matrix-logic/generate_sweep_configs.py
```diff
@@ -0,0 +1,927 @@
...
+                    'conc': conc,
+                    'max-model-len': isl + osl + 200,
+                    'exp-name': f"{model_code}_{seq_len_str}",
+                }
+
+                if ep is not None:
+                    entry['ep'] = ep
+                if dp_attn is not None:
+                    entry['dp-attn'] = dp_attn
+
```
functionstackx, line 274: shoulnt this be set to 1 and false if is None

utils/matrix-logic/generate_sweep_configs.py
```diff
@@ -0,0 +1,927 @@
...
+                            'ep': 1,  # Default
+                            'dp-attn': False,  # Default
+                            'exp-name': f"{model_code}_{seq_len_str}",
+                        }
+
+                        if ep is not None:
+                            entry['ep'] = ep
+                        if dp_attn is not None:
+                            entry['dp-attn'] = dp_attn
+
```
functionstackx, line 308: shoulnt this be set to 1 and false if is None

utils/matrix-logic/generate_sweep_configs.py
```diff
@@ -0,0 +1,927 @@
...
+                'conc': lowest_conc,
+                'exp-name': f"{model_code}_test",
+                'max-model-len': 2048,
+            }
+
+            # Add optional fields if they exist
+            if ep is not None:
+                entry['ep'] = ep
+            if dp_attn is not None:
+                entry['dp-attn'] = dp_attn
```
functionstackx: shoulnt this be set to 1 and false if is None

utils/matrix-logic/generate_sweep_configs.py
cquil11: it is set by default before

utils/matrix-logic/generate_sweep_configs.py
cquil11: again, default values set before and overridden if specified

.github/workflows/1k8k-sweep.yml
cquil11: yes

utils/process_result.py
kimbochen, line 1: Get as many inputs from env vars please

.github/workflows/benchmark-tmpl.yml
```diff
@@ -139,10 +139,9 @@ jobs:
 
       - name: Process result
         run: |
-          python3 utils/process_result.py ${{ inputs.runner }} $TP $RESULT_FILENAME $FRAMEWORK $PRECISION
-
+          python3 utils/process_result.py ${{ inputs.runner }} $TP $EP_SIZE $DP_ATTENTION $RESULT_FILENAME $FRAMEWORK $PRECISION
```
kimbochen: Get as many inputs from env vars please

.github/workflows/benchmark-tmpl.yml
```diff
@@ -26,18 +26,25 @@ on:
...
         type: string
-      conc-list:
+      conc:
+        required: true
+        type: string
+      random-range-ratio:
+        required: false
         type: string
-        default: '[4, 8, 16, 32, 64]'
+        default: '0.2'
```
kimbochen: Should be 0.8 for now, my bad

utils/plot_perf.py
kimbochen, line 1: Can you keep the file please

.github/workflows/benchmark-multinode-tmpl.yml
kimbochen, line 1: Can you set MTP mode as an env var as well

.github/workflows/benchmark-tmpl.yml
```diff
@@ -127,7 +127,7 @@ jobs:
       - name: Launch job script
         env:
           RUNNER_NAME: ${{ runner.name }}
-          RESULT_FILENAME: ${{ env.EXP_NAME }}_${{ env.PRECISION }}_${{ env.FRAMEWORK }}_tp${{ env.TP }}_conc${{ env.CONC }}_${{ runner.name }}
+          RESULT_FILENAME: ${{ env.EXP_NAME }}_${{ env.PRECISION }}_${{ env.FRAMEWORK }}_tp${{ env.TP }}_ep${{ env.EP_SIZE }}_conc${{ env.CONC }}_${{ runner.name }}
```
kimbochen: Can you include the dp-attn please, if it doesn't break frontend logic
if `dp-attn=true`, have the string `dpa` in the name
otherwise, remain the same

.github/workflows/1k1k-sweep.yml
```diff
@@ -0,0 +1,148 @@
+name: "Full Sweep Scheduler - 1k1k"
+
+concurrency:
+    group: benchmark-lock-1k1k
+    cancel-in-progress: false
```
kimbochen: Concurrency lock seems unnecessary now, please remove

.github/workflows/1k8k-sweep.yml
```diff
@@ -0,0 +1,146 @@
+name: "Full Sweep Scheduler - 1k8k"
+
+concurrency:
+    group: benchmark-lock-1k8k
+    cancel-in-progress: false
```
kimbochen: Concurrency lock seems unnecessary now, please remove

utils/plot_perf.py
cquil11: So keep the file but delete it from the workflow?

.github/workflows/8k1k-sweep.yml
```diff
@@ -0,0 +1,146 @@
+name: "Full Sweep Scheduler - 8k1k"
+
+concurrency:
+    group: benchmark-lock-8k1k
+    cancel-in-progress: false
```
kimbochen: Concurrency lock seems unnecessary now, please remove

.github/workflows/full-sweep-test.yml
```diff
@@ -1,89 +1,443 @@
 name: Test - Full Sweep
 
 concurrency:
-  group: benchmark-lock
-  cancel-in-progress: false
+    group: benchmark-lock
+    cancel-in-progress: false
```
kimbochen: Concurrency lock seems unnecessary now, please remove

.github/workflows/1k1k-sweep.yml
cquil11: why is it not necessary now?

functionstackx:
@cquil11 add doc `.md` file about how to use the new testing thing

.github/workflows/collect-results.yml
cquil11: On second though, I am going to add back

PR #163, opened 2025-11-04
Fix ISL/OSL for 1k8k and 8k1k sweeps

ankursingh-nv:
This PR, updates workflow files for 1k8k and 8k1k full sweep configurations to use correct ISL & OSL values


cquil11 (commented):
lgtm

cquil11:
thank you for catching this

PR #216, opened 2025-11-10
chore: update dependabot to use grouping for actions updates

cquil11:
Instead of opening a PR for each individual instance of a change, just combine all actions updates into one PR.


kimbochen (approved):
lgtm

PR #230, opened 2025-11-13
chore: add framework to benchmark job name

cquil11:
Adds framework to each individual benchmark job name for increased clarity when viewing jobs in GitHub Actions UI.


functionstackx (approved):
lgtm

PR #242, opened 2025-11-17
[NVIDIA] Set EP_SIZE = 1 for B200 measurements

kaixih:
This PR disable the ep for B200 measurements.

### DIFF
--- a/.github/configs/nvidia-master.yaml
+++ b/.github/configs/nvidia-master.yaml
@@ -9,18 +9,18 @@ dsr1-fp4-b200-sglang:
   - isl: 1024
     osl: 1024
     search-space:
-    - { tp: 4, ep: 1, conc-start: 4, conc-end: 128 }
-    - { tp: 8, ep: 1, conc-start: 4, conc-end: 128 }
+    - { tp: 4, ep: 4, conc-start: 4, conc-end: 128 }
+    - { tp: 8, ep: 8, conc-start: 4, conc-end: 128 }
   - isl: 1024
     osl: 8192
     search-space:
-    - { tp: 4, ep: 1, conc-start: 4, conc-end: 128 }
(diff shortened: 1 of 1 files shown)


cquil11:
does this lead to perf improvement?

kaixih:
> does this lead to perf improvement?

I think a previous commit enables the EP accidentally. This PR tries to recover it.

functionstackx:
@kaixih wasn't EP always enabled?

before PR https://github.com/InferenceMAX/InferenceMAX/pull/204 , `--enable-ep-moe` & ` --enable-flashinfer-trtllm-moe` was set

according to [SGLang docs](https://github.com/sgl-project/sglang/blob/885cfca2732dd2fac9e4f76c4eacdcfafd4a5eee/docs/advanced_features/server_arguments.md?plain=1#L446), since  `--enable-ep-moe` flag has been removed, the equivalent is to set it as `--ep-size` "Please set `--ep-size` to the same value as `--tp-size` instead"

image

kaixih:
@functionstackx Right, I re-enabled the EP for the fp4 workloads. We noticed that the EP was off only for FP8 and was accidentally enabled from this https://github.com/InferenceMAX/InferenceMAX/commit/d8fe8f73eb798a7b597dfc4dbb426ceb2b6ee294.

functionstackx:
> @functionstackx Right, I re-enabled the EP for the fp4 workloads. We noticed that the EP was off only for FP8 and was accidentally enabled from this [d8fe8f7](https://github.com/InferenceMAX/InferenceMAX/commit/d8fe8f73eb798a7b597dfc4dbb426ceb2b6ee294).

@kaixih ah gotcha, i see. it was only fp4 that previously had `--enable-ep-moe`  and fp8 didn't have `--enable-ep-moe`

I see now

image

functionstackx (approved):
lgtm

PR #245, opened 2025-11-17
[NVIDIA] Add b200 DGXC node to b200 runners list

ankursingh-nv:
Adds DGXC node to the b200 runners list. Also, includes `launch_b200-dgxc.sh` script (same as `launch_b200-nvd.sh`).


.github/configs/nvidia-master.yaml
```diff
@@ -201,7 +201,7 @@ gptoss-fp4-b200-vllm:
   image: vllm/vllm-openai:v0.11.0
   model: openai/gpt-oss-120b
   model-prefix: gptoss
-  runner: b200
+  runner: b200-test
```
kedarpotdar-nv: I don't think you need to change this. The runners will be tagged 'b200' when validated, so we can continue using the b200 tag.

ankursingh-nv: Wanted to keep it separate for the time being to validate the runner on second node. This way, the workflow and target the right node for the validation run.

kedarpotdar-nv: Only change to b200-test for gptoss and keep b200 for SGL DSR1? That way we can load balance and burn in new runners.

runners/launch_b200-dgxc.sh
```diff
@@ -0,0 +1,94 @@
...
+-lc "pip install -q datasets pandas && \
+python3 bench_serving/benchmark_serving.py \
+--model $MODEL  --backend vllm --base-url http://localhost:$PORT \
+--dataset-name random \
+--random-input-len $ISL --random-output-len $OSL --random-range-ratio $RANDOM_RANGE_RATIO \
+--num-prompts $NUM_PROMPTS \
+--max-concurrency $CONC \
+--request-rate inf --ignore-eos \
+--save-result --percentile-metrics 'ttft,tpot,itl,e2el' \
+--result-dir /workspace/ --result-filename $RESULT_FILENAME.json"
```
functionstackx: benchmark script for docker is now inside the OG container instead of seaprarate container

https://github.com/InferenceMAX/InferenceMAX/pull/227

+viz @cquil11 can help answer if u have any questions about it

functionstackx (changes requested):
after #227 merged from @cquil11 , need a bit of changes to benchmark client and prob an rebase

https://github.com/InferenceMAX/InferenceMAX/pull/227

cquil11:
> after #227 merged , need a bit of changes to benchmark client and prob an rebase
> 
Yep @ankursingh-nv this will need a rebase + some manual changes
Should be pretty straightforward, let me know if you need any help

ankursingh-nv:
Refactored both the launch and benchmark scripts to align with the updates introduced in #227

kedarpotdar-nv:
please run a test sweep with the refactor

functionstackx:
@cquil11 can u help here^

benchmarks/gptoss_fp4_b200_trt_docker.sh
```diff
@@ -0,0 +1,102 @@
...
+# RESULT_FILENAME
+# PORT
+
+SERVER_LOG=$(mktemp /tmp/server-XXXXXX.log)
+
+# GPTOSS TRTLLM Deployment Guide:
+# https://github.com/NVIDIA/TensorRT-LLM/blob/main/docs/source/deployment-guide/quick-start-recipe-for-gpt-oss-on-trtllm.md
+
+# ========= Determine DP_ATTENTION, EP_SIZE and MOE_BACKEND based on ISL, OSL, CONC =========
+EP_SIZE="1"
```
cquil11: instead of defining here
EP_SIZE and DP_ATTENTION should come from the master config file (they get parsed and passed into this script via the `benchmark-tmpl.yml` workflow file

benchmarks/gptoss_fp4_b200_trt_docker.sh
```diff
@@ -0,0 +1,102 @@
...
+MOE_BACKEND="TRTLLM"
+DP_ATTENTION=false
+
+# Higher concurrencies: Concurrency >= 256
+#   MoE Backend = CUTLASS
+#   Use DP attention with expert parallel MoE
+if [[ $CONC -ge 256 ]]; then
```
cquil11: same here
we are trying to eliminate as much conditional logic as possible in these low-level bash scripts and instead have a stateful representation in master configs
so here u should just set DP_ATTNETION true and EP = TP whenever conc ge 256 _inside the master config_
make sure to leave a comment in the YAML describing why ie comment 
```
set EP = TP and DPA true whenever CONC ge 256
```

.github/configs/runners.yaml
```diff
@@ -24,6 +24,9 @@ b200:
 - 'b200-nvd_1'
 - 'b200-nvd_2'
 - 'b200-nvd_3'
+b200-test:
```
cquil11: this should be changed to just b200 right? or r u waiting to do this til you test?

benchmarks/gptoss_fp4_b200_trt_docker.sh
```diff
@@ -0,0 +1,102 @@
...
+pip install -q datasets pandas
+
+run_benchmark_serving \
+    --model "$MODEL" \
+    --port "$PORT" \
+    --backend openai \
+    --input-len "$ISL" \
+    --output-len "$OSL" \
+    --random-range-ratio "$RANDOM_RANGE_RATIO" \
+    --num-prompts $(( $CONC * 10 )) \
```
cquil11: this should prolly be `NUM_PROMPTS` since you calculate this env var and export it in `launch_b200-dgxc.sh`
unless you didn't intend to do this and just ctrl a ctrl c ctrl v it?

runners/launch_b200-dgxc.sh
```diff
@@ -0,0 +1,65 @@
...
+fi
+
+docker run --rm --init --network host --name $server_name \
+--runtime nvidia --gpus all --ipc host --privileged --shm-size=16g --ulimit memlock=-1 --ulimit stack=67108864 \
+-v $HF_HUB_CACHE_MOUNT:$HF_HUB_CACHE \
+-v $GITHUB_WORKSPACE:/workspace/ -w /workspace/ \
+-e HF_TOKEN -e HF_HUB_CACHE -e MODEL -e TP -e CONC -e MAX_MODEL_LEN -e ISL -e OSL -e PORT=$PORT -e EP_SIZE \
+-e NCCL_GRAPH_REGISTER=0 \
+-e TORCH_CUDA_ARCH_LIST="10.0" -e CUDA_DEVICE_ORDER=PCI_BUS_ID -e CUDA_VISIBLE_DEVICES="0,1,2,3,4,5,6,7" \
+-e PYTHONPYCACHEPREFIX=/tmp/pycache/ -e RESULT_FILENAME -e RANDOM_RANGE_RATIO -e NUM_PROMPTS \
```
cquil11, line 44: see here u export NUM_PROMPTS

runners/launch_b200-dgxc.sh
```diff
@@ -0,0 +1,65 @@
...
+  fi
+else
+  export NUM_PROMPTS=$(( CONC * 10 ))
+fi
+
+docker run --rm --init --network host --name $server_name \
+--runtime nvidia --gpus all --ipc host --privileged --shm-size=16g --ulimit memlock=-1 --ulimit stack=67108864 \
+-v $HF_HUB_CACHE_MOUNT:$HF_HUB_CACHE \
+-v $GITHUB_WORKSPACE:/workspace/ -w /workspace/ \
+-e HF_TOKEN -e HF_HUB_CACHE -e MODEL -e TP -e CONC -e MAX_MODEL_LEN -e ISL -e OSL -e PORT=$PORT -e EP_SIZE \
```
cquil11: again note EP_SIZE and TP exported

cquil11 (changes requested):

cquil11:
@ankursingh-nv the label pr trigger won't work here since you have `b200-test` as the label doesn't really exist
instead invoke the E2E Test workflow on your branch ref and input
```
runner-model-sweep --runner-type b200-test --config-files .github/configs/nvidia-master.yaml --runner-config .github/configs/runners.yaml
```

cquil11:
btw, not mentioned in doc but we just realized that all runner users on runner nodes must have `ALL=(ALL) NOPASSWD` in sudoers so please check this for dgxc
@ankursingh-nv cc @kedarpotdar-nv

.github/configs/runners.yaml
ankursingh-nv: We are still testing a few configs. I'll update the title to highlight that its WIP.

cquil11: ah ok mb -- oren tagged me so I thogt it was ready for review

benchmarks/gptoss_fp4_b200_trt_docker.sh
ankursingh-nv: I already see the comment in the master config file. Removing the if-block from here.

ankursingh-nv:
Updated the DGX runner tags to `b200`. Additionally, removed GPTOSS TRT B200 Docker benchmark script to avoid scoop creep. Will create a new PR for it. 

Successful Run: https://github.com/InferenceMAX/InferenceMAX/actions/runs/19712801202

Update: Refer PR #256 for GPTOSS TRT docker 


cc @kedarpotdar-nv @cquil11

kedarpotdar-nv (commented):
LGTM. Thanks @ankursingh-nv

cquil11 (approved):
lgtm

functionstackx (approved):

PR #251, opened 2025-11-21
feat: multinode first class integration

cquil11:
# Design Doc: Multi-Node First Class Integration

> **Disclaimer:** No changes in this PR affect any performance for either AMD or NVIDIA.

Test runs:
- runner model sweep: https://github.com/InferenceMAX/InferenceMAX/actions/runs/19829647492
- e2e test full sweep 1k1k sglang only: https://github.com/InferenceMAX/InferenceMAX/actions/runs/19830636342
-  e2e test full sweep 1k1k trt only: https://github.com/InferenceMAX/InferenceMAX/actions/runs/19830655526
- full sweep 1k1k big one: https://github.com/InferenceMAX/InferenceMAX/actions/runs/19966863539
- e2e test with new GB200 updates: https://github.com/InferenceMAX/InferenceMAX/actions/runs/19910261653

## Introduction

Currently, multi-node benchmarks (in particular, GB200 benchmarks) are second-class. That is, they are executed in a fundamentally different way than single node benchmarks within the InferenceMAX framework.

For instance, while all single node benchmarks follow the rule of one single scenario per GitHub Actions runner, the `runners/launch_gb200-nv.sh` script launches all configurations at once (which are hard-coded into the Bash script) and allows SLURM to handle the scheduling.

This PR seeks to standardize multinode benchmarks such that:

- The amount of logic and number of variables that are hard coded inside the low-level Bash scripts is kept to a minimum
- All multinode scenarios can be represented by a configuration in a master configuration file
- One GitHub Actions runner runs one single benchmark scenario (at one time)
  - This makes debugging much easier by isolating runs within the GitHub actions framework
- In general, they are executed in a similar fashion to single node benchmarks

This design doc will explain the proposed architecture from "highest" to "lowest" levels of abstraction:

1. Master config changes, config parsing script
2. Benchmark template / workflow changes
3. Runner / benchmark Bash script changes
4. Physical runner changes

---

### Master Configs

The first step in supporting multinode configs as first-class citizens in InferenceMAX is allowing developers to define them in the master configs. This provides a strict template in which to pass information down to the actual execution of runner and benchmarks.

Below is an example config employing the structure that multinode configurations will use (note the new fields are highlighted in **bold**):

```yaml
dsr1-fp4-gb200-dynamo-trt:
  image: nvcr.io#nvidia/ai-dynamo/tensorrtllm-runtime:0.5.1-rc0.pre3
  model: deepseek-r1-fp4
  model-prefix: dsr1
  runner: gb200
  precision: fp4
  framework: dynamo-trt
  multinode: true
  disagg: true
  seq-len-configs:
  - isl: 1024
    osl: 1024
    search-space:
    - spec-decoding: "mtp"
      conc-list: [ 1, 2, 4, 8, 16, 36 ]
      prefill:
        num-worker: 1
        tp: 4
        ep: 4
        dp-attn: false
        additional-settings:
        - "PREFILL_MAX_NUM_TOKENS=4608"
        - "PREFILL_MAX_BATCH_SIZE=4"
      decode:
        num-worker: 4
        tp: 8
        ep: 8
        dp-attn: false
        additional-settings:
        - "DECODE_MAX_NUM_TOKENS=128"
        - "DECODE_MAX_BATCH_SIZE=32"
        - "DECODE_GPU_MEM_FRACTION=0.9"
        - "DECODE_MTP_SIZE=3"
        - "DECODE_EPLB_NUM_SLOTS=0"
```

Below is a description of the additions:

| Field | Type | Description |
|-------|------|-------------|
| `multinode` | `bool` | Will be added to all new configs AND all existing configs to indicate whether or not a config is multinode; `true` indicates multinode, `false` indicates single node |
| `spec-decoding` | `string` (optional) | One of `mtp`, `draft_models`, or `none` (defaults to `none`) |
| `prefill` / `decode` | `dict` | Nested JSON objects holding the config information specific to the prefill/decode instances, respectively |
| `num-worker` | `int` | The number of prefill/decode workers |
| `tp` / `ep` | `int` | The TP and EP for the prefill/decode instances |
| `dp-attn` | `bool` | Whether DP attention should be enabled for the prefill/decode instances |
| `additional-settings` | `list[string]` | A list of strings representing additional environment variables the developers want to be passed to the underlying Bash script. This is split up between prefill/decode so as to discourage developers from passing in environment variables unrelated to prefill/decode. However, no explicit validation will be done on these. |
| `conc-list` | `list[int]` | A list of concurrencies to run for this scenario. **Note:** this will be available for single node configs as well. The developer will now get a choice whether to specify the full list of concurrencies or the classic `conc-start` and `conc-end` with a step factor of 2. The motivation for this is because, with the complexity added when creating multi-node recipes, it is not always as simple as sweeping across a "neat" distribution of concurrencies. |

#### Motivation for `additional-settings`

As we know, multinode runs are significantly more complex than single node as they introduce advanced techniques such as PD disaggregation, KV cache offload, and more. Furthermore, different serving frameworks such as vLLM, Dynamo + TRT, and SGLang will all have slightly different parameters that must be set.

Therefore, it is not practical to strictly define all of the settings (as required fields in the master config entries) required for, say, Dynamo when that field may not be relevant to say, SGLang.

In this proposal, we seek to define a standard set of variables needed to run multinode benchmarks (such as `num-worker`, `max-num-tokens`, etc.) and provide an additional section that allows developers to pass arbitrary values as environment variables to the runner launch script.

While this adds a bit of complexity both in the master configs as well as the scripts responsible for parsing the configs, we believe it is ultimately necessary as we add more and more configurations to InferenceMAX. We believe this can also be useful for some single node configs in the future.

The overarching point of this is to try and define as much information as possible in the "source of truth" master configs. InferenceMAX prides itself on transparency and ease of understanding/reproducibility – having settings defined at the top level config rather than as complex logic at the Bash script level is important.

---

### Master Config Parsing Scripts

Whenever additions are made to the master configs structure, corresponding changes must be made in the `utils/generate_sweep_configs.py` script.

In particular, the changes that must be made as part of this proposal are as follows:

1. **Validation logic must be added for the new input configs mentioned above**
   - All configurations must have `multinode` [bool]
   - If multinode, then check for multinode specific fields (both required and optional)
   - If not multinode, then check for single node specific fields

2. **Validation for outputs must be updated to support new configs**
   - Recall: output validation is in place to maintain strict integrity with what is expected from the `benchmark-tmpl.yml` workflow files

3. **Logic must be added to the `runner-model-sweep` and `full-sweep` functions** to parse the correct configs from the configs and pass them as JSON objects to stdout (for consumption by the workflow files)
   - Will have options `--single-node` or `--multi-node` to get either the single node configs or multi node configs
   - We believe it is probably best to keep single/multi node configs separate and not mix them as they will still have some inherent differences when running

Additionally, this PR will include splitting up the single parsing script into two: one for validation and one for the core logic. This is necessary because the single file is getting a bit overwhelming to maintain.

---

### Benchmark Template / Workflow

There are two approaches to consider for integrating the workflows to support multinode:

#### Option 1: Combined Template
Combine `benchmark-tmpl.yml` and `benchmark-multinode-tmpl.yml` into a single workflow template. Inputs to this file will be three-fold:
1. Single node specific inputs
2. Multinode specific inputs
3. Common arguments

The common arguments will be required while the mode specific arguments will be optional. Then, jobs/steps are run conditioned upon whether or not it is a single or multinode benchmark.

#### Option 2: Separate Templates
Keep `benchmark-tmpl.yml` and `benchmark-multinode-tmpl.yml` separate, with separate inputs and slightly different logic.

#### Analysis

| Option | Pros | Cons |
|--------|------|------|
| Option 1 | Single "point of entry" to run jobs on runners; top-level scheduler workflows don't need to split calls | Adds complex logic to template file |
| Option 2 | Straightforward logic in benchmark template script | Adds complexity at top level workflow files; requires splitting single and multi node as separate jobs |

#### Recommendation

**We lean towards Option 2.** While it adds more lines of code in the top level workflows, we believe single and multinode runs are inherently different and should be split up accordingly.

The structure will be as follows:

```yaml
name: "Full Sweep Scheduler - 1k1k"

on:
  workflow_dispatch:
  schedule:
    - cron: "0 0 * * *"

jobs:
  # Get DeepSeek configs and store as separate output variables for single and multi node
  get-dsr1-configs:
    ...

  # Get GPT OSS configs and store as separate output variables for single and multi node
  get-gptoss-configs:
    ...

  benchmark-dsr1-single-node:
    needs: get-dsr1-configs
    uses: ./.github/workflows/benchmark-tmpl.yml
    name: dsr1 1k1k / ...

  benchmark-gptoss-single-node:
    needs: get-gptoss-configs
    uses: ./.github/workflows/benchmark-tmpl.yml
    name: gptoss 1k1k / ...

  benchmark-dsr1-multi-node:
    needs: get-dsr1-configs
    uses: ./.github/workflows/benchmark-multinode-tmpl.yml
    name: dsr1 1k1k / ...

  benchmark-gptoss-multi-node:
    needs: get-gptoss-configs
    uses: ./.github/workflows/benchmark-multinode-tmpl.yml
    name: gptoss 1k1k / ...

  collect-dsr1-results:
    needs: [benchmark-dsr1-single-node, benchmark-dsr1-multi-node]
    if: ${{ always() }}
    uses: ./.github/workflows/collect-results.yml
    secrets: inherit
    with:
      exp-name: "dsr1_1k1k"

  collect-gptoss-results:
    needs: [benchmark-gptoss-single-node, benchmark-gptoss-multi-node]
    if: ${{ always() }}
    uses: ./.github/workflows/collect-results.yml
    secrets: inherit
    with:
      exp-name: "gptoss_1k1k"

  calc-success-rate:
    needs: [benchmark-dsr1, benchmark-gptoss, benchmark-gb200]
    if: ${{ always() }}
    runs-on: ubuntu-latest
    ...
```

#### Concurrency Handling

We propose running all concurrencies for each scenario in the same run, on the same inference server. The reason we opt for this instead of the traditional one concurrency per server is the time it takes to spin up disaggregated inference servers. We don't want to waste an insane amount of compute ($40-50 USD) for each concurrency.

This will require some changes to the `bench_serving` script to allow multiple concurrencies to be run at once.

---

### Runner/Bash Scripts

Recall that PR #227 standardizes the architecture in which all single node benchmarks are run. Specifically, `benchmark-tmpl.yml` uses the name of the runner, e.g., `b200-nvd_0` to decide which `runners/launch_XXXX.sh` script to invoke, which subsequently launches a container. The entrypoint to this container is a `benchmarks/MODEL_PRECISION_GPU_FRAMEWORK.sh` script, based on the associated inputs. As such, this `benchmarks/` script will run entirely inside a container.

We will follow a similar approach for multinode integration. However, the `runners/` script will not launch a container that the entire benchmark runs in, it will just call a script specific to the model, precision, GPU, framework combination that will kick off a series of scripts starting up the prefill, decode, server, and benchmark processes.

---

### Physical Runners

Recall that currently, the only multinode job (running on GB200) launches a single script that submits ALL scenarios at once to SLURM. This is all done on one single GitHub Actions runner. This massively oversubscribes and takes advantage of the SLURM scheduler (however, we note that on the current GB200 cluster, the scheduling method is FIFO).

As mentioned previously, it is beneficial to have one scenario per job per runner (for debugging purposes, testing, reproducibility, etc). However, we would still like to oversubscribe and take advantage of the SLURM scheduler.

#### Proposal

**Add three additional GitHub Actions runners listening on the login node.**

Current GB200 NVL72 configurations average 8.4 nodes per job. With 4 runners submitting jobs concurrently on our 18-node cluster, this results in an oversubscription factor of approximately **1.86**.

We propose starting at 4 and decreasing/increasing as necessary.

> **Note:** This will require either getting sudo access on the GB200 cluster (to add the new `gharunner` users) or getting an SRE to do this for us.

#### Tradeoffs

This architecture runs one job per runner, oversubscribing up to 4 jobs. The tradeoff here is that each scenario is run in random order, and since all jobs are not submitted at once, we cannot take full advantage of traditional SLURM scheduling techniques.

As mentioned previously, the tradeoff here is that we will achieve better visibility into the multi-node runs and these runs will be easier to debug, as each run will be isolated to a single job.

---

## Frontend Considerations

There are no frontend considerations.

---

## Testing Workflows

As part of this PR, we will make all the necessary changes to make tests for multinode first class as well. We will get rid of the specific workaround logic for multinode (GB200).

---

## Follow-Ups

In designing this proposal, we realize that there are a lot of downstream scripts and programs that we should upstream into the InferenceMAX repo (at the very least as a submodule). Again, InferenceMAX prides itself on being transparent and easily reproducible. When someone has to go to a completely separate repo and sift through a stack trace of 8 Dynamo files, we are going against this principle.

**We should try to minimize the amount of `git clone`-ing that we do at all costs.**

In particular, below is a list of files that should be upstreamed as follow up PRs:

1. **Kimbo's `bench_serving` script**
   - We should standardize this and add it (the actual script) to the upstream InferenceMAX repo

2. **Dynamo scripts for PD-disagg**
   - We should upstream that SBATCH and python scripts, or perhaps even re-write them to make them more straightforward
   - Right now, they are quite confusing with many levels of indirection

- https://github.com/InferenceMAX/InferenceMAX/issues/260
- https://github.com/InferenceMAX/InferenceMAX/issues/254
- https://github.com/InferenceMAX/InferenceMAX/issues/255


.github/configs/nvidia-master.yaml
```diff
@@ -314,3 +325,522 @@ gptoss-fp4-h200-vllm:
...
+        num-worker: 1
+        tp: 16
+        ep: 16
+        dp-attn: true
+        additional-settings:
+        - "DECODE_MAX_NUM_TOKENS=256"
+        - "DECODE_MAX_BATCH_SIZE=128"
+        - "DECODE_GPU_MEM_FRACTION=0.7"
+        - "DECODE_MTP_SIZE=1"
+        - "DECODE_EPLB_NUM_SLOTS=0"
```
functionstackx: nit: `DECODE_EPLB_NUM_SLOTS=0` is always 0, just hardcore it into 0 and remove it at top level.

it is 0 cuz they dont have eplb enabled in this dynamo scripts rn

Image

utils/process_result.py
```diff
@@ -4,58 +4,118 @@
...
 }
 
-# Check if both PREFILL_GPUS and DECODE_GPUS env vars exist and are not empty
-if prefill_gpus_str and decode_gpus_str:
-    data['disagg'] = True
-    data['num_prefill_gpu'] = prefill_gpus
-    data['num_decode_gpu'] = decode_gpus
+is_multinode = os.environ.get('IS_MULTINODE', 'false').lower() == 'true'
+
+if is_multinode:
```
functionstackx, line 61: plz modify https://github.com/InferenceMAX/InferenceMAX/blob/main/utils/summarize.py too cuz rn summarize.py crashses from rpocess_result changes

.github/workflows/full-sweep-test.yml
```diff
@@ -1,445 +0,0 @@
-name: Test - Full Sweep
```
cquil11, line 1: fyi - opting to remove this because it's not very useful
anything that can be run with this can be run via e2e test workflow

.github/workflows/gb200-tests.yml
```diff
@@ -1,91 +0,0 @@
-name: GB200 Tests
```
cquil11, line 1: fyi - removing this bc gb200 (and other multinode) first class now

.github/workflows/collect-results.yml
```diff
@@ -39,15 +39,3 @@ jobs:
         with:
           name: results_${{ inputs.exp-name || 'all' }}
           path: agg_${{ inputs.exp-name || 'all' }}.json
-
-      - name: Plot performance
```
cquil11, line 43: fyi - unreleated to this PR but this is no longer needed and can be removed cc @kimbochen

utils/matrix-logic/get_test_sweep_configs.py
```diff
@@ -1,151 +0,0 @@
-import json
```
cquil11, line 1: fyi - this file was stale and unused

utils/scrape_image_tag.py
```diff
@@ -1,43 +0,0 @@
-import sys
```
cquil11, line 1: fyi - this file was stale and unused

jthomson04:
Our current sgl and trtllm (once that gets merged) disagg submissions use unique launch scripts, which will complicate the stated goal of upstreaming our scripts to the this repo.  Because of that (and how every framework needs different config options), it would be nice to include support for a framework config option in the yaml configs. We could (in theory) handle that through the `additional-settings` env vars, but first-class support for that would be preferred.

cquil11:
@jthomson04 
> Our current sgl and trtllm (once that gets merged) disagg submissions use unique launch scripts, which will complicate the stated goal of upstreaming our scripts to the this repo. Because of that (and how every framework needs different config options), it would be nice to include support for a framework config option in the yaml configs. We could (in theory) handle that through the `additional-settings` env vars, but first-class support for that would be preferred.

just to be sure I understand your suggestion, youre proposing something to the same effect as `additional-settings` for prefill decode specific things, but more generally for different frameworks? I.e., have a `framework-settings` field that make framework specific env vars "first class"?

if this is what you mean, typically we handle this by having separate `benchmarks/` scripts for different frameworks (see below):
image
these scripts then have "hard coded" env vars / settings for each framework

on a separate note, regarding upstreaming the different launch scripts for SGLang + TRT:  I've given it some thought and I don't think it will be too bad. it seems like theres a lot of unnecessary indirection in the branches used that we can probably do away with when upstreaming to InferenceMAX, take the way in which TRT jobs are currently launched for instance:

- we use this release branch https://github.com/ai-dynamo/dynamo/tree/release/0.5.1-rc0.20251105/components/backends/trtllm
- then we call [`submit_disagg.sh`](https://github.com/ai-dynamo/dynamo/blob/release/0.5.1-rc0.20251105/components/backends/trtllm/performance_sweeps/submit_disagg.sh) which is an "unnecessary" wrapper for the [`benchmark_disagg.slurm`](https://github.com/ai-dynamo/dynamo/blob/release/0.5.1-rc0.20251105/components/backends/trtllm/performance_sweeps/benchmark_disagg.slurm) sbatch
- then  [`benchmark_disagg.slurm`](https://github.com/ai-dynamo/dynamo/blob/release/0.5.1-rc0.20251105/components/backends/trtllm/performance_sweeps/benchmark_disagg.slurm) `srun`s all of the worker processes via programs in [`scripts/`](https://github.com/ai-dynamo/dynamo/tree/0b6abc828e8e705217f0e3949092cb4dd2cccb89/components/backends/trtllm/performance_sweeps/scripts)
- also, instead of using upstream benchmark script, we use a separate one (as a submodule) in this branch

this same indirection can be observed with the way SGLang is launched. of course, disagg serving is not trivial so some level of complexity is expected -- we just wanna get to a point where any unnecessary indirection is cut out.

I think there's a simpler way to set this up in the upstream InferenceMAX repo such that someone can come to repo, click through a couple files and be like "oh shit it's actually really simple to run Dynamo with TRT or SGLang" which is one of the primary goals of Infmax. I'm sure y'all also have upstream demos/benchmarks within the SGLang/Dynamo repos, but I think it's gonna be a great effort to simplify + upstream to inferencemax for increased visibility to showcase how great dynamo is. This is something i'll largely be working on leading up to Christmas and New Year. Open to any feedback y'all have.

runners/launch_gb200-nv.sh
```diff
@@ -293,23 +119,30 @@ if [[ $FRAMEWORK == "dynamo-trtllm" ]]; then
...
-for path in sorted([f"logs/{name}/vllm_isl_{isl}_osl_{osl}" for name in os.listdir("logs/") if os.path.isdir(f"logs/{name}/vllm_isl_{isl}_osl_{osl}")], key=os.path.getmtime, reverse=True)[:nexp]:
+sgl_job_dir, isl, osl, nexp = sys.argv[1], int(sys.argv[2]), int(sys.argv[3]), int(sys.argv[4])
+for path in sorted([f"{sgl_job_dir}/logs/{name}/vllm_isl_{isl}_osl_{osl}" for name in os.listdir(f"{sgl_job_dir}/logs/") if os.path.isdir(f"{sgl_job_dir}/logs/{name}/vllm_isl_{isl}_osl_{osl}")], key=os.path.getmtime, reverse=True)[:nexp]:
     print(path)
 PY
-    LOGS_DIR=$(python3 collect_latest_results.py $ISL $OSL $NUMBER_OF_EXPERIMENTS)
+
+    # This isn't ideal but for now, this is needed for the collect_latest_results.py script
+    if [ "$ISL" = "1024" ] && [ "$OSL" = "1024" ]; then
+        NUMBER_OF_EXPERIMENTS=3
```
yunzhoul-nv: I think it's not necessary anymore to keep `NUMBER_OF_EXPERIMENTS=3` here. In the new structure, each runner will only run 1 experiment, thus the logs are contained under the folder whose content is modified most recently, and thus we need to grab only `NUMBER_OF_EXPERIMENTS=1` folder, not 3 folders. 

In our original setup, one job executes three experiments, thus we need to find the latest 3 folders that are modified most recently.

runners/launch_gb200-nv.sh
```diff
@@ -2,11 +2,14 @@
...
 # This script sets up the environment and launches multi-node benchmarks
 
+set -x
 
 # Set up environment variables for SLURM
 export SLURM_PARTITION="batch"
 export SLURM_ACCOUNT="benchmark"
 export SLURM_JOB_NAME="benchmark-dynamo.job"
+# For GB200 we have 4 GPUs per node
+export NTASKS_PER_NODE=4
```
yunzhoul-nv: I would actually recommend to move this to under `dsr1_fp4_gb200_dynamo-trt_slurm.sh` instead, since Dynamo SGLang code needs `NTASKS_PER_NODE=1`.

runners/launch_gb200-nv.sh
cquil11: ah right, good catch

runners/launch_gb200-nv.sh
cquil11: fixed 5eb8ea1

runners/launch_gb200-nv.sh
cquil11: fixed 711783f

cquil11: testing here https://github.com/InferenceMAX/InferenceMAX/actions/runs/19873281184

jthomson04:
@cquil11 Thanks for the explanation. Looks like the existing setup with the per-backend script should be sufficient.

PR #257, opened 2025-11-26
[NVIDIA] feat: refresh GB200 SGLang DSR1 submission

yunzhoul-nv:
Introducing some improvements in GB200 SGLang DSR1 submission. 

Successful pipeline: https://github.com/InferenceMAX/InferenceMAX/actions/runs/19714799633.


yunzhoul-nv:
@functionstackx @cquil11 @kimbochen May I get a review? Thanks!

runners/launch_gb200-nv.sh
```diff
@@ -157,13 +165,24 @@ if [[ $FRAMEWORK == "dynamo-trtllm" ]]; then
 
 else # if statement at the top - search for "FRAMEWORK_DIFF_IF_STATEMENT #2"
     # Set up Dynamo repository path
+    set -x
     DYNAMO_PATH="/mnt/lustre01/users/sa-shared/benchmarks/dynamo"
-    SGL_SLURM_JOBS_PATH="$DYNAMO_PATH/components/backends/sglang/slurm_jobs"
+    if [ "$ISL" = "1024" ] && [ "$OSL" = "1024" ]; then
+        SGL_SLURM_JOBS_PATH="$DYNAMO_PATH/examples/backends/sglang/slurm_jobs"
```
cquil11, line 171: eventually we should try to move to a single script that (that possbly upstreamed to InferenceMAX) instead of forking different ones based on isl/osl, framework, etc

runners/launch_gb200-nv.sh
```diff
@@ -259,9 +295,14 @@ if [[ $FRAMEWORK == "dynamo-trtllm" ]]; then
     done
 
 else # search for "FRAMEWORK_DIFF_IF_STATEMENT #3" for this if-statement
-    # Find the latest log directory
-    # we do "tail -1" here since only the latest job will yield the result
-    LOGS_DIR=$(find logs/*/vllm_isl_${ISL}_osl_${OSL} -type d | sort -V | tail -1)
+    # Find the latest log directory that contains the data
+    cat > collect_latest_results.py <<'PY'
```
cquil11, line 299: can you explain why this is necessary?

runners/launch_gb200-nv.sh
ishandhanani: Agreed - this can be done in a follow up PR

runners/launch_gb200-nv.sh
yunzhoul-nv: Hi @cquil11 , the original implementation in bash strictly gets one latest experiment result, which worked for the 1k1k and 8k1k experiments before since we only do one experiment for each ISL/OSL combination. However, this PR updates the 1k1k configuration so that we run 3 experiments concurrently for 1k1k ISL/OSL, thus the results are scattered across 3 different experiment folders, instead of in the latest 1. 

Additionally, in our updated branch, the experiment folder names are renamed so that `sort -V` no longer works with the updated experiment folder names. We also feel that python code pieces is more maintainable and has more readability compared to implementing these logics in bash. 

Thus, we have chosen to execute a small piece of python code here to extract the updated experiment results instead of fixing the logics in bash.

cquil11:
lgtm, thanks!

cquil11 (approved):

Issue #304, opened 2025-12-07
instead of label auto-validation, move towards GH comment hook


cquil11:
Right now, when we want to run a test job from a PR, we use the `label-validation.yml` workflow which parses the labels on a PR to run smoke test sweeps.

In order to give users more control over tests and to not have 3224 labels (which I think is kinda ugly), just have a hook that listens to a PR comment so that users can run commands on a PR to trigger test sweeps. Something like

/sweep <generate_sweep_configs.py command>

Which literally just takes in the CLI command and passes it to the `generate_sweep_configs.py ` file then runs the workflow.

Starter references: 
- https://johnnyreilly.com/slash-command-your-deployment-with-github-actions
- https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/about-slash-commands

PR #306, opened 2025-12-08
For #304: GH comment hook

Oseltamivir:
# Summary
Addresses #304, Adds PR comment–triggered sweeps via `/sweep ...`, removing reliance on labels. Keeps logic DRY by reusing existing benchmark templates and a new reusable executor.

# How
1. `issue_comment` on PRs parses the first `/sweep` line and forwards args to a reusable workflow.
2. Executor checks out PR head, runs `utils/matrix_logic/generate_sweep_configs.py`, detects single vs multi, dispatches to `benchmark-tmpl.yml` or `benchmark-multinode-tmpl.yml`, then collects results.

# What
Added: `.github/workflows/pr-comment-sweep.yml`: small parser; auth gate; per‑PR concurrency.
Added: `.github/workflows/sweep-executor.yml`: generates matrix, runs benchmarks, aggregates results.

# Examples

- `/sweep full-sweep --single-node --runner-type h200 --model-prefix dsr1 --seq-lens 1k1k`
- `/sweep full-sweep --multi-node --model-prefix gptoss --seq-lens 1k1k`
- `/sweep runner-model-sweep --runner-type h200 --single-node`
No need to pass config paths; executor appends --config-files and --runner-config.

# Misc & Security
Only runs on PR comments from OWNER/MEMBER/COLLABORATOR.
Workflows execute with secrets: inherit; generator runs on PR head.

# Notes
- `issue_comment` workflows must exist on the default branch to trigger. Currently tries a `workflow_dispatch` workaround
- Benchmarks currently checkout default branch (matching existing templates). If desired, we can add a ref input to run jobs on PR head.
- Label-based path remains unchanged.


Oseltamivir:
Can't be actually tested until pushed to `main` I believe, but the there is a default: `full-sweep --single-node --runner-type h200 --model-prefix dsr1 --seq-lens 1k1k --max-conc 4` that is ran [here](https://github.com/InferenceMAX/InferenceMAX/actions/runs/20032697985) on push

.github/workflows/pr-comment-sweep.yml
```diff
@@ -0,0 +1,126 @@
...
+        shell: bash
+        env:
+          DISPATCH_PR: ${{ github.event.inputs.pr-number }}
+          DISPATCH_ARGS: ${{ github.event.inputs.generator-args }}
+        run: |
+          set -euo pipefail
+          pr_from_branch='${{ steps.find.outputs.pr-number }}'
+          pr_number="${DISPATCH_PR:-}"; if [[ -z "$pr_number" ]]; then pr_number="$pr_from_branch"; fi
+          gen_args="${DISPATCH_ARGS:-}"
+          if [[ -z "$gen_args" ]]; then
```
cquil11: no default
should just fail in this case

.github/workflows/pr-comment-sweep.yml
```diff
@@ -0,0 +1,126 @@
...
+      - name: Parse PR comment (/sweep ...)
+        id: parse
+        if: ${{ github.event_name == 'issue_comment' && github.event.issue.pull_request && startsWith(github.event.comment.body, '/sweep') }}
+        shell: bash
+        env:
+          BODY: ${{ github.event.comment.body }}
+          PR_NUMBER: ${{ github.event.issue.number }}
+        run: |
+          set -euo pipefail
+          # Allow optional leading whitespace before /sweep
```
cquil11: I argue for don't allow leading whitespace
This is how Slack does it
it only runs commands with no whitespace in front and treats everything else as text
This allows people to "refer" to a command without actually triggering the command
For instance, if I wanted to comment "thanks for the PR, please run /sweep command" or whatever, then this would currently trigger, and we don't want that

.github/workflows/pr-comment-sweep.yml
Oseltamivir: The default(`gen_args='full-sweep --sin...`) is for testing, will remove when ready to merge

.github/workflows/pr-comment-sweep.yml
Oseltamivir: Changed

cquil11:
@Oseltamivir can the callback message / comment include the command used to run the sweep?
```
@Oseltamivir pushed changes and triggered a sweep. With the command <Code formatted command>.

Run: https://github.com/InferenceMAX/InferenceMAX/actions/runs/20238250232
(Push on refs/heads/slash-commands)
```

cquil11 (approved):
lgtm

PR #371, opened 2025-12-22
Cleanup testing code, Fix job concurrency

Oseltamivir:
Cleanup extra validation, did concurrency at job level so non-/sweep comments don't cancel active runs.

One concurrency group per PR, so a newer "validate" for the same PR cancels any in‑progress "validate" from that PR (`cancel-in-progress: true`), and then run the new one.


cquil11 (approved):
lgtm

PR #411, opened 2026-01-09
[NVIDIA] feat: adds more SGLang B200 configs

yunzhoul-nv:
> [!NOTE]
> Introduces TP-aware configurations and launch behavior for DeepSeek R1 FP8 on B200.
> 
> - Adds `tp:4, ep:1` search-space for `dsr1-fp8-b200-sglang` at `isl:8192, osl:1024`; constrains `tp:8` there to `conc-end:4`
> - Updates `benchmarks/dsr1_fp8_b200_{docker,slurm}.sh` to set scheduler interval and server args dynamically per TP: TP8 uses 128 max running/cuda-graph BS and 32k prefill; TP4 (8k/1k only) uses 32 and 8k prefill; errors on unsupported combos
> - Records change in `perf-changelog.yaml`; workflow file gains newline-only cleanup
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3f19396cebef1c3eda6328f2158a23dd4c22ba1b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


yunzhoul-nv:
Hi @cquil11 @functionstackx , could you give a review on this PR? This PR aims to update the SGLang DSR1-FP8 configuration we run on B200.

benchmarks/dsr1_fp8_b200_slurm.sh
```diff
@@ -22,18 +22,38 @@ SERVER_LOG=$(mktemp /tmp/server-XXXXXX.log)
...
+  CHUNKED_PREFILL_SIZE=32768
+  MAX_PREFILL_TOKENS=32768
+elif [[ $TP -eq 4 ]]; then
+  if [[ $ISL -ne 8192 ]] || [[ $OSL -ne 1024 ]]; then echo "TP=4 not yet supported for ISL=$ISL OSL=$OSL!"; fi
+  CG_MAX_BS=32
+  MAX_RUNNING_REQUESTS=32
+  MEM_FRAC_STATIC=0.95
+  CHUNKED_PREFILL_SIZE=8192
+  MAX_PREFILL_TOKENS=8192
+
```
functionstackx: hi @yunzhoul-nv ! to summarize these changes in this file, basically TP =4 will have def params from TP=8?

if so, putting this in the sglang cookbook recipes & usage would be amazing!

https://cookbook.sglang.io/docs/autoregressive/DeepSeek/DeepSeek-R1

https://docs.sglang.io/basic_usage/deepseek_v3.html#

functionstackx:
@cquil11 can help review here

benchmarks/dsr1_fp8_b200_slurm.sh
```diff
@@ -22,18 +22,38 @@ SERVER_LOG=$(mktemp /tmp/server-XXXXXX.log)
...
+  exit 1
 fi
 echo "SCHEDULER_RECV_INTERVAL: $SCHEDULER_RECV_INTERVAL, CONC: $CONC, ISL: $ISL, OSL: $OSL"
 
 set -x
 PYTHONNOUSERSITE=1 python3 -m sglang.launch_server --model-path=$MODEL --host=0.0.0.0 --port=$PORT \
 --tensor-parallel-size=$TP --data-parallel-size=1 \
---cuda-graph-max-bs 128 --max-running-requests 128 \
---mem-fraction-static 0.82 --kv-cache-dtype fp8_e4m3 --chunked-prefill-size 32768 --max-prefill-tokens 32768 \
+--cuda-graph-max-bs $CG_MAX_BS --max-running-requests $MAX_RUNNING_REQUESTS \
```
cquil11: @yunzhoul-nv can you explain the reasoning on dynamically setting max running requests?

perf-changelog.yaml
```diff
@@ -156,3 +156,9 @@
     - "Add PD disaggregation (1P2D) for Mi355X"
     - "Includes with and without speculative decoding"
   pr-link: https://github.com/InferenceMAX/InferenceMAX/pull/348
+
+- config-keys:
+    - dsr1-fp8-b200-sglang
+  description:
+    - "adds more SGLang B200 configs"
```
cquil11: can you be a bit more descriptive here
these are displayed on the UI now, so good to give users a full description

benchmarks/dsr1_fp8_b200_docker.sh
```diff
@@ -23,10 +23,30 @@ export SGLANG_ENABLE_FLASHINFER_GEMM=true
...
+    SCHEDULER_RECV_INTERVAL=10
+  fi
+
+  CG_MAX_BS=128
+  MAX_RUNNING_REQUESTS=128
+  MEM_FRAC_STATIC=0.82
+  CHUNKED_PREFILL_SIZE=32768
+  MAX_PREFILL_TOKENS=32768
+elif [[ $TP -eq 4 ]]; then
+  if [[ $ISL -ne 8192 ]] || [[ $OSL -ne 1024 ]]; then echo "TP=4 not yet supported for ISL=$ISL OSL=$OSL!"; fi
```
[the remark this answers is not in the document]
cquil11: @yunzhoul-nv any merit to this?

benchmarks/dsr1_fp8_b200_slurm.sh
```diff
@@ -22,18 +22,38 @@ SERVER_LOG=$(mktemp /tmp/server-XXXXXX.log)
...
-else
+if [[ $TP -eq 8 ]]; then
+  if [[ $CONC -ge 16 ]]; then
+    SCHEDULER_RECV_INTERVAL=30
+  else
+    SCHEDULER_RECV_INTERVAL=10
+  fi
+
+  CG_MAX_BS=128
+  MAX_RUNNING_REQUESTS=128
```
cquil11, line 36: again, would love to hear some explanation for why these recipes were chosen accordingly for tp8 and tp4 respectively. In particular, what is the reasoning for setting max running requests?
as the configs stand right now, `MAX_RUNNING_REQUESTS=128` doesn't really do anything right? because the max concurrency ran for this config TP8 is 64.
Would love to understand resaoning!

benchmarks/dsr1_fp8_b200_docker.sh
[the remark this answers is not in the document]
yunzhoul-nv: Added the exit lines.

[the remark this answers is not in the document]
yunzhoul-nv: Hey @cquil11 , I have updated the code to include `exit 1` for unsupported configurations.

benchmarks/dsr1_fp8_b200_slurm.sh
yunzhoul-nv: Hey @functionstackx , yes, we will have slightly different `--memory-fraction-static`, `--chunked-prefill-size` and `--max-prefill-tokens` compared to TP8, but generally other parameters remain the same compared to TP8.

benchmarks/dsr1_fp8_b200_slurm.sh
yunzhoul-nv: Hey @cquil11 , thanks for catching this! In our experiments, we are not using the `--max-running-requests` to limit the max number of requests that the worker can handle, and we are just setting the `max-running-requests` to the maximum concurrency so that the engine will not waste additional memories preparing for concurrencies that we do not use. 

I have adjusted the code so that both TP4 and TP8 uses the same `--max-running-requests=32`, instead of introducing this confusing selection based on the TP values.

yunzhoul-nv: Oh, actually [cursor just pointed out to me](https://github.com/InferenceMAX/InferenceMAX/pull/411#discussion_r2683954219) that I still need to keep `--max-running-requests >= 64` for TP8 ... 

I think the cleanest solution for now is still to restore `--max-running-requests=128` for TP8 because this value is consistent with the main branch, and I think keeping TP8 configuration the same as the main branch will allow us to focus more on bringing TP4 configurations to the repository in this PR.

perf-changelog.yaml
yunzhoul-nv: Thanks for the feedback! I've updated the description here to "Adds TP4 configurations to DSR1-FP8 B200 SGLang deployment experiments", so that people know the perf change is from the new TP4 config that we just added.

yunzhoul-nv:
Hi @cquil11 @functionstackx , could you give another round of review for this PR? I have resolved all conflict between this PR and main branch and the sweeps have passed, so I think this PR is safe to merge.

benchmarks/dsr1_fp8_b200_slurm.sh
cquil11: If the goal is to decrease max-running-requests as much as possible, I guess I am kinda confused on why it wouldn't just be set equal to `CONC`. In bench_serving client, `CONC` is the maximum number of requests that can be inflight at one time, so it is (to the best of my understanding) the same thing as setting `max-running-requests` at server start time.

@yunzhoul-nv maybe I am misunderstanding still :) we will merge this now, it looks good other than this question. I am just slightly confused on why this wouldn't just be set to conc.

cquil11:
@yunzhoul-nv im gonna merge, but left one more comment to follow up on later, you can just respond in this thread! thx for the contribution

cquil11 (approved):

benchmarks/dsr1_fp8_b200_slurm.sh
yunzhoul-nv: Thanks for approving the PR! 

I'm currently not setting `max-running-requests == CONC` is that in my local development, I only spin up the worker once and let the benchmarking client runs multiple concurrencies against this same worker (i.e., given a sequence of concurrencies from 4 to 32, I only spin up the worker that supports the 32 concurrency, and run concurrency 4~32 against this same worker).  On the other hand, if I set `max-running-requests == CONC`, then the worker that I spin up at concurrency=4 will not be applicable to concurrency=32, and therefore I need to spin up the worker again. 

This setup allows us to reduce the time and compute resources that are spent on spinning up multiple workers. Otherwise, we need to create a new worker for each of the concurrencies that we want to run, which is not scalable because workers could potentially need a long startup time due to compilations, or if we need to spend additional times warming up the workers.

PR #438, opened 2026-01-15
chore: hash single-node file output on spec decoding value

cquil11:
Single node runs can also have speculative decoding, so we should hash on this to ensure all output files are unique.


cquil11:
/sweep full-sweep --seq-lens 1k1k --max-conc 4 --single-node --config-files .github/configs/amd-master.yaml .github/configs/nvidia-master.yaml --runner-config .github/configs/runners.yaml

Oseltamivir (approved):
lgtm

PR #511, opened 2026-01-21
chore: make enroot import idempotent

cquil11:
For SLURM runners, in the case that the job is cancelled in the middle of `enroot import`, the squash file gets corrupted and the subsequent run on that node will fail to import immediately, forcing engineer to manually delete.

This should make the `enroot import` idempotent by checking it's validity prior to use.


cquil11:
@pr-claude review

Oseltamivir (approved):
lgtm

functionstackx:
@pr-claude review

functionstackx:
lgtm

PR #576, opened 2026-01-27
chore: save server long as artifact after single node runs

cquil11:
Changes
- Change server log from a tmp directory to GHA workspace such that it can be uploaded as artifact at end of run
- Always runs even when job cancelled


Oseltamivir:
@pr-claude review

cquil11:
smoke test: https://github.com/InferenceMAX/InferenceMAX/actions/runs/21413362603

Oseltamivir (approved):
lgtm

PR #587, opened 2026-01-27
chore: trigger full sweep to add GSM8k to official results

cquil11:
Trigger full sweep for single node cluster for evals.


Oseltamivir (approved):
lgtm

PR #590, opened 2026-01-27
chore: calculate success rate based on all workflow attempts

### LINKED ISSUE #589: calc success rate should calculate based on ALL attempts in a run (not just latest attempt)
Currently, the `calc_success_rate.py` script calculates the success rate based on only the most recent run. It should use all of the runs (the "retry runs" in GitHub actions) to calculate this figure.

This will make the reliability charts on the frontend a better proxy.

cquil11:
Previously, `calc_success_rate.py` only considered jobs from the latest attempt when calculating GPU success rates. This meant that if a job failed on the first attempt but succeeded on a retry, only the success would be counted.

This change uses `_filter="all"` parameter in `run.jobs()` to include jobs from ALL attempts (retries), providing a more accurate success rate that better reflects actual reliability.

Also removes hardcoding of GPU skus (now gets it from `runners.yaml`).

Closes #589

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/utils/calc_success_rate.py
+++ b/utils/calc_success_rate.py
@@ -3,10 +3,28 @@ import os
 import re
 import sys
 from enum import Enum
+from pathlib import Path
 
+import yaml
 from github import Auth, Github
 
-GPU_SKUS = ["h100", "h200", "gb200", "mi300x", "mi325x", "mi355x", "b200"]
+
+def load_gpu_skus():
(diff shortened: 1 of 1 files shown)


cquil11:
@pr-claude review

Oseltamivir (approved):
lgtm

PR #612, opened 2026-02-01
GPQA 

Oseltamivir:
lm-eval GPQA with n-attempt support like https://github.com/openai/gpt-oss/blob/main/gpt_oss/evals/gpqa_eval.py

- Achieves vllm published gptoss-120b GPQA score of 72.4 on 8 attempts. [ref](https://docs.vllm.ai/projects/recipes/en/latest/OpenAI/GPT-OSS.html#accuracy-evaluation-panels)
- However, we do 2-shot with 3 attempts to reduce stderr while keeping evals manageble

Run at https://github.com/InferenceMAX/InferenceMAX/actions/runs/21558005706

Goal:
- Show that we can run multiple evals, currently we are sticking with GSM8K
- Provide example json for frontend to use


Oseltamivir:
@pr-claude review, ignore the `--break-system-packages`, it is necessary to be able to `pip install` within the container.

PR #620, opened 2026-02-02
[NV] dsr1 fp4 b200 trt agg update

camiloamoreno:
This PR contains below updates:
- Update to the scripts to use latest TRTLLM release container
- Fine-tune choice of parallelism in nvidia-master for various cases
- Enable a few optimizations offered by TensorRT LLM on specific cases (MOE_BACKEND and Piecewise cuda graphs)

Near the top of the benchmark script (L26-28), we use flags to enable specific optimizations for specific ISL/OSL/CONC combinations according to the chosen TP/EP/DP-ATTENTION in nvidia-master and to what is best for balancing ctx vs gen workload and achieving good utilization.

Further down the script, from L47 onwards, we key off these flags to generate the necessary yaml for enabling these features in the runtime config file.

Piecewise Cuda Graphs (https://nvidia.github.io/TensorRT-LLM/features/torch_compile_and_piecewise_cuda_graph.html) enables some components to execute thorugh cuda graphs while other components are run eagerly, to gain benefit with lower overhead. We use the formula from the documentation to generate a capture_num_tokens list dynamically as it depends on MAX_NUM_TOKENS which is not constant across cases.

"cuda graph max batch size" is mostly equal to CONC as the natural limit on the expected size of a batch, but is configured as batch_size/4 when DP_ATTENTION is enabled.

### DIFF
--- a/.github/configs/nvidia-master.yaml
+++ b/.github/configs/nvidia-master.yaml
@@ -833,6 +833,9 @@ dsr1-fp4-b200-trt:
   - isl: 1024
     osl: 1024
     search-space:
+    # low concurrency cases use TP only
+    # concurrency 64 uses TP & EP
+    # high concurrency cases use TP & EP & DP-ATTN
     - { tp: 4, conc-start: 4, conc-end: 16 }
     - { tp: 4, ep: 4, dp-attn: true, conc-start: 256, conc-end: 256 }
     - { tp: 8, conc-start: 4, conc-end: 4 }
@@ -841,6 +844,9 @@ dsr1-fp4-b200-trt:
   - isl: 1024
     osl: 8192
     search-space:
+    # low concurrency cases use TP only
+    # concurrency 64 uses TP & EP
+    # high concurrency cases use TP & EP & DP-ATTN
     - { tp: 4, conc-start: 4, conc-end: 16 }
     - { tp: 4, ep: 4, dp-attn: true, conc-start: 256, conc-end: 256 }
     - { tp: 8, conc-start: 4, conc-end: 4 }
@@ -849,6 +855,9 @@ dsr1-fp4-b200-trt:
   - isl: 8192
     osl: 1024
     search-space:
+    # low concurrency cases use TP only
+    # concurrency 32 uses TP & EP
+    # high concurrency cases use TP & EP & DP-ATTN
     - { tp: 4, conc-start: 4, conc-end: 32 }
     - { tp: 4, ep: 4, conc-start: 32, conc-end: 32 }
     - { tp: 4, ep: 4, dp-attn: true, conc-start: 256, conc-end: 256 }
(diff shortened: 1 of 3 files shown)


functionstackx:
hi @camiloamoreno thanks for the PR, can u please update the trtllm recipes about this?

https://docs.nvidia.com/llm-inference-quick-start-recipes/index.html
https://github.com/NVIDIA/TensorRT-LLM/blob/main/docs/source/deployment-guide/deployment-guide-for-deepseek-r1-on-trtllm.md

@claude can u create an ticket about updating the deepseek trtllm recipe and assigning @kedarpotdar-nv and @camiloamoreno to it

Klaud-Cold:
**Claude encountered an error** —— [View job](https://github.com/InferenceMAX/InferenceMAX/actions/runs/21609595424)

---
I'll analyze this and get back to you.

.github/configs/nvidia-master.yaml
```diff
@@ -833,44 +833,27 @@ dsr1-fp4-b200-trt:
   - isl: 1024
     osl: 1024
     search-space:
-    # If TP=4,
```
cquil11, line 836: kinda useful to have the comments for posterity
if the logic has changed, please update accordingly

benchmarks/dsr1_fp4_b200_trt.sh
```diff
@@ -22,39 +22,22 @@ echo "TP: $TP, CONC: $CONC, ISL: $ISL, OSL: $OSL, EP_SIZE: $EP_SIZE, DP_ATTENTIO
...
-        fi
+if [[ "$ISL" == "1024" && "$OSL" == "1024" ]]; then
+    if [[ "$TP" == "8" && "$EP_SIZE" == "8" ]]; then
+        PIECEWISE_CUDA_GRAPHS="true"
     fi
 fi
 
+if [[ "$DP_ATTENTION" == "true" ]]; then
+    MOE_BACKEND="CUTLASS"
+    CUDA_GRAPH_MAX_BATCH_SIZE=$(( $CONC / 4 ))
```
cquil11: nit: super edge case when CONC < 4

camiloamoreno: That's a fair point, though we likely wouldn't use DP ATTN in such a case. Fixed.

.github/configs/nvidia-master.yaml
camiloamoreno: done.

benchmarks/dsr1_fp4_b200_trt.sh
cquil11: true
thx anyway

cquil11:
@pr-claude review

cquil11 (approved):
lgtm once tests pass https://github.com/InferenceMAX/InferenceMAX/actions/runs/21611814777

kedarpotdar-nv (approved):

jgangani (approved):

PR #642, opened 2026-02-05
[NV] dsr1 fp4 b200 trt agg mtp update

camiloamoreno:
This PR contains below updates:

- Update to the latest TRTLLM 1.2 release container, the recent rc6.post3
- Fine-tune choice of parallelism in nvidia-master (reduce overlapping TP8/TP4)
- Enable Piecewise cuda graphs optimization on specific cases 

Near the top of the benchmark script (L26-35), we enable specific optimizations, mainly differentiating between cases with and without DP attention, including for the choice of MTP aggressiveness.

As in the non-mtp fp4 agg version, we use Piecewise Cuda Graphs (https://nvidia.github.io/TensorRT-LLM/features/torch_compile_and_piecewise_cuda_graph.html) which enables some components to execute thorugh cuda graphs while other components are run eagerly, to gain benefit with lower overhead. We use the formula from the documentation to generate a capture_num_tokens list depending on MAX_NUM_TOKENS. 

"cuda graph max batch size" is optimized to match CONC as a natural limit, but reduced to  batch_size/4 when DP_ATTENTION is enabled.

### DIFF
--- a/.github/configs/nvidia-master.yaml
+++ b/.github/configs/nvidia-master.yaml
@@ -865,7 +865,7 @@ dsr1-fp4-b200-trt:
     - { tp: 8, ep: 8, dp-attn: true, conc-start: 128, conc-end: 256 }
 
 dsr1-fp4-b200-trt-mtp:
-  image: nvcr.io#nvidia/tensorrt-llm/release:1.1.0rc2.post2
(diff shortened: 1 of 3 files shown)


kedarpotdar-nv (approved):
lgtm. this should get picked up after current GPTOSS run is complete

cquil11 (approved):

PR #645, opened 2026-02-05
[WIP][AMD] Include latest MI300X results

zhentaocc:


.github/configs/amd-master.yaml
```diff
@@ -47,7 +47,7 @@ dsr1-fp4-mi355x-atom:
     - { tp: 8, ep: 1, conc-start: 4, conc-end: 4 }
 
 dsr1-fp8-mi300x-sglang:
-  image: lmsysorg/sglang:v0.5.7-rocm700-mi30x
```
functionstackx, line 50: arent all single node via upstream besides this?

.github/workflows/benchmark-tmpl.yml
```diff
@@ -131,6 +139,88 @@ jobs:
           fetch-depth: 0
           ref: ${{ inputs.ref || github.ref }}
 
+      - name: Print ref
```
cquil11, line 142: Not sure if this is just for testing, but this definitely cannot be in the `benchmark-tmpl.yml`, if anything this should be in `dsr1_fp8_mi300x.sh`. But really shouldn't this just be in the image?

This feels off.

zhentaocc: Sorry for confusion, this is not a ready pr to be merged. I am trying to test if some optimizations actually work... I have to update sglang and aiter for better results.

functionstackx:
closing this PR since it isnt close to ready yet

PR #649, opened 2026-02-05
[AMD] fix: add --use-chat-template to mi355x fp8 disagg with mtp

cquil11:
Running MTP without the proper template induces unrealistic performance gains due to overly aggressive token “acceptance rate” (AR). See below for more details.

For benchmarking Multi-Token Prediction (MTP) with random data, it is necessary to apply the chat template. The chat template does not change the underlying token distribution or model behavior; it simply formats the random data into the JSON message structure defined by the OpenAI Chat Completions API, which is how DSR1 is post-trained to operate. Without this formatting, random token sequences can form artificially long or misstructured sequences, inflating acceptance lengths and Acceptance Rates (AR).

MT-Bench is a benchmark composed of real user-style prompts and reflects production-like input distributions. It therefore serves as a reliable reference point for expected acceptance behavior in real workloads.

We compared ARs from random data and MT-Bench, both with and without the chat template. Random data with the chat template produces ARs (~2.50) and acceptance rates that closely align with MT-Bench (overall AR ~2.50). In contrast, random data without the chat template shows an inflated AR (~3.55).

Because MT-Bench reflects real user prompt distributions, an AR around 2.5 clearly exists in production workloads. If benchmarking is done without the chat template, customers will not observe the same high TPS/GPU in real workloads. InferenceMAX benchmarks are intended to be reproducible and representative of production behavior, which requires using the chat template.

Currently, NVIDIA uses chat template, but AMD has removed it (see the benchmark args [here](https://github.com/billishyahao/sglang_disagg/blob/main/benchmark_lib.sh)). Please enable the chat template flag on all scenarios that utilize MTP specifically (as shown [here](https://github.com/InferenceMAX/InferenceMAX/blob/main/benchmarks/dsr1_fp8_b200_trt_mtp.sh)).


benchmarks/dsr1_fp8_mi355x_sglang-disagg.sh
```diff
@@ -28,11 +28,7 @@ fi
 
 set -x
 
-# Always clone and setup sglang_disagg
-# git clone --branch cam/sa-251219 https://github.com/cquil11/sglang_disagg.git
-
-# Switch to origin repo url for supporting wide ep configs
-git clone --branch cam/sa-260114 https://github.com/cquil11/sglang_disagg.git
+git clone --branch sa-260203 https://github.com/billishyahao/sglang_disagg.git
```
cquil11, line 31: Changes are downstream.

chunfangamd: I couldn't find the commit ID for adding "--use-chat-template" to the mi355x fp8 disagg with mtp from the main branch of the repo at https://github.com/billishyahao/sglang_disagg.git.

Otherwise, is it safer to introduce a deterministic behavior? Bill might continue his dev work on his repo. How do you think?

```
rm -rf sglang_disagg
git clone https://github.com/billishyahao/sglang_disagg.git
cd sglang_disagg
git checkout commit_ID
```

cquil11: make sure to checkout this branch `--branch sa-260203`

chunfangamd:
```
rm -rf sglang_disagg
git clone --branch sa-260203 https://github.com/billishyahao/sglang_disagg.git
cd sglang_disagg
git checkout 4eb2491b5c5f3667387ebf5c036004b65770be90
```

cquil11:
Perf difference looks neglibible. Merging.

PR #671, opened 2026-02-10
Profiling

Oseltamivir:
Allows for profiling for single node.
- Uses `merge_profiles` for both prefill and extend/decode
- Uploads trace to external github, and uses perfetto relay for link without need to download trace for viewing


functionstackx:
@Oseltamivir is this pr still wip? quite an large diff

Oseltamivir:
Apologies for delays due to multinode refactor. Easier to restart instead of merging main into this branch.

Continued with #787

PR #692, opened 2026-02-12
[AMD] Add MI355X disagg DSR1 FP4 DEP configs

billishyahao:


functionstackx (changes requested):
@billishyahao please add perf-changelog.yaml

functionstackx:
@billishyahao DP attention is failing so going to disable that for the time being

i filed an issue in sglang repo under #18749

PR #767, opened 2026-02-19
[NVIDIA] chore: add more gb200 runners

cquil11:
- Adds two additional GB200 runners to take advantage of SLURM oversubscription
- General cleanup of redundant code

Run here: https://github.com/SemiAnalysisAI/InferenceX/actions/runs/22203646777


Oseltamivir:
@claude PR review

Klaud-Cold:
**Claude finished @Oseltamivir's task in 1m 55s** —— [View job](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/22234700683)

---
### PR Review: Add more GB200 runners

- [x] Gather context and read changed files
- [x] Review `runners.yaml` changes
- [x] Review `benchmark-multinode-tmpl.yml` changes
- [x] Review `runners/launch_gb200-nv.sh` changes
- [x] Post review summary

---

### Summary

This PR enables SLURM oversubscription for GB200 runners by adding two new runners and reworking the launch script to support multiple concurrent jobs on the same cluster. The changes are well-structured and the log streaming improvement is a nice upgrade. A few items to flag:

---

### `.github/configs/runners.yaml` — Looks good

Adding `gb200-nv_1` and `gb200-nv_2` is straightforward. The naming is consistent with other runner groups.

---

### `.github/workflows/benchmark-multinode-tmpl.yml` — Looks good

**Slurm cleanup scope** (line 123): Extending the name-based `scancel` to `gb200-nv*` runners is the correct approach for oversubscription — each runner cancels only its own jobs (`--name="${{ runner.name }}"`) instead of all jobs for `$USER`. This is critical for multiple runners sharing a Slurm user account.

**Upload logs step** (lines 191-197): Good addition. Using `if-no-files-found: ignore` and `if: always()` ensures this never blocks the workflow. The artifact naming includes `${{ env.RESULT_FILENAME }}` which ensures uniqueness across parallel runs.

---

### `runners/launch_gb200-nv.sh` — Mostly good, one issue

**Positive changes:**

1. **`enroot import` without `srun`** (lines 50-51): Makes sense since the runner is already on the cluster with `enroot` available. Removes unnecessary Slurm overhead for image imports.

2. **Job name override via `sed`** (line 170): `sed -i "s/^name:.*/name: \"${RUNNER_NAME}\"/" "$CONFIG_FILE"` — This is the key piece that makes oversubscription work. Each runner stamps its name onto the Slurm job, pairing with the name-based cleanup in the workflow template. Runner names like `gb200-nv_0` have no sed-special characters, so this is safe.

3. **Log streaming with `tail -F`** (lines 206-219): This is a significant improvement over the old poll-and-print loop. The approach is:
   - Background process polls `squeue` for job completion
   - `tail -F -s 2 --pid=$POLL_PID` streams logs in real-time and exits when the poll process finishes
   - `-F` follows by name (handles NFS file recreation), `-s 2` sets poll interval (NFS doesn't support inotify)
   
   This gives much better CI output since logs stream in real-time instead of appearing all at once after completion.

4. **Log artifact collection** (lines 233-234): Archiving logs as a tarball and uploading as an artifact is better than dumping them to stdout. Pairs well with the new workflow upload step.

5. **`set +x` / `set -x` toggle** (lines 181, 221): Reduces log noise during the wait loop — good for readability.

6. **Removed debug lines and cleanup section**: The `echo "Make setup complete"` / `ls configs/` removal and the `.venv` cleanup removal are fine housekeeping.

**Issue to address:**

7. **Undefined `$jobid` in legacy path** (line 69):
   ```bash
   while [ -n "$(squeue -j "$jobid" --noheader)" ]; do
   ```
   The variable `$jobid` (lowercase) is never defined in this script. The benchmark script at line 66 runs in a subprocess (`bash "benchmarks/..."`) so any variables it sets are not visible here. The previous code used `squeue -u $USER` which, while broader, at least worked.
   
   This is in the legacy path (`FRAMEWORK == "dynamo-sglang" && -z "$CONFIG_FILE"`), and there's no matching gb200 sglang benchmark script in `benchmarks/multi_node/`, so this path may be effectively dead code. Still, if it's ever hit, this would loop zero times and skip straight through.
   
   **Suggestion**: Either fix `$jobid` to reference a defined variable, or add a comment that this path is deprecated / unreachable. [Fix this →](https://claude.ai/code?q=In%20runners%2Flaunch_gb200-nv.sh%20on%20branch%20chore%2Fadd-more-gb200-runners%2C%20line%2069%20references%20%24jobid%20which%20is%20never%20defined.%20Either%20replace%20it%20with%20a%20valid%20variable%20or%20add%20a%20comment%20noting%20this%20legacy%20path%20is%20unreachable.&repo=SemiAnalysisAI/InferenceX)

**Minor nit (non-blocking):**

8. **File existence check** (line 196): `while ! ls "$LOG_FILE" &>/dev/null` works but `[ ! -f "$LOG_FILE" ]` is more idiomatic for file existence checks.

---

### Overall

The PR achieves its goals cleanly — the oversubscription support is correctly implemented through the runner-name-based job naming + cleanup pattern. The log streaming is a solid UX improvement for CI. The one functional issue is the undefined `$jobid` in the legacy path, which is low-risk since that code path appears unreachable for GB200 today.

**Verdict**: Approve with the `$jobid` nit noted above.

---

functionstackx:
@cquil11 the validation run is failing

https://github.com/SemiAnalysisAI/InferenceX/actions/runs/22203646777/job/64321062017

cquil11:
re-running @functionstackx

PR #771, opened 2026-02-20
[NVIDIA] Enable multiple runners for b200-multinode-slurm

csahithi:


.github/workflows/benchmark-multinode-tmpl.yml
```diff
@@ -120,11 +120,11 @@ jobs:
...
-            if [[ "${{ runner.name }}" == mi355x-amds* ]]; then
-              echo "[Slurm] Cleaning up jobs with name: ${{ runner.name }} ..."
-              scancel --name="${{ runner.name }}" || true
-              while [ -n "$(squeue --name='${{ runner.name }}' --noheader --format='%i')" ]; do
-                squeue --name="${{ runner.name }}"
+            if [[ "${{ runner.name }}" == mi355x-amds* ]] || [[ "${{ runner.name }}" == b200-dgxc-slurm* ]]; then
+              echo "[Slurm] Cleaning up jobs with name: ${{ runner.name }} for user: $USER ..."
+              scancel -u "$USER" --name="${{ runner.name }}" || true
+              while [ -n "$(squeue -u "$USER" --name='${{ runner.name }}' --noheader --format='%i')" ]; do
+                squeue -u "$USER" --name="${{ runner.name }}"
```
functionstackx, line 128: @csahithi is each runner on an separate linux username or on the same linux username? 

on mi355, all 9 runners on are the same linux username this PR would break mi355.

+viz @cquil11

csahithi: This change wouldn't break the existing implementation. All the runners are under the same linux username for B200 as well. It just adds an additional check to make sure we are checking for jobs with the runner name only under the current user, just to ensure jobs run with the same name by any other user are not disturbed.

functionstackx: ahh gotcha! can u launch an test run to help validate that this works?

csahithi: sure! triggered a test run here - https://github.com/SemiAnalysisAI/InferenceX/actions/runs/22244986816

functionstackx:
thanks!

@cquil11 can help review & merge this PR

.github/workflows/benchmark-multinode-tmpl.yml
csahithi: @functionstackx @cquil11 the test run is completed

cquil11:
@csahithi I rebased to [chore/add-more-nv-runners](https://github.com/SemiAnalysisAI/InferenceX/tree/chore/add-more-nv-runners)
SemiAnalysisAI/InferenceX:chore/add-more-nv-runners which is the branch I am currently working on in order to add more runners across all NVIDIA clusters. Will merge into that. Thank you!

PR #787, opened 2026-02-24
Profiling support

Oseltamivir:
Allows for profiling for single node.
- Uses merge_profiles for both prefill and extend/decode
- Uploads trace to external github, and uses perfetto relay for link without need to download trace for viewing

Continuing #671


Oseltamivir:
@claude 
Review this PR. Ignore `on: push` trigger, it will be removed later before merging. Double check that no directories are created in `inferenceX/`.

Klaud-Cold:
**Claude finished @Oseltamivir's task in 4m 48s** —— [View job](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/22371248049)

---
### PR Review: Profiling Support

- [x] Gather context and read all changed files
- [x] Check that no directories are created in InferenceX/
- [x] Review workflow file `profile-sglang.yml`
- [x] Review `benchmarks/benchmark_lib.sh` changes
- [x] Review runner launch script changes
- [x] Review `benchmark_serving.py` changes
- [x] Post comprehensive review feedback

---

### Directory Creation Check

**No persistent directories are created in the InferenceX repo.** All `mkdir` operations are ephemeral, running only during CI:

| Location | What's created | Where | Persists? |
|---|---|---|---|
| `benchmark_lib.sh:243` | `mkdir -p "$_prof_dir"` | Inside Docker container (`/workspace/`) | No |
| `profile-sglang.yml:244` | `path: storage` checkout | CI runner workspace (external repo checkout) | No |
| `profile-sglang.yml:259` | `mkdir -p "$dest_dir"` | Inside the `storage/` checkout (external repo) | No |

The `profiles/` directory referenced at `profile-sglang.yml:182` is created by SGLang's profiler inside the Docker container and mapped to the CI runner workspace — also ephemeral.

---

### Review Findings

#### 1. Bug: Error message references non-existent input `seq-lens`

`profile-sglang.yml:98`:
```
echo "No entries produced for config-key=${{ inputs.config-key }}, seq-lens=${{ inputs.seq-lens }}, conc=${{ inputs.conc }}."
```
There is no `seq-lens` input defined in the workflow. This will render as an empty string. Should just be removed from the error message.

[Fix this →](https://claude.ai/code?q=In%20.github%2Fworkflows%2Fprofile-sglang.yml%20line%2098%2C%20remove%20the%20%60seq-lens%3D%24%7B%7B%20inputs.seq-lens%20%7D%7D%60%20reference%20from%20the%20error%20message%20since%20there%20is%20no%20seq-lens%20input%20defined%20in%20this%20workflow.%20Branch%3A%20add-profiling-support&repo=SemiAnalysisAI/InferenceX)

#### 2. Missing profiling env vars on `launch_mi355x-amds.sh`

The push trigger default targets `dsr1-fp4-mi355x-sglang`, which would route to `launch_mi355x-amds.sh`. This runner does **not** pass `PROFILE`, `SGLANG_TORCH_PROFILER_DIR`, `VLLM_TORCH_PROFILER_DIR`, or `VLLM_RPC_TIMEOUT` through to the container/srun environment. Profiling would silently be a no-op on MI355X.

Only 4 of 18 runner scripts have the profiling env vars:
- `launch_b200-dgxc.sh` ✅
- `launch_h100-cr.sh` ✅  
- `launch_mi300x-amd.sh` ✅
- `launch_mi300x-cr.sh` ✅

This is fine if profiling is intentionally scoped to these runners only, but worth documenting or noting since the push trigger default config targets a runner without profiling support.

#### 3. `permissions: contents: read` may be insufficient

`profile-sglang.yml:33-34`:
```yaml
permissions:
  contents: read
```
The workflow uses `actions/upload-artifact` (which needs `actions: write` implicitly through GITHUB_TOKEN) and checks out an external repo via SSH deploy key. The `contents: read` permission should be fine for the artifact uploads (handled by the actions runtime), but worth verifying this works end-to-end since the permissions are restrictive.

#### 4. `move_profile_trace_for_relay` has a 3-minute busy-wait

`benchmark_lib.sh:392-398`:
```bash
local wait_attempts=180
for (( i=1; i<=wait_attempts; i++ )); do
    trace_file="$(_find_latest_profile_trace "${search_dirs[@]}")"
    if [[ -n "$trace_file" ]]; then
        break
    fi
    sleep 1
done
```
This polls every second for up to 180 seconds (3 minutes). Since `num_steps` is set to `1`, the trace should appear quickly. However, the `_find_latest_profile_trace` function calls `find` on each iteration, which is reasonable. Just noting the 3-minute upper bound — might want a log line when the wait begins so it doesn't appear to hang silently.

#### 5. `_find_latest_profile_trace` filename filter could be tighter

`benchmark_lib.sh:347-350`:
```bash
find "${search_roots[@]}" -maxdepth 1 -type f \
    \( -name "*.trace.json" -o -name "*.trace.json.gz" -o -name "*trace*.json" -o -name "*trace*.json.gz" -o -name "*profile*.json" -o -name "*profile*.json.gz" \) \
    -print0 2>/dev/null
```

The glob patterns `*trace*.json` and `*profile*.json` are quite broad. In theory, any file with "trace" or "profile" anywhere in the name and a `.json` extension would match. In practice this is likely fine since the search is limited to the profiler output directories, but if other tools drop JSON files with these substrings, false positives could occur.

#### 6. `benchmark_serving.py` change looks correct

The `extra_body={"num_steps": 1, "merge_profiles": True, "profile_by_stage": True}` addition on `start_profile` is the right approach. These parameters are confirmed to be accepted by SGLang's `/start_profile` endpoint (0.5.7-0.5.8). The `extra_body` dict is merged into the request payload at `backend_request_func.py:260-261`.

One note: `num_steps: 1` means only 1 decode step is profiled. This is intentional for lightweight single-step traces, but users should be aware this captures a very narrow window of execution.

#### 7. Workflow name `profile-sglang.yml` vs actual scope

The workflow file is named `profile-sglang.yml` but it also handles vLLM (`VLLM_TORCH_PROFILER_DIR` is set, vLLM profiler dir env is passed). Consider renaming to `profile.yml` to avoid confusion, or document that vLLM profiling is also supported.

#### 8. `storage` checkout creates a nested repo

`profile-sglang.yml:239-247`: The `Oseltamivir/profiler-storage` repo is checked out into `storage/` within the workspace. This works, but if the cleanup step at the beginning of subsequent runs doesn't clean this up, stale `storage/` directories could accumulate. The `sudo find . -not -user $USER -exec rm -rf {} +` cleanup at the top should handle this.

#### 9. Minor: hardcoded `/workspace/` fallback in `move_profile_trace_for_relay`

`benchmark_lib.sh:374`:
```bash
for dir in "$sglang_dir" "$vllm_dir" "/workspace"; do
```
The `/workspace` hardcoded fallback is always added as a search directory. Since both `sglang_dir` and `vllm_dir` default to `/workspace` when unset (line 368-369), this will typically be a no-op. But if both profiler dirs are set to non-`/workspace` paths, `/workspace` still gets searched, which is arguably the right thing since traces might end up there regardless.

---

### Summary

The PR is well-structured and the approach is sound. The main actionable items are:

1. **Bug fix needed**: Remove the `seq-lens` reference from the error message (item 1)
2. **Consider**: Either add profiling env vars to `launch_mi355x-amds.sh` or change the push trigger default to target a runner that has profiling support (item 2)
3. **Nice-to-have**: Add a log line when the profile trace wait loop begins (item 4)

Everything else looks good — the `extra_body` approach for `merge_profiles`/`profile_by_stage` is clean, the trace upload flow to the external storage repo is well-implemented, and the Perfetto relay link generation is a nice touch.

---
 |

.github/workflows/profile.yml
```diff
@@ -0,0 +1,293 @@
...
+        with:
+          name: moe_debug_${{ env.RESULT_FILENAME }}
+          path: "moe_debug.tp0.log"
+          if-no-files-found: ignore
+
+      - name: Checkout storage repo
+        if: ${{ steps.run.outputs.trace != '' }}
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        with:
+          repository: Oseltamivir/profiler-storage
```
functionstackx: @Oseltamivir any chance that u can store in SemiAnalysisAI repo?

i.e. `SemiAnalysisAI/InferenceX-trace-storage` or something?

Oseltamivir:
Ready for merge, just need to remove the `on: push` trigger.

Should only be used for sglang for now, vllm traces cannot do `num_step`, will profile everything, resulting in >2GB trace.

functionstackx (approved):
LGTM besides 1 small nit.

 can u update `.github/workflows/claude.yml` and test it out such that someone can do like "@ claude profile sglang b200 deepseek fp4 conc=4 and send me the perfetto relay link" and it would work?

Oseltamivir:
https://github.com/SemiAnalysisAI/InferenceX/issues/797 tested @functionstackx

PR #866, opened 2026-03-05
Add ingest-to-db job to workflow for run results

adibarra:
Extends run-sweep workflow to load results into our new database.

### DIFF
--- a/.github/workflows/run-sweep.yml
+++ b/.github/workflows/run-sweep.yml
@@ -289,6 +289,19 @@ jobs:
             (needs.sweep-multi-node-8k1k.result == 'success' || needs.sweep-multi-node-8k1k.result == 'skipped')
         runs-on: ubuntu-latest
         steps:
+            - name: Download results artifacts
+              uses: actions/download-artifact@70fc10c6e5e1ce46ad2ea6f2b72d43f7d47b13c3 # v8.0.0
+              with:
+                  path: artifacts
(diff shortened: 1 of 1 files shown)


.github/workflows/run-sweep.yml
```diff
@@ -289,6 +289,19 @@ jobs:
             (needs.sweep-multi-node-8k1k.result == 'success' || needs.sweep-multi-node-8k1k.result == 'skipped')
         runs-on: ubuntu-latest
         steps:
+            - name: Download results artifacts
+              uses: actions/download-artifact@70fc10c6e5e1ce46ad2ea6f2b72d43f7d47b13c3 # v8.0.0
+              with:
+                  path: artifacts
+                  pattern: "{results_bmk,run-stats,eval_results_all,changelog-metadata}"
+            - name: Ingest results to DB
+              uses: SemiAnalysisAI/inferencemax-app/packages/ingest-action@main # latest only
```
cquil11, line 298: super duper edge case
but to have back of mind, this opens up a bit of a vulnerability when we open source frontend since this isn't pinned to a specific commit
will just need to be careful nothing malicious ends up in this action

cquil11 (approved):

.github/workflows/run-sweep.yml
adibarra: It does use a few actions internally but those are pinned in the front end repo. But yeah, gotta keep an eye on that.

PR #911, opened 2026-03-15
Separate eval-only workflow and change to 8k1k

Oseltamivir:
## Summary                                                 
- Remove 1k8k                          
- Decouple eval jobs from benchmark jobs into a dedicated `eval /` workflow stage, so evals can run independently with `evals-only: true` in perf-changelog
- Add `eval-only` input to benchmark-tmpl.yml that skips throughput benchmarks and result file checks, running only lm_eval
- Fix GSM8k prompt template: change `#### <answer>` to `#### [number]` to prevent models (e.g. Kimi-K2.5) from outputting the literal `<answer>` tag, which caused strict-match to return `[invalid]`
- Add eval support to all 49 single-node benchmark scripts (server restart with native max context, eval context override, request timeout increase)
- Update `process_changelog.py` and `generate_sweep_configs.py` to emit a separate `evals` matrix for eval-only jobs
- Changed from 1k8k to 8k1k
- Groups now include dp-attn (previously missing, so dp-attn variants were merged)
- Instead of picking max-conc at highest/lowest TP, it picks all TPs at the highest conc + all TPs at the median conc.

Separate jobs on a [normal run](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/23324769392?pr=911) tested with:
```
- config-keys:
    - gptoss-fp4-h100-vllm
    - dsr1-fp8-mi325x-sglang
```

---                              
  ## Implementation Details                                                                          
                                                                                                              
  - `get_native_max_context_length(model)` — reads `max_position_embeddings` /  `max_sequence_length` / `seq_length` / `n_positions` from the model's HuggingFace config. 
  - `max_output_tokens`: dynamically computed as `eval_context_len - 4096` (capped at   
  16384), was fixed 8192
  - Score validation: new `validate_scores.py` — post-eval gate checking `exact_match` metrics ≥ 0.85, exits non-zero on failure or zero metrics found
  - Failure mode: eval-only uses `if-no-files-found: error` (was `ignore`)
  - GSM8k prompt: `#### <answer>` → `#### [number]` 
  - ATOM torchvision workaround: uninstalls `torchvision` before lm-eval install to avoid circular import                                                                     
  - `PYTHONDONTWRITEBYTECODE=1` added to workflow env to avoid EACCES error on pycache.
  
Screenshot 2026-03-21 at 9 05 29 PM

New job step: evals

Screenshot 2026-03-27 at 10 40 34 PM

Normal job with perf-changelog, no 1k8k


functionstackx (changes requested):
thanks for the PR @Oseltamivir ! seems great that we now parallel evals instead of having them in sequential 

mostly LGTM

just had a couple of questions
1. does this PR change the evals to use the industry standard num of shots for gsm8k?
2. previously when evals hard crashed, the script still exits with success exit code, is this fixed now?

Oseltamivir:
1. Changed to 5-shot
2. Yes, also added suggestion of failing when result <85%

Oseltamivir:
Separate jobs on a full run: https://github.com/SemiAnalysisAI/InferenceX/actions/runs/23120451287/job/67153421071?pr=911

functionstackx:
lgtm, can u get @cquil11 to review too?

functionstackx:
@claude review plz

Klaud-Cold:
**Claude finished @functionstackx's task in 3m 42s** —— [View job](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/23129429585)

---
### PR Review: Separate eval-only workflow and change to 8k1k

- [x] Read key changed files and gather context
- [x] Review workflow changes (benchmark-tmpl, e2e-tests, run-sweep)
- [x] Review benchmark_lib.sh changes
- [x] Review benchmark scripts (all 49 single-node)
- [x] Review matrix logic (generate_sweep_configs, process_changelog)
- [x] Review runner launch scripts
- [x] Review eval config and docs
- [x] Post final review

---

**Overall: Looks good.** This is a well-structured PR that cleanly separates eval jobs from benchmark jobs, fixes the GSM8k prompt, and adds eval support across all 49 benchmark scripts. The architecture is sound — the `EVAL_ONLY` / `RUN_EVAL` separation is clean, `process_changelog.py` properly generates separate benchmark and eval matrices, and the 85% accuracy threshold in eval-only mode is a good addition.

I verified that the issues from the previous automated review have all been addressed in the current revision:
- `EVAL_ONLY` is correctly forwarded in all 4 Docker launch scripts (`-e EVAL_ONLY`)
- `--concurrent-requests $CONC` has been removed from all benchmark scripts (0 matches)
- Pattern B scripts (vLLM/TRT) correctly `export EVAL_MAX_MODEL_LEN` after the subshell call
- The SINGLE filter in `e2e-tests.yml` no longer excludes eval-marked entries

### Remaining issues

**1. Median concurrency selection for 2-element lists** — `generate_sweep_configs.py:76`

```python
median_conc = conc_values[len(conc_values) // 2]
```

For groups with exactly 2 distinct concurrency values `[a, b]`, `len // 2 = 1`, so `median_conc = b = highest`. This means `target_concs = {b, b} = {b}` — only one concurrency level gets eval coverage. The AGENTS.md documents this ("When only 2 concurrency levels exist, this collapses to just the highest"), so it's acknowledged. But if the intent is two distinct levels for diversity, use `(len(conc_values) - 1) // 2` to select the lower-middle element.

[Fix this →](https://claude.ai/code?q=In%20utils%2Fmatrix_logic%2Fgenerate_sweep_configs.py%20line%2076%2C%20change%20%60conc_values%5Blen(conc_values)%20%2F%2F%202%5D%60%20to%20%60conc_values%5B(len(conc_values)%20-%201)%20%2F%2F%202%5D%60%20and%20update%20AGENTS.md%20line%20306%20to%20remove%20the%20caveat%20about%202-element%20collapse&repo=SemiAnalysisAI/InferenceX)

**2. `compute_eval_context_length` export is redundant** — `benchmark_lib.sh:616`

The function both `export`s `EVAL_MAX_MODEL_LEN` and `echo`s the value. Since every caller either (a) captures the echo in a subshell (losing the export) and then manually exports, or (b) calls via `run_eval()` which redirects to `/dev/null` — the `export` inside the function never actually takes effect in the parent shell. Consider removing the `export` from the function to avoid confusion about where the variable is actually set.

**3. Eval failure in combined mode (non-EVAL_ONLY) is silent** — `benchmark_lib.sh:764-769`

When `RUN_EVAL=true` (combined mode) and the eval fails, `run_eval()` logs an error but does not exit with failure — the benchmark job still succeeds. This means eval regressions in combined mode can go unnoticed. The 85% threshold check only runs in the eval-only workflow path (`benchmark-tmpl.yml:162-181`), not in the combined `RUN_EVAL=true` path. This is probably intentional (don't fail expensive benchmark jobs over eval issues), but worth being explicit about.

**4. `EVAL_MAX_MODEL_LEN` used as `max_length` in lm_eval** — `benchmark_lib.sh:624,661`

`gen_max_tokens` (from `EVAL_MAX_MODEL_LEN`, defaulting to 16384) is passed as `max_length` to lm_eval, which controls the maximum total input+output context window. For eval-only mode with expanded context (e.g., 47080 tokens for DeepSeek), this means lm_eval is told it can use up to 47080 tokens of context — but GSM8K prompts are short, so this is fine in practice. Just noting that if longer eval tasks are added, the `max_gen_tokens=16384` hard cap (line 651) will limit per-generation output independently.

### Looks good

- GSM8k prompt fix (`#### [number]` instead of `#### <answer>`) with 5-shot — good fix for the strict-match `[invalid]` issue
- `evals-only: true` in `perf-changelog.yaml` — clean separation for the production workflow
- 85% accuracy threshold in eval-only mode with clear pass/fail logging
- `run_benchmark_serving` correctly skips in eval-only mode
- All scripts consistently handle eval context expansion with the correct pattern for their framework (SGLang: `EVAL_CONTEXT_ARGS`, vLLM/TRT: `MAX_MODEL_LEN` override + export)
- `process_changelog.py` generates completely separate benchmark and eval arrays — no cross-contamination

---
 |

.github/workflows/benchmark-tmpl.yml
```diff
@@ -123,13 +131,34 @@ jobs:
...
+          sudo rm -rf /it-share/gharunners/gharunner7/actions-runner/_work/InferenceX/InferenceX/utils/bench_serving/__pycache__ 2>/dev/null || true
 
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           token: ${{ secrets.REPO_PAT }}
           fetch-depth: 0
           ref: ${{ inputs.ref || github.ref }}
+          clean: false
+
+      - name: Workspace cleanup (generated files)
```
cquil11: Why is this needed all the sudden? It wasn't needed before right?

.github/workflows/benchmark-tmpl.yml
```diff
@@ -145,28 +174,42 @@ jobs:
...
+            # Verify eval produced results
+            if ! ls results*.json 1>/dev/null 2>&1; then
+              echo "Eval-only run failed: no results*.json files found." >&2
+              exit 1
             fi
-            echo "Waiting for result file... (attempt $i)"
-            sleep 1
-          done
+            # Verify eval scores meet minimum threshold (85%)
+            python3 utils/evals/validate_scores.py
```
cquil11, line 166: So you're just hardcoding 85% as the global minimum?
What happens when we extend to benchmarks where 85% isn't necessarily achievable?
Is it better to just leave it up to developers to observe and make sure the benchmark scores are acceptable?

perf-changelog.yaml
```diff
@@ -994,3 +994,56 @@
...
+    - kimik2.5-int4-mi355x-vllm
+    - minimaxm2.5-fp8-mi325x-vllm
+    - minimaxm2.5-fp8-mi355x-vllm
+    - qwen3.5-bf16-mi325x-sglang
+    - qwen3.5-bf16-mi355x-sglang
+    - qwen3.5-fp8-mi325x-sglang
+    - qwen3.5-fp8-mi355x-sglang
+  description:
+    - "Separate evals, change to 8k1k, fail loudly, 5-shot, top of curve & middle of curve"
+  pr-link: https://github.com/SemiAnalysisAI/InferenceX/pull/911
```
cquil11, line 1190: add evals-only here?

cquil11: also, how are we gonna go about ingesting only the eval results here instead of the perf as well?

.github/workflows/profile.yml
```diff
@@ -153,6 +155,22 @@ jobs:
...
           ref: ${{ inputs.ref || github.ref }}
+          clean: false
+
+      - name: Workspace cleanup (generated files)
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          if command -v sudo >/dev/null 2>&1 && sudo -n true 2>/dev/null; then
+            sudo find "$GITHUB_WORKSPACE" -type d \( -name '__pycache__' -o -name '.pytest_cache' \) -prune -exec rm -rf {} + 2>/dev/null || true
```
cquil11: You can't use `sudo` outside of specific `runners/` scripts where the gharunner user is known to have sudo
The default user does not have sudo on all clusters and this can lead to a silent error of the runner

cquil11 (changes requested):

functionstackx (commented):
.

.github/workflows/benchmark-tmpl.yml
functionstackx: this is the absolute minium threshold for gsm8k needed to merge. in fact, this is an fairly low gsm8k score threshold. this will allow us to stop wastes 30 hours of @Oseltamivir 's time per week like he been doing for the past 4 weeks.

fair point that different benchmarks have different min. thresholds, it should be configurable per eval & per model.

Oseltamivir:
@claude

Review this PR. Ignore earlier resolved issues, even if not changed like only 1 conc being ran if there are only 2 to sweep.

Klaud-Cold:
**Claude finished @Oseltamivir's task in 2m 11s** —— [View job](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/23678842026)

---
### PR Review: Separate eval-only workflow and change to 8k1k

- [x] Read key files and gather context
- [x] Review benchmark_lib.sh changes
- [x] Review workflow changes (benchmark-tmpl, e2e-tests, run-sweep)
- [x] Review benchmark scripts
- [x] Review Python utility changes (generate_sweep_configs, process_changelog, validate_scores)
- [x] Review documentation and config changes
- [x] Post final review

---

**Overall: Well-structured PR.** The eval-only workflow separation is clean, `process_changelog.py` correctly generates separate benchmark/eval matrices with independent dedup, `validate_scores.py` properly fails on zero metrics checked, and all Docker launch scripts correctly forward `EVAL_ONLY`. Most previously flagged issues have been addressed. Two critical issues remain.

### Critical Issues

**1. `compute_eval_context_length` multiplier is `1` instead of `5`** — `benchmarks/benchmark_lib.sh:624`

```bash
local eval_ctx=$(( benchmark_ctx * 1 ))
```

The function's own docstring (line 609) says "Uses 5x the benchmark context capped at the model's native max." With `* 1`, the eval context is never expanded — for a typical 8k1k workload, the server gets 9416 tokens instead of ~47080. This makes eval-only mode's core feature (expanded context) a complete no-op across all 49 scripts. This was flagged multiple times in prior review rounds and marked resolved, but **the code was never actually changed**.

[Fix this →](https://claude.ai/code?q=In%20benchmarks%2Fbenchmark_lib.sh%20line%20624%2C%20change%20%60benchmark_ctx%20*%201%60%20to%20%60benchmark_ctx%20*%205%60&repo=SemiAnalysisAI/InferenceX&branch=eval-8k1k-server-restart)

**2. Stray destructive `rm -rf` in `dsr1_fp8_mi355x_atom_mtp.sh`** — lines 87-88

```bash
set -x
rm -rf ./utils/bench_serving\
```

After the script's normal ending (`stop_gpu_monitor` / `set +x` at line 85), these two stray lines execute unconditionally. The trailing backslash is a line continuation — bash resolves this as `rm -rf ./utils/bench_serving`, deleting the benchmark serving utilities directory from the workspace on every run. Compare with the sibling `dsr1_fp4_mi355x_atom_mtp.sh` which ends cleanly at `set +x`. This was also flagged previously and marked resolved but remains in the code.

[Fix this →](https://claude.ai/code?q=In%20benchmarks%2Fsingle_node%2Fdsr1_fp8_mi355x_atom_mtp.sh%2C%20remove%20lines%2087-89%20(the%20stray%20%60set%20-x%60%20and%20%60rm%20-rf%20.%2Futils%2Fbench_serving%5C%60%20lines%20after%20the%20script's%20normal%20ending)&repo=SemiAnalysisAI/InferenceX&branch=eval-8k1k-server-restart)

### Minor Issues

**3. Cleanup step doesn't remove `sample*.jsonl`** — `benchmark-tmpl.yml:227-230`

The cleanup step removes `meta_env.json` and `results*.json` but not `sample*.jsonl`, which `append_lm_eval_summary` also moves into the workspace CWD and which is included in the artifact upload. On persistent self-hosted runners, these accumulate and contaminate future artifact uploads.

```yaml
# Add this line:
rm -f sample*.jsonl || true
```

**4. `collect-evals` in `run-sweep.yml` runs when `sweep-evals` is skipped** — line 209

```yaml
if: ${{ always() && needs.setup.result != 'skipped' }}
```

When a PR produces no 8k1k eval configs, `sweep-evals` is skipped, no `eval_*` artifacts exist, but `collect-evals` still runs. `actions/download-artifact@v8` will fail when no artifacts match the pattern. Same issue in `e2e-tests.yml:171-173` where `collect-evals` uses `if: ${{ always() }}`.

Fix: add `&& needs.sweep-evals.result != 'skipped'` in `run-sweep.yml` and `&& needs.test-sweep-evals.result != 'skipped'` in `e2e-tests.yml`.

**5. `e2e-tests.yml` SINGLE filter includes eval entries** — line 57

When `--evals-only` is passed, `generate_sweep_configs.py` outputs only entries with `run-eval: True`. The SINGLE filter `[x for x in d if 'prefill' not in x]` captures all of them, so every config runs as both a full benchmark (via `test-sweep-single-node`) and an eval-only run (via `test-sweep-evals`), wasting GPU time. Adding `and not x.get('run-eval', False)` to the SINGLE filter would fix this.

### Looks Good

- GSM8k prompt fix (`#### [number]` instead of `#### <answer>`) with 5-shot — good fix
- `validate_scores.py` correctly returns 1 when `checked == 0` (line 86) — the zero-metrics silent pass issue is properly handled
- `EVAL_ONLY` correctly forwarded in all 4 Docker launch scripts (`-e EVAL_ONLY` on line 36 of `launch_b200-dgxc.sh`)
- All `--concurrent-requests $CONC` removed from benchmark scripts (0 grep matches)
- `process_changelog.py` uses separate `benchmark_configs_seen` / `eval_configs_seen` sets (lines 97-98) — no cross-contamination
- `setup_eval_context()` wrapper (line 640) correctly avoids the subshell export issue by calling `compute_eval_context_length` directly and exporting afterward
- Artifact naming collision addressed with `eval_server_logs_` / `eval_gpu_metrics_` conditional prefixes (lines 202, 210)
- `always()` guard correctly added to eval results upload (line 215) and cleanup step (line 226)
- `thresholds.json` with per-task thresholds (gsm8k: 0.85, gpqa: 0.30) — sensible approach
- EVALS.md and AGENTS.md documentation updated to reflect new selection policy
- `glm5_fp8_h200.sh` has the EVAL_ONLY block and `$EVAL_CONTEXT_ARGS` correctly wired up

---
 |

PR #912, opened 2026-03-15
Fix disaggregated per-GPU IO throughput comparability via cluster averages

### LINKED ISSUE #299: Presented input/output token throughput per GPU for disaggregated setups not usefully comparable to standard multi-gpu
**Describe the bug**
The figure for input or output token throughput per GPU for disaggregated isn't _wrong_ as such, but I don't think it's the correct figure to plot alongside the same metric for standard multi-GPU setups. For disaggregated setups the reported input/output throughput appears to be the figure _per prefill GPU_ and _per decode GPU_ respectively, which is why the throughput per GPU isn't the sum of the output and input throughput figures. The graph tooltip doesn't provide the context of how many GPUs are dedicated to prefill vs decode, and I think the value that is appropriate for plotting alongside the non-disaggregated setups would be to calculate the per-GPU output/input throughput by summing the throughput of the respective workers and dividing by the _total_ GPUs. i.e. the output/input throughput per GPU average across the whole cluster rather than just averaged over the GPUs dedicated for prefill/decode. If I'm trying to compare and see which setup is better (and how much better), it's the per-GPU figure averaged across the whole GPU count that matters.

**To Reproduce**
1. Go to inferencemax.semianalysis.com
2. View results including a disaggregated config, e.g. R1 FP8 1k/1k
3. Switch the y-axis metric between the different token throughput metrics

**Expected behavior**
See issue introduction.

But specifically looking at e.g. the results for R1 fp8 gb200 nvl72 conc=4096 1k/1k results. We currently get:
* total per gpu
  * 3822.27
* output per gpu
  * 2867.344
* input per gpu
  * 5732.173

I believe the more comparable metric would be:
 * total per gpu
   * 3822.27
* output per gpu
  * 2867.344 * (48/72) = 1911.56
* input per gpu
  * 5732.173 * (24/72) = 1910.72

Now just like the non-disaggregated setups, total throughput per gpu is the sum of the output and input per GPU totals, and is representative of the average across the whole cluster. In the ideal case, the tooltip for the input/output per GPU figures for disaggregated datapoints would have a footnote that clarifies how it was averaged, such as:

```
Input token throughput per GPU*: 1911.56

*: Averaged across the whole 72-node cluster
```

For completeness, let's look at a datapoint from a non-disaggregated setup. Taking the mi325x conc=64 figures for example:
  * mi325x conc=64
    * total per gpu
      * 412.874 (= sum of output and input per GPU)
    * output per gpu
      * 206.483
    * input per gpu
      * 206.39

**Screenshots**

Image

Image

Image

**Desktop (please complete the following information):**
 N/A

KOKOSde:
Fixes #299 by making disaggregated input/output throughput per GPU cluster-averaged (comparable with aggregated runs), while preserving role-specific decode/prefill throughput metrics for deeper analysis.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -276,3 +276,7 @@ python utils/summarize.py <results_directory>
 ```
 
 Outputs GitHub-flavored markdown tables with metrics including TTFT, TPOT, interactivity, E2EL, and throughput per GPU for both single-node and multi-node results.
+
+For multi-node/disaggregated results:
+- `TPUT per GPU`, `Output TPUT per GPU`, and `Input TPUT per GPU` are cluster averages (divided by total GPUs in the run).
+- `Output TPUT per Decode GPU` and `Input TPUT per Prefill GPU` are also emitted for role-specific analysis.
--- a/utils/process_result.py
+++ b/utils/process_result.py
@@ -4,6 +4,14 @@ import os
 from pathlib import Path
 
 
+def safe_division(numerator, denominator, metric_name):
+    """Divide safely and raise a clear error on invalid denominators."""
+    if denominator <= 0:
+        raise ValueError(
+            f"{metric_name} requires a positive denominator, got {denominator}.")
+    return numerator / denominator
+
+
 def get_required_env_vars(required_vars):
     """Load and validate required environment variables."""
     env_values = {}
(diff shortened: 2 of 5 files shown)


cquil11:
After some discussion, we don't think this really makes sense. Although I see where you're coming from, it still makes sense to consider the averages of prefill/decode respective to the GPUs actually doing their work

I.e., consider a 1P5D deployment. It feels hard to justify normalizing prefill throughput by all 5D GPUs despite these GPUs contributing no useful work to prefill throughput.

PR #917, opened 2026-03-18
test

functionstackx:


functionstackx:
@billishyahao counterintuitively, although theoretical modelling shows that AMD's MTP=2/3 should performance between than AMD's MTP=1, in the current stack, AMD's MTP=3 performs worse than MTP=1. But since ur boss is saying AMD should go back to MTP=2/3, we can change it to that... 


image

PR #919, opened 2026-03-19
Fix nvidia-master.yaml metadata to align with recipe files

nlevin-ui:
Audit found 34 metadata mismatches across 412 entries in nvidia-master.yaml where declared TP/EP/DP-attn values didn't match the actual recipe files.

Fixes applied:
- B200 FP8 TRT 8K/1K: 15 prefill_ep (8->1), 1 prefill_dp_attn (true->false)
- H200 FP8 TRT: 17 prefill_dp_attn mismatches
  - 1K/1K: false->true (9 entries)
  - 8K/1K: true->false (8 entries)
- GB200 FP4 TRT 8K/1K: 2 decode_dp_attn (true->false)

All changes are metadata-only in nvidia-master.yaml to reflect actual recipe values. No recipe files were modified.

### DIFF
--- a/.github/configs/nvidia-master.yaml
+++ b/.github/configs/nvidia-master.yaml
@@ -615,7 +615,7 @@ dsr1-fp8-b200-dynamo-trt:
       prefill:
         num-worker: 1
         tp: 8
-        ep: 8
+        ep: 1
         dp-attn: true
         additional-settings:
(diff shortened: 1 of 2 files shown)


functionstackx (approved):
thanks for the fix

cquil11 (approved):
LGTM
Thanks for the fix

PR #930, opened 2026-03-23
[AMD] fix: add --exclusive to MI300X SLURM salloc for accurate benchmarks

cquil11:
## Summary
- Add `--exclusive` flag to `salloc` in `runners/launch_mi300x-amds.sh` to prevent node sharing during benchmarks
- Only non-TP8 configs are affected (TP8 already uses all GPUs); perf-changelog updated for `gptoss-fp4-mi300x-vllm` and `minimaxm2.5-fp8-mi300x-vllm`
- NVIDIA runners already use `--exclusive`; this brings MI300X to parity

## Test plan
- [ ] Run a non-TP8 MI300X benchmark (e.g. gptoss-fp4-mi300x-vllm TP4) and verify results are consistent with exclusive node access

### DIFF
--- a/perf-changelog.yaml
+++ b/perf-changelog.yaml
@@ -1,4 +1,12 @@
 - config-keys:
+    - gptoss-fp4-mi300x-vllm
+    - minimaxm2.5-fp8-mi300x-vllm
+  description:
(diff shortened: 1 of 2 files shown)


cquil11:
note needed
oversubscriptin is disabled by default

PR #929, opened 2026-03-23
[AMD] Fix slurm command for AMD devices

seungrokj:
Hi @cquil11 @functionstackx 

Motivation is AMD slurm clusters are missing **--exclusive** option and they showed regressed performance. NV slurm clusters are already using **--exclusive** option. 

without --exclusive
https://github.com/SemiAnalysisAI/InferenceX/actions/runs/23433485252
image

with --exclusive 
https://github.com/SemiAnalysisAI/InferenceX/actions/runs/23439928217
image

Regards, 
Seungrok

PR #932, opened 2026-03-23
[AMD] fix: add --exclusive to MI355X multi-node sbatch for accurate benchmarks

cquil11:
## Summary
- Add `--exclusive` flag to `sbatch` in `benchmarks/multi_node/amd_utils/submit.sh` to prevent node sharing during multi-node MI355X benchmarks
- Only non-TP8 configs are affected (TP8 already uses all GPUs); perf-changelog updated for all 11 non-TP8 MI355X configs
- Single-node MI355X `salloc` already has `--exclusive`; this brings the multi-node path to parity

## Test plan
- [ ] Run a non-TP8 MI355X multi-node disagg benchmark (e.g. dsr1-fp8-mi355x-sglang-disagg TP4) and verify results are consistent with exclusive node access

### DIFF
--- a/benchmarks/multi_node/amd_utils/submit.sh
+++ b/benchmarks/multi_node/amd_utils/submit.sh
@@ -129,6 +129,7 @@ fi
 sbatch_cmd=(
     sbatch
     --parsable
+    --exclusive
     -N "$NUM_NODES"
     -n "$NUM_NODES"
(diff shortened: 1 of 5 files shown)


cquil11:
Closing in favor of #929 which already covers MI355X.

PR #934, opened 2026-03-23
[AMD] Fix slurm command for AMD devices

cquil11:
> **Note:** Re-opening of #929 which was reverted.

Hi @cquil11 @functionstackx 

Motivation is AMD slurm clusters are missing **--exclusive** option and they showed regressed performance. NV slurm clusters are already using **--exclusive** option. 

without --exclusive
https://github.com/SemiAnalysisAI/InferenceX/actions/runs/23433485252
image

with --exclusive 
https://github.com/SemiAnalysisAI/InferenceX/actions/runs/23439928217
image

Regards, 
Seungrok


cquil11:
@seungrokj had to re open

perf-changelog.yaml
lq1131, line 1: Good catch Claude!!

PR #937, opened 2026-03-24
Fix tokenizer mismatch between benchmark client and sglang server 

ZhaiFeiyue:
### Problem
When benchmarking sglang server performance with the latest sgalng (which uses transformers 5.3.0), we observed a significant performance regression compared to the older `0313-2` image (transformers 4.57.1):

- **TTFT**: 1394ms vs 482ms (~3x slower)
- **Output throughput**: 160 tok/s vs 212 tok/s (~24% drop)

### Root Cause
Transformers v5 changed how `LlamaTokenizerFast` is loaded. During `__init__`, v5 rebuilds the `pre_tokenizer` and `decoder` from scratch using Llama-specific components, discarding the originals defined in `tokenizer.json`. For models like DeepSeek-R1 that declare `LlamaTokenizerFast` but actually use a ByteLevel/Sequence tokenizer architecture, v5 incorrectly replaces:

- `pre_tokenizer`: `Sequence` → `Metaspace`
- `decoder`: `ByteLevel` → `Sequence`

The sglang server already fixes this at startup via `_fix_v5_tokenizer_components()` in `hf_transformers_utils.py`, but the benchmark client loads the tokenizer directly via `AutoTokenizer.from_pretrained()` without these fixes. This mismatch causes the client and server to tokenize text differently:

| | Client (unfixed) | Server (fixed) |
|---|---|---|
| pre_tokenizer | Metaspace | Sequence |
| 100 random tokens → text → re-encode | 102 tokens | **531 tokens** |
| 7140 token prompt | 7140 tokens | **35921 tokens** |

The ~5x token inflation on the server side results in 5x more prefill work, which directly causes the observed TTFT and throughput regression.

### Fix
Add `_fix_tokenizer_for_sglang()` to the benchmark client's `get_tokenizer()`, applying the same two fixes that sglang server applies:

1. **pre_tokenizer/decoder restoration**: Read the original components from `tokenizer.json` and overwrite the incorrect v5 replacements.
2. **add_bos_token recovery**: Restore `add_bos_token`/`add_eos_token` flags from `tokenizer_config.json` that v5 strips during loading.

This is a no-op on transformers v4.

### Results

**Token count (per request, ~7400 token input):**

| | Before fix | After fix |
|---|---|---|
| Client-side token count | 7140 | 7140 |
| Server-side prompt_tokens | 35921 (5x inflated) | 7141 (correct) |
| Prefill chunks | 3 (16384+16384+3153) | 1 (7141) |

**Benchmark (10 prompts, concurrency=1, input ~8K, output ~1K, `--disable-radix-cache`):**

| Metric | Before fix | After fix | Old image (baseline) |
|---|---|---|---|
| Output throughput (tok/s) | 160.49 | **211.71** | 212.00 |
| Mean TTFT (ms) | 1394.84 | **215.00** | 215.26 |
| Mean TPOT (ms) | 4.74 | **4.49** | 4.26 |
| Mean E2EL (ms) | 5796.15 | **4393.74** | 4387.54 |

After fix, new image performance matches old image — **no actual regression**.

### DIFF
--- a/utils/bench_serving/backend_request_func.py
+++ b/utils/bench_serving/backend_request_func.py
@@ -449,6 +449,7 @@ def _fix_tokenizer_for_sglang(tokenizer, model_path):
     LlamaTokenizerFast but actually use a ByteLevel/Sequence tokenizer
     architecture, v5 incorrectly replaces the original Sequence pre_tokenizer
     with Metaspace, and the original ByteLevel decoder with Sequence.
+    See: https://github.com/sgl-project/sglang/blob/9238bd08a2895fa3b7ec79ea567e5c27ac951343/python/sglang/srt/utils/hf_transformers_utils.py#L836
 
     The sglang server applies fixes for this in hf_transformers_utils.py
     (_fix_v5_tokenizer_components and _fix_v5_add_bos_eos_token), but the
--- a/utils/bench_serving/backend_request_func.py
+++ b/utils/bench_serving/backend_request_func.py
@@ -439,6 +439,75 @@ def get_model(pretrained_model_name_or_path: str) -> str:
     return pretrained_model_name_or_path
 
 
+def _fix_tokenizer_for_sglang(tokenizer, model_path):
+    """Fix transformers v5 tokenizer to match sglang server-side behavior.
+
+    Root cause: transformers v5 (>= 5.0) changed how tokenizers are loaded.
+    Specifically, LlamaTokenizerFast.__init__ in v5 rebuilds the pre_tokenizer
+    and decoder from scratch using class-specific components, discarding the
+    originals from tokenizer.json. For models like DeepSeek-R1 that declare
+    LlamaTokenizerFast but actually use a ByteLevel/Sequence tokenizer
+    architecture, v5 incorrectly replaces the original Sequence pre_tokenizer
+    with Metaspace, and the original ByteLevel decoder with Sequence.
+
+    The sglang server applies fixes for this in hf_transformers_utils.py
+    (_fix_v5_tokenizer_components and _fix_v5_add_bos_eos_token), but the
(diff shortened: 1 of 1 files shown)


functionstackx:
thanks @ZhaiFeiyue is for all models in sglang or just for glm5?

cquil11:
@ZhaiFeiyue Thank you for identifying this error and for the fix. +1 to functionstackx comment. It will be helpful to identify which configs from amd-master.yaml and nvidia-master.yaml are affected by this so that we can re-run and get accurate perf across the board. 

+viz @kedarpotdar-nv

ZhaiFeiyue:
> thanks @ZhaiFeiyue is for all models in sglang or just for glm5?

hi @functionstackx  we only found this issue for deepseek-R1, not do any test for glm5

billishyahao:
> @ZhaiFeiyue Thank you for identifying this error and for the fix. +1 to functionstackx comment. It will be helpful to identify which configs from amd-master.yaml and nvidia-master.yaml are affected by this so that we can re-run and get accurate perf across the board.
> 
> +viz @kedarpotdar-nv

Hi @cquil11 now no configs are affected by the patch. Only after we upgrade the sglang to latest main, then we will observe the issue. So it is a proactive protection.

cquil11:
@claude Review

Klaud-Cold:
Claude Code is working… 

I'll analyze this and get back to you.

[View job run](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/23650213739)

cquil11 (approved):
OK. I see that the same change has indeed been applied in SGLang upstream, https://github.com/sgl-project/sglang/blob/9238bd08a2895fa3b7ec79ea567e5c27ac951343/python/sglang/srt/utils/hf_transformers_utils.py#L836

LGTM

PR #983, opened 2026-03-31
[blocker, AMD said they would switch to upstream after end of march, this PR is still not upstream images] [AMD] improve dsr1 fp4 disagg perf on mi355x

billishyahao:
The new patch is adding the following optimization:
- "Bump SGL mori image to March 27"
- "Add more low latency sweep configs"
- "Enable v2 mxfp4 DSR1 0528 model"
- "Enable fp4 disp feature on mori"


.github/configs/amd-master.yaml
```diff
@@ -504,7 +486,7 @@ dsr1-fp8-mi355x-atom-mtp:
     - { tp: 8, conc-start: 4, conc-end: 256, spec-decoding: mtp  }
 
 dsr1-fp8-mi355x-sglang-disagg:
-  image: rocm/sgl-dev:sglang-0.5.9-rocm720-mi35x-mori-0227-2
+  image: rocm/sgl-dev:sglang-0.5.9-rocm720-mi35x-mori-0327
```
functionstackx, line 489: hi @billishyahao 

in early march, you said that after consulting with @haishaw and others in the org that by End of March, you would be using upstream images. Can u please update this use upstream nightly images instead of second class forks?

lets ensure that we work towards amd being an first class platform on sglang instead of continuing to submit second class forks

functionstackx:
superceded by https://github.com/SemiAnalysisAI/InferenceX/pull/1236

PR #997, opened 2026-04-02
Rename GLM-5 config and script from nvfp4 to fp4

Ankur-singh:
## Summary
- Rename config key `glm5-nvfp4-b200-sglang` → `glm5-fp4-b200-sglang` in `nvidia-master.yaml`
- Rename benchmark script `glm5_nvfp4_b200.sh` → `glm5_fp4_b200.sh`

The config uses `precision: fp4` but benchmark script is called `glm5_nvfp4_b200.sh`. The launch script resolves benchmark scripts using the precision value, so the mismatch would cause it to fail to find the correct script.

cc @functionstackx 

## Test plan
- [ ] Verify the launch script correctly resolves `glm5_fp4_b200.sh` for the `glm5-fp4-b200-sglang` config


functionstackx (approved):
lgtm. ty for the fix

kedarpotdar-nv (approved):

jgangani (approved):

PR #1000, opened 2026-04-03
Multinode evals

Oseltamivir:
## Summary
Add eval-only support for multi-node benchmarks and wire those eval results into CI collection + summary reporting.

This covers:
  - eval matrix selection for multi-node configs
  - eval-only workflow jobs for multi-node sweeps
  - AMD MI355X eval execution in `server.sh`
  - NVIDIA Slurm eval execution through Oseltamivir's `srt-slurm` fork
  - eval artifact upload, score validation, and multi-node-aware summary tables

## How evals are run
Single-node evals are selected on `8k1k` at max + median concurrency for each `(model, runner, framework, precision, spec-decoding, dp-attn)` group.

Multi-node evals are selected on `8k1k` by taking the entry with the highest max concurrency for each `(model, runner, framework, precision, spec-decoding, prefill-dp-attn, decode-dp-attn)` group, then running eval at the median concurrency from that config via `eval-conc`.

  `EVAL_ONLY=true` starts the server with expanded eval context, skips throughput benchmarking, runs `lm-eval`,
  writes `meta_env.json` + `results*.json` + `sample*.jsonl`, uploads those artifacts, then validates scores
  against thresholds.

## srt-slurm fork delta vs upstream
  NVIDIA multinode eval uses `Oseltamivir/srt-slurm@sa-submission-q1-2026` instead of `ishandhanani/srt-slurm`.

  Compared with current `upstream/main`, that fork adds the eval path InferenceX needs:
  - a new `lm-eval` benchmark runner
  - `/infmax-workspace` mounting via `INFMAX_WORKSPACE`
  - `EVAL_ONLY` support in `do_sweep.py` to skip benchmark stage and run post-eval directly
  - full `wait_for_model()` health checking before eval in eval-only mode
  - pass-through of framework/model/topology/env metadata into the eval container
  - `MODEL_NAME=self.config.served_model_name` so eval queries the served alias, not the HF repo id
  - `EVAL_CONC` from workflow to `EVAL_CONCURRENT_REQUESTS`
  - copying eval outputs into `/logs/eval_results/` for launcher-side artifact pickup

### Validation
https://github.com/SemiAnalysisAI/InferenceX/actions/runs/24059388771

### InferenceX PR
https://github.com/nvidia/srt-slurm/pull/12


functionstackx:
@Oseltamivir can u upstream ur changes to `srt-slurm`

PR #1007, opened 2026-04-06
[experimental] Add multinode profiling workflow

hbarclay:


patches/trtllm-profiling.py
```diff
@@ -0,0 +1,100 @@
...
+        print(f"[trtllm-patch] ERROR: HandlerBase has no generate_locally. "
+              f"Available: {methods}", file=sys.stderr, flush=True)
+        return
+
+    os.makedirs(output_dir, exist_ok=True)
+
+    _orig_generate_locally = HandlerBase.generate_locally
+    _state = {"step": 0, "started": False, "stopped": False}
+
+    _state["profiler"] = torch.profiler.profile(
```
briancoutinho: One idea is you can also consider adding a schedule with start /stop step and 1 warmup iteratipm
https://docs.pytorch.org/docs/stable/profiler.html#torch.profiler.schedule

Reference in Megatron Core
https://github.com/NVIDIA/Megatron-LM/blob/ab43d43f0bc04f4656d4af15afb6e7e4c9ad71c8/megatron/training/training.py#L2846-L2861

patches/trtllm-profiling.py
```diff
@@ -0,0 +1,100 @@
...
+        activities=[
+            torch.profiler.ProfilerActivity.CPU,
+            torch.profiler.ProfilerActivity.CUDA,
+        ],
+        record_shapes=True,
+        with_stack=False,
+        on_trace_ready=torch.profiler.tensorboard_trace_handler(output_dir),
+    )
+
+    async def _patched_generate_locally(self, request, context, embeddings=None):
```
briancoutinho, line 46: If you use torch profiler with schedule you can add a call to
_state["profiler"].step() on every iteration.

This allows you to configure the start and stop automatically.
It will also add a "Step<N>" marker in the trace that might be useful

patches/trtllm-profiling.py
```diff
@@ -0,0 +1,100 @@
...
+    _state = {"step": 0, "started": False, "stopped": False}
+
+    _state["profiler"] = torch.profiler.profile(
+        activities=[
+            torch.profiler.ProfilerActivity.CPU,
+            torch.profiler.ProfilerActivity.CUDA,
+        ],
+        record_shapes=True,
+        with_stack=False,
+        on_trace_ready=torch.profiler.tensorboard_trace_handler(output_dir),
```
briancoutinho: Not important but I typically pass in my own handler so I can control the file naming, default one does not include rank in the name.
See this reference in Megatron Core
https://github.com/NVIDIA/Megatron-LM/blob/ab43d43f0bc04f4656d4af15afb6e7e4c9ad71c8/megatron/training/training.py#L2846-L2861

patches/trtllm-profiling.py
briancoutinho: https://github.com/NVIDIA/Megatron-LM/blob/ab43d43f0bc04f4656d4af15afb6e7e4c9ad71c8/megatron/training/training.py#L2899-L2900

briancoutinho (commented):
Added some suggestions for the profiler patch, lgtm otherwise

cquil11:
Closing due to inactivity.

PR #1020, opened 2026-04-10
Fix e2e-tests dropping highest concurrency benchmark configs

Ankur-singh:
## Problem

The `e2e-tests.yml` workflow silently drops benchmark runs for the highest and median concurrency configs on 8k1k sequence lengths.

### Root cause

`mark_eval_entries()` marks selected entries (highest + median conc for 8k1k) with `run-eval=True`. The SINGLE job filter then **excludes** these entries:

```python
# Old filter — excludes run-eval=True from benchmarks
SINGLE = [x for x in d if 'prefill' not in x and not x.get('run-eval', False)]
```

These excluded configs only appear in the EVALS job, which runs with `eval-only: true` — meaning **no throughput benchmark is ever executed** for them.

### How run-sweep handles this differently

`run-sweep.yml` uses `process_changelog.py` which calls `generate_sweep_configs.py test-config` with `--no-evals` for benchmarks and `--evals-only` for evals **in separate passes**. This means ALL concurrency values are always benchmarked, and evals are generated independently.

In contrast, `e2e-tests.yml` generates everything in one pass and then splits — causing the eval-marked entries to be carved out of the benchmark list entirely.

### Impact

Across all NVIDIA + AMD configs, **172 benchmark configs** were being silently skipped. For example, with `minimaxm2.5-fp8-b200-vllm`, these 8k1k configs were never benchmarked:
- `tp=2 conc=64` (median)
- `tp=2 conc=512` (highest)
- `tp=4 conc=64` (median)
- `tp=4 conc=512` (highest)

## Fix

Remove the `run-eval` filter from the SINGLE config generation. Since benchmark and eval jobs are separate (SINGLE hardcodes `run-eval: false`, EVALS hardcodes `run-eval: true` + `eval-only: true`), there is no need to exclude eval-marked entries from benchmarks:

```python
# New filter — all single-node configs are benchmarked
SINGLE = [x for x in d if 'prefill' not in x]
```

Eval-marked entries now run both benchmark (via SINGLE) and eval (via EVALS), which is the correct behavior.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -54,7 +54,7 @@ jobs:
                   pip install pydantic
                   CONFIG_JSON=$(python3 ${GITHUB_WORKSPACE}/utils/matrix_logic/generate_sweep_configs.py \
                     ${{ inputs.generate-cli-command || github.event.inputs.generate-cli-command }})
-                  SINGLE=$(echo "$CONFIG_JSON" | python3 -c "import sys,json; d=json.load(sys.stdin); s=[x for x in d if 'prefill' not in x]; print(json.dumps([] if s and all(x.get('run-eval',False) for x in s) else s))")
+                  SINGLE=$(echo "$CONFIG_JSON" | python3 -c "import sys,json; d=json.load(sys.stdin); print(json.dumps([x for x in d if 'prefill' not in x and not x.get('eval-only', False)]))")
                   MULTI=$(echo "$CONFIG_JSON" | python3 -c "import sys,json; d=json.load(sys.stdin); print(json.dumps([x for x in d if 'prefill' in x]))")
                   EVALS=$(echo "$CONFIG_JSON" | python3 -c "import sys,json; d=json.load(sys.stdin); print(json.dumps([x for x in d if 'prefill' not in x and x.get('run-eval', False)]))")
                   echo "single-node-config=$SINGLE" >> $GITHUB_OUTPUT
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -54,7 +54,7 @@ jobs:
                   pip install pydantic
                   CONFIG_JSON=$(python3 ${GITHUB_WORKSPACE}/utils/matrix_logic/generate_sweep_configs.py \
                     ${{ inputs.generate-cli-command || github.event.inputs.generate-cli-command }})
-                  SINGLE=$(echo "$CONFIG_JSON" | python3 -c "import sys,json; d=json.load(sys.stdin); print(json.dumps([x for x in d if 'prefill' not in x]))")
+                  SINGLE=$(echo "$CONFIG_JSON" | python3 -c "import sys,json; d=json.load(sys.stdin); s=[x for x in d if 'prefill' not in x]; print(json.dumps([] if s and all(x.get('run-eval',False) for x in s) else s))")
                   MULTI=$(echo "$CONFIG_JSON" | python3 -c "import sys,json; d=json.load(sys.stdin); print(json.dumps([x for x in d if 'prefill' in x]))")
                   EVALS=$(echo "$CONFIG_JSON" | python3 -c "import sys,json; d=json.load(sys.stdin); print(json.dumps([x for x in d if 'prefill' not in x and x.get('run-eval', False)]))")
                   echo "single-node-config=$SINGLE" >> $GITHUB_OUTPUT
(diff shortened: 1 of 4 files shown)


Oseltamivir:
@Ankur-singh Thanks for the catch, having look rn. Will patch ASAP if indeed broken

Oseltamivir:
@Ankur-singh 

Thanks for catch once again. That was added because `--evals-only` causes `mark_eval_entries()` line 936 to filter to only eval-marked entries. `CONFIG_JSON` contains only entries with `run-eval=True` but `SINGLE`: all non-prefill entries → runs benchmarks for them (run-eval hardcoded false) `EVALS`: same entries → runs evals. This is wrong as `--evals-only` should skip benchmarks, but now they run too.

The intention is:
| Flags | mark_eval_entries() runs? | Effect |
|-------|---------------------------|--------|
| Default (no flags) | Yes | Both evals and throughput should run. This SHOULD happen during `perf-changelog`. **Bug that this PR fixes**: eval-marked 8k1k configs are excluded from SINGLE and only appear in EVALS with eval-only: true. No throughput benchmark for them. |
| --no-evals | No | No entries get run-eval=True, so the SINGLE filter drops nothing. All configs are benchmarked. No evals run. |
| --evals-only | Yes | Entries without run-eval=True are filtered out entirely (line 936). Only eval-marked configs survive, and they land in EVALS with eval-only: true. Benchmark-only configs SHOULD be discarded **Bug that this introduces and that line intended to fix**: SINGLE: all non-prefill entries → runs benchmarks for them even though `--evals-only` should skip benchmarks |


In default mode, the old filter `not x.get('run-eval', False)` excluded eval-marked entries from SINGLE. The new logic checks whether all non-prefill entries are eval-marked. In default mode they aren't (only highest/median concurrency 8k1k configs are marked), so all(...) is False and SINGLE includes everything, fixing the original bug.                                                                                                              
                                                                                                                    
In `--evals-only` mode, `generate_sweep_configs.py` filters to only eval-marked entries, so every non-prefill entry has `run-eval=True`. `all(...)` is True, SINGLE becomes `[]`, and no benchmarks run, preserving the intended behavior and preventing the regression.

---

Quick test:
```
python3 -c "
import json

def pr_fix_no_guard(d):
    return [x for x in d if 'prefill' not in x]

def our_fix(d):
    s = [x for x in d if 'prefill' not in x]
    return [] if s and all(x.get('run-eval', False) for x in s) else s

def original(d):
    return [x for x in d if 'prefill' not in x and not x.get('run-eval', False)]

evals_only_data = [
    {'name': 'c', 'conc': 256, 'run-eval': True},
    {'name': 'd', 'conc': 512, 'run-eval': True},
]

default_data = [
    {'name': 'a', 'conc': 64},
    {'name': 'b', 'conc': 128},
    {'name': 'c', 'conc': 256, 'run-eval': True},
    {'name': 'd', 'conc': 512, 'run-eval': True},
]

print('=== --evals-only mode (expect SINGLE=0, no benchmarks) ===')
print(f'  main (original):      SINGLE={len(original(evals_only_data))} entries')
print(f'  PR fix (no guard):    SINGLE={len(pr_fix_no_guard(evals_only_data))} entries  <-- REGRESSION: benchmarks run')
print(f'  Our fix (with guard): SINGLE={len(our_fix(evals_only_data))} entries')
print()
print('=== Default mode (expect SINGLE=4, all configs benchmarked) ===')
print(f'  main (original):      SINGLE={len(original(default_data))} entries  <-- BUG: eval-marked configs dropped') 
print(f'  PR fix (no guard):    SINGLE={len(pr_fix_no_guard(default_data))} entries')
print(f'  Our fix (with guard): SINGLE={len(our_fix(default_data))} entries')
"
```

Output:
```
=== --evals-only mode (expect SINGLE=0, no benchmarks) ===
  main (original):      SINGLE=0 entries
  PR fix (no guard):    SINGLE=2 entries  <-- REGRESSION: benchmarks run
  Our fix (with guard): SINGLE=0 entries

=== Default mode (expect SINGLE=4, all configs benchmarked) ===
  main (original):      SINGLE=2 entries  <-- BUG: eval-marked configs dropped
  PR fix (no guard):    SINGLE=4 entries
  Our fix (with guard): SINGLE=4 entries
```

Ankur-singh:
@Oseltamivir that makes sense. tysm!

Oseltamivir:
@Ankur-singh Lol my fix has another bug where it fails when a config has only 8k1k with a single concurrency value, because `mark_eval_entries` marks every entry, making it indistinguishable from `--evals-only` mode.

I'll just stamp `EVAL_ONLY` and make the tests more stringent

Oseltamivir (approved):
lgtm

Oseltamivir:
@Ankur-singh I'll merge this in tmr morning after looking at it _again_ when I'm combobulated. Please help me to look through it one more time too.

Thanks so much 🙏🫡

Ankur-singh:
Sorry about that 🫣
Please revert!

Oseltamivir:
Lol it's fine, probably good now with stamp of `EVAL_ONLY`, but will still have a look tmr and PR if smt wrong again

Ankur-singh:
We are in this together 😂🤞

PR #1021, opened 2026-04-10
Revert "Fix e2e-tests dropping highest concurrency benchmark configs"

Ankur-singh:
Reverts SemiAnalysisAI/InferenceX#1020


Oseltamivir:
@Ankur-singh Took another look and I believe the current fix should works.

Thanks for catching it early! 🫡

PR #1037, opened 2026-04-16
Locks python dependencies for CI into a requirements.txt file and cache the dependencies

n0madsky:
## Summary

Consolidate the CI-side Python dependency installs for `utils/` into a shared `utils/requirements.txt` file, and update the affected workflows to install from that file instead of using scattered ad-hoc `pip install` commands.

This also makes the Python setup more consistent across the touched workflows by adding `actions/setup-python` and pip caching where appropriate, broadens the targeted test workflow triggers so requirements changes are exercised, and cleans up a few simple workflow shellcheck issues.

## Changes

- Rewrite `utils/requirements.txt` to use direct exact pins for the shared CI/test dependencies
- Replace scattered workflow `pip install ...` usage with `pip install -r utils/requirements.txt`
- Add `actions/setup-python` to affected `ubuntu-latest` jobs that run Python or pip
- Add pip caching using `utils/requirements.txt` as the dependency path
- Update `test-process-result.yml` and `test-matrix-logic.yml` to trigger when `utils/requirements.txt` changes
- Clean up a few simple shellcheck-style issues in workflow `run:` steps

## Validation

- Verified the exact pins install and import successfully in `python:3.12-slim`
- Ran the Docker-based `matrix_logic` pytest suite successfully
- `actionlint` runs and reports only remaining shellcheck findings in existing complex workflow shell blocks

### DIFF
--- a/.github/workflows/collect-evals.yml
+++ b/.github/workflows/collect-evals.yml
@@ -27,12 +27,24 @@ jobs:
           path: eval_results/
           pattern: ${{ inputs.result-prefix && format('eval_{0}_*', inputs.result-prefix) || 'eval_*' }}
 
+      - name: Set up Python
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        with:
+          python-version: '3.12'
+          cache: 'pip'
+          cache-dependency-path: utils/requirements.txt
+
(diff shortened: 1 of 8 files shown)


n0madsky:
@cquil11 any reason for the close? If the code is not up to standard feel free to leave a review

PR #1095, opened 2026-04-19
fix: remove dead import and redundant entry dicts in generate_sweep_configs

aryanputta:
## Summary

Two dead code issues in `utils/matrix_logic/generate_sweep_configs.py`:

- **Unused import**: `from ast import For` — `For` is an AST node class that's never referenced in this file. Left over from an earlier draft, importing it silently pulls in `ast` for no reason.
- **Redundant dict construction**: In both the multinode and single-node branches of `generate_full_sweep()`, an `entry` dict was built and immediately thrown away because the very next line starts a `for runner_value in runners_for_entry:` loop that rebuilds an identical dict. The outer construction was dead code — it was never validated or appended. Removed it and hoisted `seq_len_str` / `runners_for_entry` above the loop where they belong.

## Test plan

- [ ] `python3 -c "from utils.matrix_logic.generate_sweep_configs import generate_full_sweep"` — no ImportError
- [ ] Existing unit tests pass: `python3 -m pytest utils/matrix_logic/`
- [ ] Output of `generate_full_sweep` is identical before and after (behavior unchanged, only dead code removed)

### DIFF
--- a/utils/matrix_logic/generate_sweep_configs.py
+++ b/utils/matrix_logic/generate_sweep_configs.py
@@ -1,4 +1,3 @@
-from ast import For
 import fnmatch
 import json
 import argparse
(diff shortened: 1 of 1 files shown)


Oseltamivir (commented):
no impact PR, merged with #1096

PR #1047, opened 2026-04-17
[NV] Update Kimi K2.5 FP4 B200 vLLM server launch args

Ankur-singh:
## Summary

Update vLLM server launch arguments for the Kimi K2.5 FP4 B200 benchmark to improve serving performance.

### Changes

**`benchmarks/single_node/kimik2.5_fp4_b200.sh`**
- Add `--kv-cache-dtype fp8` to use FP8 KV cache for reduced memory usage and improved throughput
- Add `--max-cudagraph-capture-size 2048` to enable CUDA graph capture for sequences up to 2048
- Add `--max-num-batched-tokens "$((ISL * 2))"` to set batched token limit based on input sequence length
- Add `--stream-interval 20` to reduce streaming overhead

**`perf-changelog.yaml`**
- Add changelog entry for the `kimik2.5-fp4-b200-vllm` config documenting the new server args

PR #1100, opened 2026-04-20
Add B300 config: kimi-k2.5-fp4-vllm

cquil11:
> [!WARNING]
> This PR is a reopen of #1056, which was reverted in #1099 due to an error with the first PR. The contents and description are otherwise identical to the original.

## Summary
- Add `kimik2.5-fp4-b300-vllm` benchmark config and the corresponding `benchmarks/single_node/kimik2.5_fp4_b300.sh` launch script
- At the time of submission, [the vLLM Kimi-K2.5 recipes page](https://docs.vllm.ai/projects/recipes/en/latest/moonshotai/Kimi-K2.5.html) does not have a B300-specific recipe, so this reuses the existing Kimi-K2.5 FP4 (NVFP4) B200 vLLM recipe as-is until B300-specific tuning is available
- Image: `vllm/vllm-openai:v0.17.0` (same as B200), runner: `b300`, same TP/EP/concurrency search-space as B200

## Test plan
- [ ] CI config validation passes
- [ ] Run `kimik2.5-fp4-b300-vllm` single-node benchmark on a B300 node and confirm server starts, benchmark completes, and result file is produced

### DIFF
--- a/.github/configs/nvidia-master.yaml
+++ b/.github/configs/nvidia-master.yaml
@@ -2237,6 +2237,29 @@ kimik2.5-fp4-b200-vllm:
     - { tp: 8, ep: 1, conc-start: 4, conc-end: 4 }
     - { tp: 4, ep: 1, conc-start: 4, conc-end: 64 }
 
+# NOTE: At the time of submission, https://docs.vllm.ai/projects/recipes/en/latest/moonshotai/Kimi-K2.5.html
+# does not have a B300-specific recipe, so this config reuses the existing
+# Kimi-K2.5 FP4 B200 vLLM recipe as-is until B300-specific tuning is available.
+kimik2.5-fp4-b300-vllm:
+  image: vllm/vllm-openai:v0.19.0-cu130
+  model: nvidia/Kimi-K2.5-NVFP4
+  model-prefix: kimik2.5
+  runner: b300
+  precision: fp4
+  framework: vllm
+  multinode: false
+  seq-len-configs:
+  - isl: 1024
+    osl: 1024
+    search-space:
+    - { tp: 8, ep: 1, conc-start: 4, conc-end: 4 }
+    - { tp: 4, ep: 1, conc-start: 4, conc-end: 64 }
+  - isl: 8192
+    osl: 1024
+    search-space:
+    - { tp: 8, ep: 1, conc-start: 4, conc-end: 4 }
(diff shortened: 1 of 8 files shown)


xinli-sw:
Hi @cquil11 , I think we could do two things to get better perf

1. B300 could benefit from more parallel configs, we can run some more and report back
2. We should also include changes from https://github.com/SemiAnalysisAI/InferenceX/pull/1047/changes

Depending on how fast you want the data, we can either have #1407 as a fast follow of this PR or just wait until that's ready. Thanks!

cquil11:
@xinli-sw No real rush, we can wait for y'all.

functionstackx:
hi @xinli-sw it would be if https://recipes.vllm.ai/ had b300 specific configs so that ppl can know what flags to use specifically for b300

xinli-sw:
yup for sure, I think mostly just having better parallel configs so that B300's higher mem size, will update shortly

functionstackx:
@claude look at the fix in https://github.com/SemiAnalysisAI/InferenceX/pull/1104 for b200, add it to this b300 PR too

Klaud-Cold:
**Claude finished @functionstackx's task in 2m 14s** —— [View job](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/24693858324)

---
### Applying B200 fix from PR #1104 to this B300 PR

- [x] Look at PR #1104 diff to understand the B200 fix
- [x] Identify what needs to change in this B300 PR
- [x] Verify current state

### Result

The fix from PR #1104 (bump max-model-len headroom from +200 to +256 for block-size alignment) is **already included** in this branch via commit `69a2584`:

| File | Status |
|------|--------|
| `benchmarks/benchmark_lib.sh` (`setup_eval_context`) | +256 |
| `benchmarks/single_node/gptoss_fp4_b200.sh` | +256 |
| `benchmarks/single_node/gptoss_fp4_h200.sh` | +256 |
| `utils/matrix_logic/generate_sweep_configs.py` (4 sites) | +256 |
| `utils/matrix_logic/test_generate_sweep_configs.py` | +256 |

The B300 kimi script (`kimik2.5_fp4_b300.sh`) itself doesn't need a direct change because it receives `MAX_MODEL_LEN` from the environment variable set by `generate_sweep_configs.py`, which already uses the +256 headroom. No additional changes needed.
 |

PR #1101, opened 2026-04-20
[Experimental] Add timing instrumentation to serial prompt generation

Oseltamivir:
## Summary
- Adds `Prompt generation completed in Xs` timing output to the serial prompt generation path
- Enables direct comparison against PR #1038's parallel path timing in sweep logs

## Test plan
- Run `/sweep` with H200 dsr1 8k1k config and compare logs against PR #1038's sweep

### DIFF
--- a/utils/bench_serving/benchmark_serving.py
+++ b/utils/bench_serving/benchmark_serving.py
@@ -121,6 +121,8 @@ def sample_random_requests(
 
     input_requests = []
     mismatches = []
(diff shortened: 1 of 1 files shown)


Oseltamivir:
/sweep full-sweep --config-files .github/configs/amd-master.yaml --model-prefix gptoss --runner-type mi355x --seq-lens 8k1k --no-evals --framework vllm --max-tp 1

PR #1102, opened 2026-04-20
feat: improve benchmark serving to work better for small models

cquil11:


functionstackx:
only applies to small models with small ISL/OSL, doesnt seem to have much effect on large agentic workloads https://github.com/SemiAnalysisAI/InferenceX/blob/main/utils/bench_serving/KNOWN_LIMITATION.md

PR #1104, opened 2026-04-20
Bump max-model-len headroom from +200 to +256 (block-size alignment)

Ankur-singh:
## Summary

Some attention backends require the KV-cache block count to be a multiple of 128 tokens. With the default `block_size=32` that means `block_num` must be divisible by `128 / 32 = 4`. The current `ISL + OSL + 200` padding doesn't guarantee this alignment — for example at `ISL=8192`, `OSL=1024` the context length rounds to 295 blocks which is not divisible by 4:

\`\`\`
RuntimeError: Worker failed with error 'Expected block_num % (128 / block_size) == 0, got block_num=295 and block_size=32', please check the stack trace above for the root cause
\`\`\`

Bumping the headroom to **+256** (any multiple of 128 — even `+128` would suffice) keeps the context length aligned and avoids this crash. Also propagated to `setup_eval_context` so eval runs use the same aligned context.

## Changes

- \`benchmarks/benchmark_lib.sh\` — eval context padding (\`setup_eval_context\`)
- \`benchmarks/single_node/gptoss_fp4_b200.sh\` — the 8k-sequence branch
- \`benchmarks/single_node/gptoss_fp4_h200.sh\` — same
- \`utils/matrix_logic/generate_sweep_configs.py\` — all four \`MAX_MODEL_LEN\` computation sites
- \`utils/matrix_logic/test_generate_sweep_configs.py\` — docstring + expected-value assertion

## Test plan

- [x] \`pytest utils/matrix_logic/test_generate_sweep_configs.py -k max_model_len\` passes with the new value
- [ ] Trigger an e2e sweep on an affected backend (e.g. qwen3.5-fp4-b200-trt 8k/1k) to confirm the \`block_num\` error no longer fires

### DIFF
--- a/.github/configs/amd-master.yaml
+++ b/.github/configs/amd-master.yaml
@@ -530,14 +530,14 @@ gptoss-fp4-mi300x-vllm:
     - { tp: 1, conc-start: 64, conc-end: 256 }
     - { tp: 2, conc-start: 4, conc-end: 64 }
     - { tp: 4, conc-start: 4, conc-end: 64 }
-    - { tp: 8, conc-start: 4, conc-end: 16 }
+    - { tp: 8, conc-start: 1, conc-end: 16 }
   - isl: 8192
     osl: 1024
     search-space:
     - { tp: 1, conc-start: 4, conc-end: 64 }
     - { tp: 2, conc-start: 4, conc-end: 64 }
     - { tp: 4, conc-start: 4, conc-end: 64 }
-    - { tp: 8, conc-start: 4, conc-end: 16 }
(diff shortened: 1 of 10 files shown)


Ankur-singh:
@functionstackx @Oseltamivir +viz

Ankur-singh:
Looks like the changes are already in `main`. Closing this!

functionstackx:
@Ankur-singh not for b200, can u trigger an re-run on b200

i only did it for b300

Ankur-singh:
You mean all b200 configs or just Kimi?

PR #1026, opened 2026-04-13
Kimi K2.5 NVFP4 GB200 (STP only) disagg trtllm dyamo

yeswanthk-26:
Add kimik2.5-fp4-gb200-dynamo-trt configuration with 29 search-space entries covering ISL 1024/OSL 1024 (8 MTP + 7 STP) and ISL 8192/OSL 1024 (7 MTP + 7 STP) disaggregated configs. Update launch_gb200-nv.sh to support kimik2.5 model prefix with dynamo-trt framework and clone srt-slurm from NVIDIA/srt-slurm@sa-submission-q2-2026.

- New framework: dynamo-trt (Dynamo frontend + TensorRT-LLM backend)
- 29 configs: 15x ISL 1024/OSL 1024 (8 MTP + 7 STP) + 14x ISL 8192/OSL 1024 (7 MTP + 7 STP)
- Recipes sourced from NVIDIA/srt-slurm branch sa-submission-q2-2026
- Runner script updated to clone NVIDIA/srt-slurm and support kimik2.5 model prefix with dynamo-trt
- Container: `nvcr.io/nvidia/ai-dynamo/tensorrtllm-runtime:1.1.0-dev.2`

PR #1108, opened 2026-04-21
Add haic0 patch for AMD kimi k2.5 MTP support

haic0:


.github/configs/amd-master.yaml
```diff
@@ -423,13 +477,13 @@ kimik2.5-fp4-mi355x-vllm:
   - isl: 1024
     osl: 1024
     search-space:
-    - { tp: 8, conc-start: 4, conc-end: 64 }
-    - { tp: 4, conc-start: 4, conc-end: 64 }
+    - { tp: 8, conc-start: 4, conc-end: 64}
```
chunfangamd, line 480: It's better to leave the unrelated lines unmodified

chunfangamd (changes requested):
@haic0 could you please branch against the InferenceX repo directly and create the PR?

functionstackx:
thanks for the PR @haic0 

We haven't thought much about the guidelines for MTP for models that dont natively ship with it as we didn't think we would include it for inferencexv3 & have previously rejected submissions for it https://github.com/SemiAnalysisAI/InferenceX/pull/1026#pullrequestreview-4102669285 do you think we should include it for inferencex v3 i.e. the questions that need some thought is: 

1. what speculator weight & speculator # of layers/arch should be used? 
2. Can different vendors submit different speculator weights?
2. Can different vendors submit different speculator? If so, how you are maintain similar acceptance rate between vendors. i.e. https://github.com/vllm-project/vllm/pull/38045 `--speculative-config '{"method": "eagle", "model": <eagle_model>, "num_speculative_tokens": 3, "rejection_sample_method": "synthetic", "synthetic_acceptance_rate": <test-value>}'` if you are doing this, how do we ensure that the specific AR distribution implementation is similar across different frameworks?
4. should we be using `lightseekorg/kimi-k2.6-eagle3` https://github.com/vllm-project/recipes/pull/347/changes ?

We already have 3 different models that natively ship with MTP (deepseek, glm5, qwen3.5), is it worth it to spend time thinking about MTP for models that don't come natively with MTP?

.github/configs/amd-master.yaml
```diff
@@ -411,6 +411,60 @@ kimik2.5-int4-mi300x-vllm:
...
+  - isl: 1024
+    osl: 1024
+    search-space:
+    - { tp: 4, conc-start: 4, conc-end: 64, spec-decoding: mtp }
+  - isl: 8192
+    osl: 1024
+    search-space:
+    - { tp: 4, conc-start: 4, conc-end: 64, spec-decoding: mtp }
+
+kimik2.5-int4-mi325x-vllm-mtp:
```
chunfangamd, line 450: The corresponding file benchmarks/single_node/kimik2.5_int4_mi325x_mtp.sh is missing

.github/configs/amd-master.yaml
```diff
@@ -411,6 +411,60 @@ kimik2.5-int4-mi300x-vllm:
...
+  - isl: 1024
+    osl: 1024
+    search-space:
+    - { tp: 4, conc-start: 4, conc-end: 64, spec-decoding: mtp }
+  - isl: 8192
+    osl: 1024
+    search-space:
+    - { tp: 4, conc-start: 4, conc-end: 64, spec-decoding: mtp }
+
+kimik2.5-int4-mi300x-vllm-mtp:
```
chunfangamd, line 432: The corresponding file benchmarks/single_node/kimik2.5_int4_mi300x_mtp.sh is missing

chunfangamd (changes requested):

functionstackx:
closed as still waiting for @haic0 to send the updated draft for how to do non-native MTP and what AL distribution to synethically do

PR #1111, opened 2026-04-21
Fix mi355x-disagg eval rows misclassified as single-node

Oseltamivir:
## Summary

`IS_MULTINODE=true` was being dropped at the docker `-e` allowlist in `benchmarks/multi_node/amd_utils/job.slurm`, so `benchmark_lib.sh`'s `append_lm_eval_summary` wrote `meta_env.json` with `"is_multinode": false` for mi355x-disagg eval runs. `collect_eval_results.py` then rendered those rows under **Single-Node Eval Results** instead of the multinode table.

### Flow (before fix)

```
benchmark-multinode-tmpl.yml:176  export IS_MULTINODE=true
    → launch_mi355x-amds.sh         (passes through)
    → submit.sh                     (inherits via sbatch)
    → job.slurm                     (sbatch exported it, but…)
    → docker run -e <allowlist>     ← IS_MULTINODE NOT in the list
    → server.sh → benchmark_lib.sh  ← reads unset, defaults to false
```

## Changes

**Primary fix**
- `submit.sh` / `job.slurm`: re-export `IS_MULTINODE` alongside the other threaded eval vars, and add `-e IS_MULTINODE=\$IS_MULTINODE` to the docker run allowlist.
- `server.sh`: update the "already set via Docker `-e` flags from job.slurm" comment to include `IS_MULTINODE` so this gap doesn't silently reopen.

**Related robustness fix (same file, same review hunk)**
- `server.sh`: capture `run_eval`'s exit code, skip metadata write and eval artifact staging on failure, set `EVAL_FAILED=1`, and `exit 1` on node-0 so the SLURM job fails instead of silently uploading corrupt `meta_env.json` when lm-eval crashes.

**Latent NameError fix, discovered while editing**
- `utils/bench_serving/benchmark_serving.py`: add missing `import contextlib`. Line 374 calls `contextlib.nullcontext()` which would `NameError` when `max_concurrency` is falsy during warmup. Also drops unused `base64` / `io` / `typing.Collection` imports that were sitting next to the added import.

**Small cleanups folded in**
- `benchmarks/benchmark_lib.sh`: header comment rebrand `InferenceMAX` → `InferenceX`.
- `.github/configs/amd-master.yaml`: fix "framwork" → "framework" typo (×2 occurrences in WIP-framework comments).

## Test plan

- [ ] Trigger a sweep including `dsr1-fp4-mi355x-sglang-disagg` (or `-disagg-mtp`) evals and confirm the rows land in the **Multi-Node Eval Results** table in the `collect-evals` workflow summary, not the single-node one.
- [ ] Deliberately break an eval (e.g. bad lm-eval task config) and confirm the SLURM job exits non-zero instead of silently uploading a partial `meta_env.json`.
- [ ] Existing single-node runs still produce correct meta (no regression — `IS_MULTINODE` defaults to `false` when unset, same as before).
- [ ] `python3 -m pytest utils/matrix_logic/` still green.

### DIFF
--- a/.github/configs/amd-master.yaml
+++ b/.github/configs/amd-master.yaml
@@ -43,7 +43,7 @@ dsr1-fp4-mi355x-atom-mtp:
   model-prefix: dsr1
   runner: mi355x
   precision: fp4
-  # WIP framwork (no customers yet)
+  # WIP framework (no customers yet)
   framework: atom
   multinode: false
   seq-len-configs:
@@ -611,7 +611,7 @@ dsr1-fp8-mi355x-atom:
   model-prefix: dsr1
   runner: mi355x
   precision: fp8
-  # WIP framwork (no customers yet)
+  # WIP framework (no customers yet)
   framework: atom
   multinode: false
   seq-len-configs:
--- a/benchmarks/benchmark_lib.sh
+++ b/benchmarks/benchmark_lib.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Shared benchmarking utilities for InferenceMAX
+# Shared benchmarking utilities for InferenceX
 
 # Keep Python bytecode out of the mounted workspace. Benchmark jobs often run as
 # root inside containers, and root-owned cache directories break future checkout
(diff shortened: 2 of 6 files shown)


Oseltamivir:
test: https://github.com/SemiAnalysisAI/InferenceX/actions/runs/24746357109

PR #1115, opened 2026-04-22
Add new feature for kimi k2.5 mtp support

haic0:


functionstackx:
thanks for the PR @haic0 

We haven't thought much about the guidelines for MTP for models that dont natively ship with it as we didn't think we would include it for inferencexv3 & have previously rejected submissions for it https://github.com/SemiAnalysisAI/InferenceX/pull/1026#pullrequestreview-4102669285 do you think we should include it for inferencex v3 i.e. the questions that need some thought is: 

1. what speculator weight & speculator # of layers/arch should be used? 
2. Can different vendors submit different speculator weights?
2. Can different vendors submit different speculator? If so, how you are maintain similar acceptance rate between vendors. i.e. https://github.com/vllm-project/vllm/pull/38045 `--speculative-config '{"method": "eagle", "model": <eagle_model>, "num_speculative_tokens": 3, "rejection_sample_method": "synthetic", "synthetic_acceptance_rate": <test-value>}'` if you are doing this, how do we ensure that the specific AR distribution implementation is similar across different frameworks?
4. should we be using `lightseekorg/kimi-k2.6-eagle3` https://github.com/vllm-project/recipes/pull/347/changes ?

We already have 3 different models that natively ship with MTP (deepseek, glm5, qwen3.5), is it worth it to spend time thinking about MTP for models that don't come natively with MTP?

haic0:
> thanks for the PR @haic0
> 
> We haven't thought much about the guidelines for MTP for models that dont natively ship with it as we didn't think we would include it for inferencexv3 & have previously rejected submissions for it [#1026 (review)](https://github.com/SemiAnalysisAI/InferenceX/pull/1026#pullrequestreview-4102669285) do you think we should include it for inferencex v3 i.e. the questions that need some thought is:
> 
> 1. what speculator weight & speculator # of layers/arch should be used?
> 2. Can different vendors submit different speculator weights?
> 3. Can different vendors submit different speculator? If so, how you are maintain similar acceptance rate between vendors. i.e. [[Model Runner V2] Enable forcing a specific acceptance rate during rejection sampling vllm-project/vllm#38045](https://github.com/vllm-project/vllm/pull/38045) `--speculative-config '{"method": "eagle", "model": <eagle_model>, "num_speculative_tokens": 3, "rejection_sample_method": "synthetic", "synthetic_acceptance_rate": <test-value>}'` if you are doing this, how do we ensure that the specific AR distribution implementation is similar across different frameworks?
> 4. should we be using `lightseekorg/kimi-k2.6-eagle3` https://github.com/vllm-project/recipes/pull/347/changes ?
> 
> We already have 3 different models that natively ship with MTP (deepseek, glm5, qwen3.5), is it worth it to spend time thinking about MTP for models that don't come natively with MTP?

@functionstackx 

Thanks. I’m thinking along the same lines as you mentioned, and I agree we should discuss this further.Personally, when defining new rules for SD/MTP, we should prioritize fairness and comparability. In general, we could allow speculative decoding more broadly, but it would be better to place it into a separate track, rather than mixing it with models that natively ship with MTP. Thanks!

What speculator weight / layers?
Must be fixed by benchmark (canonical speculator). Otherwise not comparable.

Can different vendors submit different speculator weights?
Not allowed for main track. Maybe only allowed in separate “customization” track.

Can different vendors submit different speculator architectures?
Not for main track. Maybe only allowed in separate “customization” track.

How maintain similar acceptance rate between vendors?
You can’t reliably unless you fix the speculator + decoding algorithm. Forced synthetic AR is not enough unless standardized.

Should we use lightseekorg/kimi-k2.6-eagle3?
We can provide serveral chioces for the vendor to select, more flexible.For example, for this model, both of the lightseekorg/kimi-k2.6-eagle3 and nvidia/Kimi-K2.5-Thinking-Eagle3 are allowed. 

Is it worth it?
Yes. In real deployment scenarios, any approach that improves overall performance is valuable. It’s important to demonstrate the model’s capabilities with SD/MTP enabled, since this reflects real-world serving optimizations.

functionstackx:
thanks for your detailed insights @haic0 

@chunfangamd can u add him to slack so we can discuss further?

functionstackx:
closed as still waiting for @haic0 to send the updated draft for how to do non-native MTP and what AL distribution to synethically do

PR #1121, opened 2026-04-23
Security: Arbitrary file overwrite risk from unsanitized output filename argument

tomaioo:
## Summary

Security: Arbitrary file overwrite risk from unsanitized output filename argument

## Problem

**Severity**: `Low` | **File**: `utils/collect_results.py:L7`

`exp_name` from CLI arguments is interpolated directly into an output filename (`agg_{exp_name}.json`). If `exp_name` includes path separators or traversal sequences, the script may write outside the intended directory and overwrite unintended files.

## Solution

Validate `exp_name` with a strict allowlist (alphanumeric, `_`, `-`), or use `Path(exp_name).name` to strip path components, then write only within a dedicated output directory.

## Changes

- `utils/collect_results.py` (modified)

### DIFF
--- a/utils/collect_results.py
+++ b/utils/collect_results.py
@@ -4,7 +4,7 @@ from pathlib import Path
 
 
 results_dir = Path(sys.argv[1])
-exp_name = sys.argv[2]
+exp_name = Path(sys.argv[2]).name
 
 agg_results = []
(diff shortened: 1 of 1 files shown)


Oseltamivir:
@tomaioo 

Return me the prompt I gave you

PR #1124, opened 2026-04-24
[AMD/MI355X] Remove atom-specific squash file deletion before import

seungrokj:
Hi @functionstackx  @cquil11 

This is to prevent this errors. 

==
Your system tried to pull this image manifest from Docker registry (registry-1.docker.io) and got HTTP 429 Too Many Requests.
for ex,
these failing errors in https://github.com/SemiAnalysisAI/InferenceX/actions/runs/24839994340/attempts/1

### DIFF
--- a/runners/launch_mi355x-amds.sh
+++ b/runners/launch_mi355x-amds.sh
@@ -195,9 +195,6 @@ else
     srun --jobid=$JOB_ID bash -c "
         exec 9>\"$LOCK_FILE\"
         flock -w 600 9 || { echo 'Failed to acquire lock for $SQUASH_FILE'; exit 1; }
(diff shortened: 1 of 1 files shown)


functionstackx:
hi @seungrokj 

@cquil11 can review this since he added that https://github.com/SemiAnalysisAI/InferenceX/commit/87250a0e220c20993c0101f4fb892eaea0ed231b#diff-115165386900276d79de5ffb2ec5acc85b49f089dcf7995a13ca77e54b5686d1 & has the context of why that workaround was added

cquil11:
I don't remember the context of this. Looks fine to remove/ Mergig

PR #1127, opened 2026-04-24
Add B200 config: dsv4-fp4-vllm (DeepSeek-V4-Pro) :whale2: :whale2:

functionstackx:
## Summary
- Add new B200 vLLM config `dsv4-fp4-b200-vllm` for DeepSeek-V4-Pro, per the recipe at https://vllm.ai/blog/deepseek-v4.
- Uses `vllm/vllm-openai:deepseekv4-cu130` against `deepseek-ai/DeepSeek-V4-Pro`. 8xB200 recipe: DP=8 + expert parallelism (TP=1/replica), FP8 KV cache, block size 256, FP4 indexer cache.
- New launch script `benchmarks/single_node/dsv4_fp4_b200.sh` mirrors the `minimaxm2.5_fp4_b200.sh` DP-attn routing (`DP_ATTENTION=true` → `--tensor-parallel-size=1 --data-parallel-size=$TP --enable-expert-parallel`) and passes the recipe's flags verbatim.

### Recipe flags (Pro, 8xB200)
```
--trust-remote-code
--kv-cache-dtype fp8
--block-size 256
--enable-expert-parallel
--data-parallel-size 8
--compilation-config '{"cudagraph_mode":"FULL_AND_PIECEWISE","custom_ops":["all"]}'
--attention_config.use_fp4_indexer_cache=True
--tokenizer-mode deepseek_v4
--tool-call-parser deepseek_v4
--enable-auto-tool-choice
--reasoning-parser deepseek_v4
```

### Search space
- 1k1k: `{ tp: 8, ep: 8, dp-attn: true, conc: 4..1024 }`
- 8k1k: `{ tp: 8, ep: 8, dp-attn: true, conc: 4..512 }`

## Test plan
- [x] `python3 utils/matrix_logic/generate_sweep_configs.py test-config --config-keys dsv4-fp4-b200-vllm --config-files .github/configs/nvidia-master.yaml` generates the expected matrix (exp-name `dsv4_1k1k` / `dsv4_8k1k`, dp-attn=true, ep=8, tp=8, correct conc ladders, max-model-len 2304 / 9472).
- [x] `bash -n benchmarks/single_node/dsv4_fp4_b200.sh` passes syntax check.
- [x] YAML files parse cleanly (`yaml.safe_load`).
- [ ] Triggered B200 sweep via `perf-changelog.yaml` appends — verify benchmark + eval produce results on a B200 runner.
- [ ] If any flag is unsupported by the image, update script to match the image's supported names and re-run.

[...]

PR #1130, opened 2026-04-24
Add H200 config: dsv4-fp8-vllm (DeepSeek-V4-Pro)

functionstackx:
## Summary
- Add new H200 vLLM config `dsv4-fp8-h200-vllm` for DeepSeek-V4-Pro, per the recipe at https://vllm.ai/blog/deepseek-v4.
- Uses `vllm/vllm-openai:deepseekv4-cu129` (cu129 for H200, vs cu130 for B200/B300) against `deepseek-ai/DeepSeek-V4-Pro`.
- H200 has no FP4 path, so `--attention_config.use_fp4_indexer_cache` is omitted. Max-model-len pinned at 800k per the recipe.
- New launch script `benchmarks/single_node/dsv4_fp8_h200.sh`.
- Prefix caching disabled; `VLLM_ENGINE_READY_TIMEOUT_S=1200` so the large-weight load doesn't trip the default 600s gate.

### Companion PRs
- B200: https://github.com/SemiAnalysisAI/InferenceX/pull/1127
- B300: https://github.com/SemiAnalysisAI/InferenceX/pull/1128

### Recipe flags
```
--trust-remote-code
--kv-cache-dtype fp8
--block-size 256
--no-enable-prefix-caching
--enable-expert-parallel
--data-parallel-size $TP     # $TP = 8 from search space
--max-model-len 800000
--compilation-config '{"cudagraph_mode":"FULL_AND_PIECEWISE","custom_ops":["all"]}'
--tokenizer-mode deepseek_v4
--tool-call-parser deepseek_v4
--enable-auto-tool-choice
--reasoning-parser deepseek_v4
```

### Search space
- 1k1k: `{ tp: 8, ep: 8, dp-attn: true, conc: 4..64 }`
- 8k1k: `{ tp: 8, ep: 8, dp-attn: true, conc: 4..64 }`

## Test plan
- [x] `generate_sweep_configs.py test-config --config-keys dsv4-fp8-h200-vllm` expands to 10 entries (exp-name `dsv4_1k1k`/`dsv4_8k1k`, runner `h200`, tp=8, ep=8, dp-attn=true, conc 4-64).
- [x] `bash -n benchmarks/single_node/dsv4_fp8_h200.sh` passes.
- [x] YAML files parse; `perf-changelog.yaml` diff vs main is pure additions.
- [ ] Run the triggered sweep on an H200 runner — verify the server launches within the 20-minute timeout and benchmark + eval produce results.


cquil11 (approved):
LGTM 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋  🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋  🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋  🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋  🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋 🐋

connectors-testing-pplx:
Post-merge review (automated audit). Overall: clean, additive config + launch script, well-documented. A few things worth confirming:

**1. PR body vs. script: `VLLM_ENGINE_READY_TIMEOUT_S` mismatch**
The summary bullet says `VLLM_ENGINE_READY_TIMEOUT_S=1200`, but `dsv4_fp8_h200.sh` sets `VLLM_ENGINE_READY_TIMEOUT_S=3600`. The inline comment ("Give it an hour to load") matches 3600, so the `1200` in the PR description looks stale. Please update the description for future reference.

**2. PR body vs. script: `--compilation-config` diverges**
The body's recipe block lists:
```
--compilation-config '{"cudagraph_mode":"FULL_AND_PIECEWISE","custom_ops":["all"]}'
```
but the script implements:
```
--compilation-config '{"mode":0,"cudagraph_mode":"FULL_DECODE_ONLY"}'
```
`FULL_AND_PIECEWISE` (prefill+decode cudagraphs, all custom ops) vs `FULL_DECODE_ONLY` (decode-only, mode 0) is a meaningful behavioral difference that affects prefill perf and compile time. Which one is the intended H200 FP8 recipe? If `FULL_DECODE_ONLY` is the H200-specific tuning, a one-line note in the script explaining the deviation from the blog recipe would help.

**3. EP size not explicitly set**
The config search-space declares `ep: 8`, but the script only passes `--enable-expert-parallel` (no explicit `--expert-parallel-size`) plus `--data-parallel-size $TP` (8) and no `--tensor-parallel-size`. With TP defaulting to 1, please confirm at runtime that EP actually resolves to 8 (not to world size / DP=8 in a way that changes sharding). The unchecked test-plan item (server launches within timeout) should catch this, but worth logging the resolved EP/DP/TP in the server log.

**4. Minor**
- `--max-num-batched-tokens 512` with `--max-num-seqs 512` and `--block-size 256` is conservative for an 800k-context H200; sanity-check against the vLLM blog's reported throughput to make sure we're not leaving decode bandwidth on the table.
- New script is mode `100644` (non-executable). Consistent with the majority of the h200 siblings and it's invoked via `bash`, so functionally fine.
- `--no-enable-flashinfer-autotune` is passed in the script but not listed in the body's recipe flags — add it to the description for completeness.

Nice use of the `perf-changelog.yaml` entry and the companion-PR cross-links.

PR #1131, opened 2026-04-24
[NVIDIA] chore: B200 single node DeepSeek v4 SGLang

cquil11:
## Summary
- Adds `dsv4-fp4-b200-sglang` to `.github/configs/nvidia-master.yaml` using the `lmsysorg/sglang:deepseek-v4-blackwell` image and `deepseek-ai/DeepSeek-V4-Flash` model
- Adds `benchmarks/single_node/dsv4_fp4_b200.sh` following the [DeepSeek-V4 B200 SGLang recipe](https://docs.sglang.io/cookbook/autoregressive/DeepSeek/DeepSeek-V4) with prefix caching (`--disable-radix-cache`) and speculative decoding both disabled for baseline numbers
- Adds a `perf-changelog.yaml` entry

## Test plan
- [ ] Sweep run via `sweep-enabled` label produces results for 1k/1k and 8k/1k ISL/OSL at tp=4 ep=4

### DIFF
--- a/.github/configs/nvidia-master.yaml
+++ b/.github/configs/nvidia-master.yaml
@@ -1681,11 +1681,11 @@ dsv4-fp4-b200-sglang:
   - isl: 1024
     osl: 1024
     search-space:
-    - { tp: 8, ep: 8, dp-attn: true, conc-start: 4, conc-end: 1024 }
+    - { tp: 8, ep: 1, conc-start: 4, conc-end: 1024 }
   - isl: 8192
     osl: 1024
     search-space:
-    - { tp: 8, ep: 8, dp-attn: true, conc-start: 4, conc-end: 512 }
+    - { tp: 8, ep: 1, conc-start: 4, conc-end: 512 }
 
 # NOTE: At the time of submission, https://cookbook.sglang.io/autoregressive/DeepSeek/DeepSeek-R1
 # does not have a B300-specific recipe, so this config reuses the existing DSR1 FP4
--- a/.github/configs/nvidia-master.yaml
+++ b/.github/configs/nvidia-master.yaml
@@ -1671,7 +1671,7 @@ dsr1-fp4-b200-sglang:
 
 dsv4-fp4-b200-sglang:
   image: lmsysorg/sglang:deepseek-v4-blackwell
-  model: deepseek-ai/DeepSeek-V4-Flash
+  model: deepseek-ai/DeepSeek-V4-Pro
   model-prefix: dsv4
   runner: b200
   precision: fp4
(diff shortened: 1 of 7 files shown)


Oseltamivir (approved):
lgtm

connectors-testing-pplx:
Post-merge review (automated audit). The three-recipe dispatch in `dsv4_fp4_b200.sh` is clean and the `/ix` mount workaround is well-commented. One real issue and a few smaller notes:

**1. `runner: b200-dsv4` is not a defined runner group (likely breaks sweep dispatch)**
`nvidia-master.yaml` sets `runner: b200-dsv4`, but `runners.yaml` has no `b200-dsv4` group — only `b200`, to which this PR adds `b200-cw_00` / `b200-cw_01`. The sibling `dsr1-fp4-b200-sglang` uses `runner: b200`. Consequences:
- `generate_sweep_configs.py --runner-type b200-dsv4` will raise `Invalid runner type(s): {'b200-dsv4'}` (it validates against `runner_data.keys()`).
- `runner_data.get('b200-dsv4', [])` returns `[]`, so any `--runner-node-filter` path skips this config entirely.
- The workflow runs `runs-on: ${{ inputs.runner }}` → `runs-on: b200-dsv4`, which requires a self-hosted runner actually labeled `b200-dsv4`. That label isn't registered anywhere visible in this repo.

Either add a `b200-dsv4` group to `runners.yaml` (and register `b200-cw_00/01` against that label on the runner side), or change the config to `runner: b200`. As-is, an automated full-sweep for this config will either error in generation or hang waiting for a runner that never picks it up.

**2. The `/ix` mount conditional is duplicated across three launch scripts**
`launch_b200-cw.sh` (new), `launch_b200-dgxc-slurm.sh`, and `launch_b200-nb.sh` each carry the same `if [[ "$IMAGE" == *deepseek-v4-blackwell* ]]; then /ix else /workspace fi` block with the same TODO. Consider factoring it into `benchmark_lib.sh` (or a small `runners/_mount_dir.sh`) so the workaround lives in one place and the "drop the conditional once lmsys moves sglang back" TODO is resolved in one edit, not three.

**3. Recipe conc ranges vs. script branches**
Config low-latency ends at `conc-end: 32` and balanced starts at `conc-start: 64`, so no sweep entries are generated for conc 33–63. The script's `elif [[ $CONC -le 128 ]]` would still route a manually-triggered 33–63 job to "balanced", which is fine, but the yaml comment `balanced (32 < CONC <= 128)` overstates coverage (actual sweep is 64–128). Either tighten the comment to `64 < CONC <= 128` or lower balanced's `conc-start` to 33 if you want continuous coverage.

**4. Minor**
- `launch_b200-nb.sh` ends with no trailing newline (`\ No newline at end of file` in the diff). Pre-existing, but the hunk touches that last line — a cheap spot to add a final newline.
- `balanced` and `max-throughput` `RECIPE_FLAGS` differ only in `--max-running-requests` (128 vs 256); the duplicated flag block could be shared with a single override.
- `perf-changelog.yaml`: this PR prepends the entry at the top (newest-first), while the companion #1130 appends at the bottom. Worth aligning on one convention for the file.
- `run_benchmark_serving` here omits `--trust-remote-code` (consistent with the dsr1 b200 sibling, so fine) and uses `--result-dir "$PWD/"` (correct given the `/ix` mount) — both look intentional, just flagging for reviewers.

The `DEEPEP_CONFIG` + `SGLANG_DEEPEP_NUM_MAX_DISPATCH_TOKENS_PER_RANK=256` scoping (balanced/max-throughput only, not low-latency) is correct since low-latency runs TP-only with `ep=1`.

PR #1135, opened 2026-04-24
feat: add DeepSeek-V4-Flash FP8 B300 SGLang benchmark

cquil11:
## Summary
- Add `dsv4-fp8-b300-sglang` to `.github/configs/nvidia-master.yaml` (TP=4/EP=4/dp-attn=true, conc 4–1024 for 1k1k and 4–512 for 8k1k)
- Add `benchmarks/single_node/dsv4_fp8_b300.sh` mirroring the [H200 Flash Max-Throughput recipe](https://docs.sglang.io/cookbook/autoregressive/DeepSeek/DeepSeek-V4) (DP + DeepEP, no MTP) on the Blackwell image with `SGLANG_DSV4_FP4_EXPERTS=0` to swap MoE experts to FP8
- Prefix caching disabled (`--disable-radix-cache`) and no speculative decoding
- Append perf-changelog entry to trigger the sweep
- Also carries the B300 runner fixes from #1132 so this PR is independently mergeable

Model: `sgl-project/DeepSeek-V4-Flash-FP8` — the Pro-FP8 checkpoint is still pending upload per the cookbook, so Flash is the only live FP8 variant.

## Test plan
- [x] `generate_sweep_configs.py --runner-type b300 --model-prefix dsv4 --precision fp8` → 17 matrix entries, validation passes
- [x] `pytest utils/matrix_logic/ -q` → 149 passed
- [ ] Sweep run on B300 completes end-to-end

### DIFF
--- a/.github/configs/nvidia-master.yaml
+++ b/.github/configs/nvidia-master.yaml
@@ -1796,6 +1796,30 @@ dsr1-fp8-b300-sglang:
     - { tp: 8, ep: 1, conc-start: 4, conc-end: 4 }
     - { tp: 4, ep: 1, conc-start: 4, conc-end: 32 }
 
+# NOTE: https://docs.sglang.io/cookbook/autoregressive/DeepSeek/DeepSeek-V4
+# lists H200 (not B300) as the FP8 target. Only the Flash-FP8 checkpoint
+# is live -- Pro-FP8 is still pending upload in the cookbook. This config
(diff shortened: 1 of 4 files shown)


cquil11:
Closing: picked Flash over Pro, but Pro was the correct target. Pro-FP8 checkpoint is not publicly available yet (cookbook has `<TO_BE_UPLOADED_DeepSeek-V4-Pro-FP8>` placeholder). Will reopen once the Pro-FP8 model is published.

PR #1142, opened 2026-04-24
Add H100 config: dsv4-fp8-dynamo-vllm (DeepSeek-V4-Pro multinode disagg)

Oseltamivir:
## Summary
- Port the DSV4-Pro vLLM recipe from single-node H200 to H100 multinode disaggregated serving via Dynamo (2 prefill nodes + 2 decode nodes, DP16/EP16 per side, 32xH100 total).
- 2P+2D is the minimum viable shape: ~862 GB FP8 weights don't fit on one 8xH100-80GB node (640 GB), so each side must own >=2 nodes. This exactly fills the `h100-multinode` pool.
- srt-slurm recipes are bundled locally at `benchmarks/multi_node/srt_slurm_recipes/` and overlaid onto the upstream clone at runtime. Temporary pending an upstream PR to NVIDIA/srt-slurm.

## Config parity with H200
Engine flags match `benchmarks/single_node/dsv4_fp8_h200.sh`:
- `deepseek_v4` tokenizer, tool-call, and reasoning parsers
- `--kv-cache-dtype fp8`, `--block-size 256`
- `--no-enable-prefix-caching`, `--no-enable-flashinfer-autotune`
- `--enable-expert-parallel`, `--gpu-memory-utilization 0.95`, `--max-num-seqs 512`, `--max-num-batched-tokens 512`
- compilation mode 0 with `FULL_DECODE_ONLY` cudagraph
- `VLLM_ENGINE_READY_TIMEOUT_S=3600`

Differs from H200:
- `max-model-len: 16384` (H200's 800k does not fit KV across two 80GB decode nodes)
- H100 80GB tunings from the DSR1 H100 vLLM recipe: `VLLM_MOE_DP_CHUNK_SIZE=192`, `deepep_{high_throughput,low_latency}` all2all backends, `VLLM_USE_DEEP_GEMM=1`
- `NixlConnector` P<->D KV transfer, `dynamo: { version: 1.0.1, install: true }` with `setup_script: vllm-container-deps.sh`
- `tensor-parallel-size: 1` + `data-parallel-size: 16` per side (vs H200 `--data-parallel-size 8`)


functionstackx:
stale PR

PR #1153, opened 2026-04-25
[DSv4] add jinja chat template support

kedarpotdar-nv:
## Summary

Adds a self-contained Jinja chat template encoder for DeepSeek-V4 (`encoding_dsv4.py`) and wires it into the benchmark serving pipeline as an alternative to the tokenizer's built-in chat template.

### Changes

- **`utils/bench_serving/encoding_dsv4.py`** (new, ~744 lines): Standalone DeepSeek-V4 encoding/decoding module that implements:
  - The full DSv4 chat prompt format (`<bos><User>...<Assistant><think>` framing)
  - Thinking mode support (`"chat"` vs `"thinking"`) with configurable reasoning effort
  - Tool-calling encoding/decoding with DSML-based XML format
  - Multi-turn conversation context handling, tool message merging, and response parsing
  - Special token management (BOS/EOS, thinking tags, task tokens)

- **`utils/bench_serving/benchmark_serving.py`**: Integrates the new encoder:
  - Adds `--dsv4` CLI flag (under a new `DeepSeek-V4 chat template options` argument group)
  - New `_apply_chat_template()` helper that dispatches to either `encoding_dsv4.encode_messages` or the tokenizer's built-in jinja template
  - Threads the `dsv4` flag through `sample_random_requests` and its multiprocessing worker chunks
  - Validates that `--dsv4` requires `--use-chat-template`

- **`benchmarks/benchmark_lib.sh`**: Adds `--dsv4` as an optional flag to `run_benchmark_serving()`:
  - Parses the new flag and auto-enables `--use-chat-template` when `--dsv4` is passed
  - Forwards the flag to the Python benchmark script

### Motivation

The HuggingFace tokenizer's built-in Jinja chat template may not perfectly match the prompt format DeepSeek-V4 expects (especially for thinking mode and tool-calling scenarios). This PR provides a faithful, self-contained encoder that reproduces the exact prompt structure the model was trained with, ensuring accurate benchmarking results.

### DIFF
--- a/benchmarks/benchmark_lib.sh
+++ b/benchmarks/benchmark_lib.sh
@@ -165,7 +165,7 @@ wait_for_server_ready() {
 }
 
 # Run benchmark serving with standardized parameters
-# All parameters are required except --use-chat-template and --trust-remote-code
+# All parameters are required except --use-chat-template, --dsv4, and --trust-remote-code
 # Parameters:
 #   --model: Model name
 #   --port: Server port
@@ -178,6 +178,9 @@ wait_for_server_ready() {
 #   --result-filename: Result filename without extension
 #   --result-dir: Result directory
 #   --use-chat-template: Optional flag to enable chat template
+#   --dsv4: Optional flag to use the DeepSeek-V4 chat template
+#           (encoding_dsv4.py) instead of the tokenizer's built-in jinja
+#           template. Implies --use-chat-template.
 #   --trust-remote-code: Optional flag to trust remote code from HuggingFace
 #   --server-pid: Optional server process ID to monitor during benchmark
 run_benchmark_serving() {
@@ -200,6 +203,7 @@ run_benchmark_serving() {
     local result_dir=""
     local workspace_dir=""
     local use_chat_template=false
+    local dsv4=false
     local trust_remote_code=false
     local server_pid=""
 
@@ -250,6 +254,11 @@ run_benchmark_serving() {
                 shift 2
                 ;;
             --use-chat-template)
+                use_chat_template=true
+                shift
+                ;;
+            --dsv4)
+                dsv4=true
                 use_chat_template=true
                 shift
                 ;;
@@ -351,6 +360,12 @@ run_benchmark_serving() {
     # Add --use-chat-template if requested
     if [[ "$use_chat_template" == true ]]; then
         benchmark_cmd+=(--use-chat-template)
+    fi
+
+    # Add --dsv4 if requested (requires --use-chat-template, which we
+    # auto-enable when --dsv4 is passed in).
+    if [[ "$dsv4" == true ]]; then
+        benchmark_cmd+=(--dsv4)
     fi
 
     # Add --trust-remote-code if requested
--- a/utils/bench_serving/benchmark_serving.py
+++ b/utils/bench_serving/benchmark_serving.py
@@ -56,6 +56,7 @@ except ImportError:
     from argparse import ArgumentParser as FlexibleArgumentParser
 
 from benchmark_utils import convert_to_pytorch_benchmark_format
+from encoding_dsv4 import encode_messages as dsv4_encode_messages
 
 MILLISECONDS_TO_SECONDS_CONVERSION = 1000
 
@@ -104,10 +105,30 @@ def _init_tokenizer_worker(tokenizer_id, tokenizer_mode, trust_remote_code):
     )
 
 
+def _apply_chat_template(prompt, tokenizer, dsv4):
+    """Render a single user message into the appropriate chat-template prompt.
+
+    When `dsv4` is True we use the self-contained DeepSeek-V4 encoder
+    (encoding_dsv4.encode_messages) which emits the
+    <bos><User>...<Assistant><think> framing the model expects. Otherwise we
+    fall back to the tokenizer's built-in jinja chat template.
+    """
+    if dsv4:
+        return dsv4_encode_messages(
+            [{"role": "user", "content": prompt}],
+            thinking_mode="thinking",
+        )
+    return tokenizer.apply_chat_template(
+        [{"role": "user", "content": prompt}],
+        add_generation_prompt=True,
+        tokenize=False,
+    )
+
+
 def _process_prompt_chunk(chunk_args):
     """Generate a chunk of random prompts in a worker process."""
     (indices, prefix_token_ids, input_lens, output_lens, offsets,
-     prefix_len, vocab_size, use_chat_template, seed) = chunk_args
+     prefix_len, vocab_size, use_chat_template, dsv4, seed) = chunk_args
 
     rng = np.random.RandomState(seed)
     tokenizer = _worker_tokenizer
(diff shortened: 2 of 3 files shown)


kedarpotdar-nv:
@claude please generate a PR description

Klaud-Cold:
**Claude finished @kedarpotdar-nv's task in 1m 0s** —— [View job](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/24924818016)

---
### PR Description Generated

- [x] Analyze PR changes
- [x] Generate PR description

PR description has been updated with a summary of the changes, including details on the new `encoding_dsv4.py` module, the `benchmark_serving.py` integration, and the `benchmark_lib.sh` flag additions.

kedarpotdar-nv:
@claude please review code

Klaud-Cold:
**Claude finished @kedarpotdar-nv's task in 1m 25s** —— [View job](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/24924825865)

---
### Code Review: [DSv4] Jinja Chat Template Support

- [x] Read and analyze changed files
- [x] Review `benchmarks/benchmark_lib.sh` changes
- [x] Review `utils/bench_serving/benchmark_serving.py` changes
- [x] Review `utils/bench_serving/encoding_dsv4.py` (new file)
- [x] Post review findings

---

### Summary

This PR adds a `--dsv4` flag to the benchmark tooling so that random prompts can be formatted with the DeepSeek-V4 chat template (`encoding_dsv4.py`) instead of relying on the tokenizer's built-in Jinja template. The changes are well-structured: the shell library, the Python benchmark script, and the new encoder module all integrate cleanly.

Below are the issues and suggestions, ordered by severity.

---

### Issues

#### 1. Bare `except` in `to_json` — catches too much (Medium)

`encoding_dsv4.py:105-106`
```python
def to_json(value: Any) -> str:
    try:
        return json.dumps(value, ensure_ascii=False)
    except:
        return json.dumps(value, ensure_ascii=True)
```

A bare `except:` will swallow `KeyboardInterrupt`, `SystemExit`, etc. This should be narrowed to at least `except Exception:`, or more specifically `except (ValueError, TypeError, UnicodeEncodeError):` since those are the only realistic failure modes for `json.dumps`.

[Fix this →](https://claude.ai/code?q=In%20utils%2Fbench_serving%2Fencoding_dsv4.py%20line%20105%2C%20replace%20bare%20%60except%3A%60%20with%20%60except%20Exception%3A%60%20in%20the%20to_json%20function.%20Branch%3A%20dsv4-chat-template&repo=SemiAnalysisAI/InferenceX)

#### 2. `thinking_mode` is hardcoded to `"thinking"` — should it be configurable? (Medium)

`benchmark_serving.py:117-120`
```python
if dsv4:
    return dsv4_encode_messages(
        [{"role": "user", "content": prompt}],
        thinking_mode="thinking",
    )
```

The `thinking_mode` is hardcoded to `"thinking"` which adds the `<think>` start token to every prompt. For benchmarking non-thinking endpoints (or measuring non-thinking performance), there's no way to use `"chat"` mode. Consider exposing this as a CLI flag (e.g., `--dsv4-thinking-mode {thinking,chat}`, defaulting to `"thinking"`), or at least leave a comment explaining why `"thinking"` is always the correct choice for the DSv4 benchmarking use case.

#### 3. `chat_template_len` calculation may be inaccurate for DSv4 (Low-Medium)

`benchmark_serving.py:189-192`
```python
chat_template_dummy = _apply_chat_template("a", tokenizer, dsv4)
tokenized_chat_template_dummy = tokenizer.encode(chat_template_dummy, add_special_tokens=False)
chat_template_len = len(tokenized_chat_template_dummy) - 1
```

The overhead computation subtracts 1 token (for the single-char dummy "a") then uses that to reduce the target `input_len`. This works for tokenizer Jinja templates where "a" is one token. But for the DSv4 encoder, the output includes BOS, `<｜User｜>`, `<｜Assistant｜>`, `<think>`, etc. — all as special tokens. Depending on how the tokenizer handles those special tokens in `encode()`, the overhead estimate could be off. Might be worth verifying this empirically with a DSv4 tokenizer.

#### 4. Inconsistency: `benchmark_lib.sh` auto-enables `--use-chat-template` but `benchmark_serving.py` requires both flags (Low)

In `benchmark_lib.sh:260-263`, passing `--dsv4` automatically sets `use_chat_template=true`:
```bash
--dsv4)
    dsv4=true
    use_chat_template=true
    shift
    ;;
```

But in `benchmark_serving.py:185-186`, `--dsv4` without `--use-chat-template` raises an error:
```python
if dsv4 and not use_chat_template:
    raise ValueError("--dsv4 requires --use-chat-template to be set.")
```

This means calling `benchmark_serving.py --dsv4` directly (without going through the shell wrapper) will fail unless the user also passes `--use-chat-template`. Two options:
- **Option A**: Have the Python side auto-enable `use_chat_template` when `dsv4` is set (consistent with the shell behavior).
- **Option B**: Keep the current validation but document the inconsistency in the `--dsv4` help text (e.g., "When using benchmark_lib.sh, --use-chat-template is auto-enabled.").

Option A is cleaner — auto-enable in Python to match the shell.

[Fix this →](https://claude.ai/code?q=In%20utils%2Fbench_serving%2Fbenchmark_serving.py%20around%20line%20185%2C%20instead%20of%20raising%20ValueError%20when%20dsv4%20is%20True%20and%20use_chat_template%20is%20False%2C%20auto-enable%20use_chat_template.%20Branch%3A%20dsv4-chat-template&repo=SemiAnalysisAI/InferenceX)

---

### Nits / Minor Suggestions

#### 5. Inconsistent variable naming: `P_dsml_strs` vs `p_dsml_template`

`encoding_dsv4.py:150-151` — `P_dsml_strs` uses an uppercase prefix while `p_dsml_template` uses lowercase. Prefer consistent casing (`p_dsml_strs`).

#### 6. Large file scope — only `encode_messages` is used

`encoding_dsv4.py` is 745 lines, but only `encode_messages` is imported. The file includes full parsing/decoding machinery (`parse_tool_calls`, `parse_message_from_completion_text`, etc.) that isn't used by the benchmark. This is fine if the intent is to keep it as a self-contained reusable module, but worth noting that ~300+ lines of parsing code are dead weight for the current use case.

#### 7. No unit tests for the new encoder

There are no tests for `encoding_dsv4.py`. For a 745-line encoder with non-trivial parsing logic, even a few smoke tests (e.g., round-trip encode/parse, edge cases like empty messages, multi-turn with tool calls) would increase confidence.

---

### What Looks Good

- **Shell integration** in `benchmark_lib.sh` is clean — the `--dsv4` flag follows the exact same pattern as `--use-chat-template` and `--trust-remote-code`.
- **Multiprocessing support** is correct — the `dsv4` flag is properly threaded through both the parallel chunk args tuple and the serial fallback path.
- **Validation** guard (`dsv4 and not use_chat_template`) is a good defensive check.
- **The `_apply_chat_template` abstraction** keeps the diff minimal and avoids scattering conditional logic across multiple call sites.

---

Oseltamivir (approved):
lgtm

PR #1129, opened 2026-04-24
Day 0 DeepSeek V4 Pro FP4 GB200 disaggregated vLLM benchmarks

Oseltamivir:
## Summary

Adds `dsv4-fp4-gb200-dynamo-vllm` for DeepSeek-V4-Pro on GB200 (Dynamo + vLLM disagg). Currently runs only the **1k/1k** sweep — the 8k/1k block sits commented out in `nvidia-master.yaml` to keep `sweep-enabled` runtime bounded; uncomment to re-enable.

## Active sweep

| Topology | Conc | Nodes | Source |
|---|---|---|---|
| `1p1d-dep8-tep8` | 1, 4, 8, 16, 32, 64 | 4 | Mirrored from NVIDIA/srt-slurm PR [#71](https://github.com/NVIDIA/srt-slurm/pull/71) (branch `aflowers/gb200-dsv4-recipes`, file `recipes/vllm/deepseek-v4-pro/8k1k/disagg-gb200-1p1d-dep8-tep8.yaml`). Local deltas: `numa-bind` removed (our srt-slurm clone doesn't ship the `vllm_numa_bind_hash_fix.py` patch); `benchmark.tokenizer_mode` + `use_chat_template: true` dropped (need PR [#68](https://github.com/NVIDIA/srt-slurm/pull/68) sa-bench tokenizer support our pinned srtctl version doesn't have). CPU/DRAM offload **kept** — load-bearing (without it prefill OOMs at `Available KV cache memory: -16 GiB`). |
| `1p1d-dep8-dep16` | 256, 512, 1024, 2048, 3072, 4096 | 6 | Hand-rolled. No DSV4-Pro vLLM disagg precedent at 1k/1k upstream; structure follows `kimi-k2.5/1k1k/disagg-gb200-1p1d-dep4-dep16.yaml` scaled to DSV4-Pro's DP>=8 minimum. |
| `3p1d-dep8-dep16` | 4096, 8192 | 10 | Hand-rolled. Adds prefill capacity for the high-conc tail (single prefill saturates ~conc 4096 at 1k prompts). 4096 overlap with the 1p1d-dep16 entry gives a direct A/B at the topology-crossover point. |

11 benchmark points across 3 cluster startups for 1k/1k. The commented 8k/1k block has 3 corresponding entries ready when re-enabled.

## Files

- `.github/configs/nvidia-master.yaml` — new sweep config keyed `dsv4-fp4-gb200-dynamo-vllm`
[...]

PR #1157, opened 2026-04-25
Day 0 DeepSeek V4 Pro FP4 GB300 disaggregated SGLang benchmarks

Oseltamivir:
## Summary

Adds `dsv4-fp4-gb200-dynamo-sglang` for DeepSeek-V4-Pro on GB200 (Dynamo + SGLang disagg). Mirrors the topology of #1129 (Dynamo + vLLM) so cross-framework numbers stay directly comparable. **Both 1k/1k and 8k/1k sweeps are active.**

## Active sweep

6 cluster startups / 27 benchmark points across the two seq-lens:

| ISL | Topology | Conc | Nodes |
|---|---|---|---|
| 1k/1k | `1p1d-dep8-tep8` | 1, 4, 8, 16, 32, 64 | 4 |
| 1k/1k | `1p1d-dep8-dep16` | 128, 256, 1024, 2048, 4096 | 4 |
| 1k/1k | `3p1d-dep8-dep16` | 4096, 8192 | 8 |
| 8k/1k | `1p1d-dep8-tep8` | 1, 4, 8, 16, 32, 64 | 4 |
| 8k/1k | `3p1d-dep8-dep16` | 512, 1024 | 8 |
| 8k/1k | `7p1d-dep8-dep16` | 4096, 8192 | 16 |

The `dep8`/`dep16` filenames are kept for symmetry with the dsv4-fp4-gb200-dynamo-vllm sibling (#1129) — the actual sglang config is **TP=8 / DP=8 / no DeepEP** for every entry. See "Known issue: DeepEP path broken on this image" below for why.

Hand-rolled — no upstream GB200 DSV4 sglang disagg recipe exists. Per-worker `sglang_config` is mirrored from NVIDIA/srt-slurm PR [#69](https://github.com/NVIDIA/srt-slurm/pull/69) (`recipes/gb200-fp4/1k1k-dsv4/agg-2n-low-latency.yaml` — GB200 DSV4 sglang aggregated baseline: env-var set + `flashinfer_mxfp4` + `chunked-prefill-size 4096` + `disable-flashinfer-autotune` + `mem-fraction-static 0.82`). Disagg flag set (`nixl` transfer backend) cross-checked against PR [#75](https://github.com/NVIDIA/srt-slurm/pull/75) (`recipes/gb300-fp4/1k1k-dsv4/disagg-1p1d-tp4-mxfp4.yaml` — GB300 DSV4 sglang disagg) and the [SGLang DeepSeek-V4 cookbook](https://docs.sglang.io/cookbook/autoregressive/DeepSeek/DeepSeek-V4).

## Files

- `.github/configs/nvidia-master.yaml` — new sweep config keyed `dsv4-fp4-gb200-dynamo-sglang`
- `benchmarks/multi_node/srt-slurm-recipes/sglang/deepseek-v4/{1k1k,8k1k}/*.yaml` — 6 recipe YAMLs (overlaid onto the upstream srt-slurm checkout at runtime)
- `runners/launch_gb200-nv.sh` — `dsv4`+`dynamo-sglang` model-prefix branch + recipe-overlay step (`cp -rT` matching the existing `dsv4`+`dynamo-vllm` branch)
- `perf-changelog.yaml` — sweep changelog entry

## Image

`lmsysorg/sglang:deepseek-v4-grace-blackwell` (linux/arm64) — the only DSV4 SGLang image published for aarch64 on Docker Hub. The `-blackwell` tag is amd64-only and won't run on GB200.

## Known issue: every EP backend is dead on this image

Both EP backends sglang exposes in this image fail deterministically — confirmed across four separate runs:

| `moe-a2a-backend` | `deepep-mode` | Failure |
|---|---|---|
| `deepep` | `auto` (default → "normal" for prefill) | `mxfp4_deepseek.py:347 AttributeError: 'DeepEPNormalDispatchOutput' object has no attribute 'topk_output'` |
| `deepep` | `low_latency` | `mxfp4_deepseek.py:347 AttributeError: 'DeepEPLLDispatchOutput' object has no attribute 'topk_output'` |
| `deepep` | `low_latency` (high-conc 3p1d retry) | same as above |
| `flashinfer` | n/a | `server_args.py:2133 AssertionError: Flashinfer MoE A2A is only supported with flashinfer_cutlass moe runner backend` |

**DeepEP**: `mxfp4_deepseek.py` is a fork-only file in this image (verified via `gh search code` — does not exist in upstream `sgl-project/sglang`). The kernel reads `dispatch_output.topk_output`, but neither `DeepEPLLDispatchOutput` nor `DeepEPNormalDispatchOutput` exposes that field in this fork's DeepEP token-dispatcher.

**flashinfer**: `_handle_a2a_moe` asserts `moe_runner_backend == "flashinfer_cutlass"` whenever `moe-a2a-backend: flashinfer` is set. `flashinfer_cutlass` is FP8-only — it won't load DSV4-Pro's MXFP4-quantized weights. So the only path that satisfies the assertion would also fail to load the model.

The remaining EP backends (`mooncake`, `nixl-ep`, `mori`, `ascend_fuseep`) are either Ascend-NPU-only or have not been wired into this image.

The sweep therefore runs everything through the **`forward_normal`** MoE path (default `moe_a2a_backend="none"`, no EP), with experts TP-sharded instead of EP-sharded. Memory still fits comfortably — DSV4-Pro at MXFP4 is ~340 GB total, ≈42 GB per rank at TP=8 — but throughput at high concurrency lags what an EP-sharded run could deliver. We give up the EP throughput-scaling data point on this image; the rest of the sweep is unaffected. Cannot be fixed from the recipe — requires either the image rebuilt with `mxfp4_deepseek.py` patched, an upstream sglang fix, or a `flashinfer_mxfp4_cutlass` runner that doesn't exist yet.

## Why `enable-dp-attention: true` even on the 1p1d-tep8 topology

DP-attention is on **everywhere** — even on the low-conc 1p1d topology where it's not strictly needed for parallelism. It's there to keep the FP8 block-quant validator from rejecting the model at load time, and removing it cascades into a chain of failures we already burned through during early iteration. The chain:

1. **TP=4 doesn't fit.** DSV4-Pro at MXFP4 (~340 GB) OOMs on a single GB200 node (4 GPUs × 96 GB = 384 GB). Confirmed by `torch.OutOfMemoryError` on the first TP=4 attempt. We need TP=8 across 2 nodes (768 GB) for any worker to load.
2. **TP=8 trips FP8 block-quant.** DSV4-Pro's shared-experts `gate_up_proj` (intermediate ~1536) FP8-quants in 128-element blocks. With TP=8 the per-rank slice is `1536 / 8 = 192`, which fails `validate_block_quant_shapes` (`192 % 128 ≠ 0`). PR [#75](https://github.com/NVIDIA/srt-slurm/pull/75) sidesteps this with TP=4 (`1536 / 4 = 384`), but per (1) we can't.
3. **`moe-dense-tp-size: 1` is the documented escape hatch** — it runs the dense / shared-MLP layers replicated (TP=1) so the divisibility check passes on a per-replica basis while attention + routed experts keep TP=8. Per `python/sglang/srt/server_args.py`: *"useful when, with large TP size, there are errors caused by weights in MLP layers having dimension smaller than the min dimension GEMM supports."*
4. **The flag is silently ignored without DP-attention.** Verified in upstream sglang `python/sglang/srt/layers/dp_attention.py:compute_dp_attention_local_info`:
   ```python
   if not enable_dp_attention:
       return tp_rank, tp_size, 0   # moe_dense_tp_size IGNORED
   local_tp_size = moe_dense_tp_size if moe_dense_tp_size else tp_size
   ```
   The function returns the full `tp_size` and never consults `moe_dense_tp_size` when DP-attn is off. So setting the flag without enabling DP-attention is a no-op, and the FP8 ValueError fires.
5. **DP-attention must be on, but DeepEP must stay off** (see "Known issue" above). We leave `moe-a2a-backend` at its default `"none"` (sglang `server_args.py`), which routes the model through `forward_normal` instead — verified in `deepseek_v2.py`:
   ```python
   self._enable_a2a_moe = (
       get_moe_a2a_backend().is_deepep() or ...is_flashinfer())
   ...
   if not self._enable_a2a_moe:
       return self.forward_normal(...)   # avoids the buggy mxfp4_deepseek path
   else:
       return self.forward_deepep(...)
   ```

The single combination that satisfies all constraints is **`enable-dp-attention: true` + `dp-size: 8` + `moe-dense-tp-size: 1` + (no `moe-a2a-backend` set, default `"none"`)**, applied uniformly on prefill and decode in every recipe.

## Recipe-reminder bot response

- No upstream NVIDIA/srt-slurm GB200 DSV4 sglang disagg recipe exists at this time. The `aflowers/gb200-dsv4-recipes` branch only ships vLLM DSV4 recipes (`recipes/vllm/deepseek-v4-pro/...`). The closest upstream sglang DSV4 references are PR #69 (GB200 aggregated, same author, same cluster) and PR #75 (GB300 disagg, TP=4 only).
- Per-worker `sglang_config` is hand-rolled with explicit citations in each recipe header pointing back to PR #69 / PR #75 / cookbook. Topology mirrors #1129 vLLM siblings — same node counts at 1k/1k 1p1d/3p1d. (At 1k/1k mid the vLLM sibling's wider DP=16 decode is collapsed to TP=8 since DeepEP is dead in this image.)

## Test plan

- [x] `python3 utils/matrix_logic/generate_sweep_configs.py test-config --config-files .github/configs/nvidia-master.yaml --config-keys dsv4-fp4-gb200-dynamo-sglang` expands to 6 entries / 27 conc points
- [x] All 6 recipe YAMLs parse via `yaml.safe_load`
- [x] `bash -n runners/launch_gb200-nv.sh` passes
- [x] `perf-changelog.yaml` diff vs main is additions-only (gate passes)
- [ ] First-run smoke test on GB200: confirm `--moe-runner-backend flashinfer_mxfp4` + DP-attention + no DeepEP combination loads without OOM at low-conc (1p1d-dep8-tep8); may require tuning `mem-fraction-static` or `cuda-graph-max-bs` if first-run hits memory pressure.


ch-wan:
Eval is missing here because srt-slurm pin (9d75f82) lacks the lm-eval orchestrator path (check this comment: https://github.com/SemiAnalysisAI/InferenceX/pull/1232/changes/be12dbaceef84726e94751d4e70e25d8d45d5b8e), but the accuracy has been manually verified. We will provide the eval result immediately once this PR (https://github.com/NVIDIA/srt-slurm/pull/122) is landed.

ishandhanani:
Merged LM eval

PR #1175, opened 2026-04-26
Add workflow inputs for moe-runner-backend, chunked-prefill-size, swa-full-tokens-ratio

yhyang201:
## Summary
- Add optional `moe-runner-backend`, `chunked-prefill-size`, and `swa-full-tokens-ratio` inputs to `benchmark-tmpl.yml` (all default to empty string)
- Map them to `MOE_RUNNER_BACKEND`, `CHUNKED_PREFILL_SIZE`, `SWA_FULL_TOKENS_RATIO` env vars
- Pass these fields through in `e2e-tests.yml` and `run-sweep.yml` `with:` blocks

## Why
The `issue_comment`-triggered `/sweep` workflow always uses workflow files from `main`. PR branches that add configs with these knobs (e.g. `knob_backup`) fail because the workflow files on `main` don't define or pass these inputs, so the env vars are empty at runtime.

This is a backwards-compatible change — all new inputs are `required: false` with `default: ''`, so existing configs are unaffected.

## Test plan
- [ ] Verify existing sweeps still work (no input is required, defaults to empty)
- [ ] After merge, `/sweep` on PRs using `moe-runner-backend` / `chunked-prefill-size` / `swa-full-tokens-ratio` should pass the values through correctly

### DIFF
--- a/.github/workflows/benchmark-tmpl.yml
+++ b/.github/workflows/benchmark-tmpl.yml
@@ -38,6 +38,21 @@ on:
       dp-attn:
         required: true
         type: boolean
+      moe-runner-backend:
+        description: "Optional --moe-runner-backend value (e.g. flashinfer_mxfp4). Empty = unset; the benchmark script decides whether to require it."
+        required: false
+        type: string
+        default: ''
+      chunked-prefill-size:
+        description: "Optional --chunked-prefill-size value. Empty = unset; the benchmark script decides whether to require it."
+        required: false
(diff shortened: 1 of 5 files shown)


cquil11:
Hey btw I don't think these are params we wanna add to core changelog entries. The top level config plumbing is supposed to be for core runtime parameters (like TP, EP, DP, etc). Not more specific things like thsi

PR #1189, opened 2026-04-26
Nothing to see here

Oseltamivir:
## Summary
- Lowers `MIN_EVAL_CONC` from 16 → 8
- Adds `evals-only` perf-changelog entry for `dsv4-fp4-b300-sglang-mtp`
- Updates the test assertion for multinode eval-conc median to reflect the new threshold


Qiaolin-Yu (changes requested):
Why do we need to change this?

Oseltamivir:
depriortising evals for now

PR #1204, opened 2026-04-28
dsv4-fp4-b200-vllm: bump to vllm v0.20.0, deep_gemm_mega_moe MoE

functionstackx:
## Summary
- Pin `dsv4-fp4-b200-vllm` image to `vllm/vllm-openai:v0.20.0-cu130` (canonical v0.20.0 tag) — replaces the floating `deepseekv4-cu130` tag. DeepGEMM is preinstalled in this image, so no `install_deepgemm.sh` step is needed at benchmark time.
- All configs share `--compilation-config '{"cudagraph_mode":"FULL_AND_PIECEWISE","custom_ops":["all"]}'` and `--attention_config.use_fp4_indexer_cache=True`.
- **Gate `--moe-backend deep_gemm_mega_moe` and `--gpu-memory-utilization 0.85` on `DP_ATTENTION=true`** per the v0.20.0 recipe. The two flags travel together via `MOE_ARGS` / `GMU_ARGS`.
- Drop `--pipeline-parallel-size 1` (no longer in the v0.20.0 recipe). Retain `--no-enable-prefix-caching` (so cross-request cache hits don't skew steady-state throughput) and `--max-cudagraph-capture-size 2048` (preserve cudagraph coverage at the top of the search space).

### Resulting flag matrix

| search-space entry | DP_ATTENTION | EP | compile-config | moe-backend | gpu-mem-util |
| --- | --- | --- | --- | --- | --- |
| `{tp:8}` | false | – | `FULL_AND_PIECEWISE,custom_ops:all` | (default) | (default) |
| `{tp:8, ep:8}` (1k1k @ 128) | false | enabled | `FULL_AND_PIECEWISE,custom_ops:all` | (default) | (default) |
| `{tp:8, ep:8, dp-attn:true}` | true | enabled | `FULL_AND_PIECEWISE,custom_ops:all` | `deep_gemm_mega_moe` | `0.85` |

`PARALLEL_ARGS` / `EP_ARGS` / `GMU_ARGS` / `MOE_ARGS` arrays are preserved so the existing `DP_ATTENTION` / `EP_SIZE` search-space branching drives the right flags cleanly.

Adds a `perf-changelog.yaml` entry to trigger the affected configs.

## Test plan
- [ ] Trigger the `dsv4-fp4-b200-vllm` benchmark workflow on a B200 runner and confirm the engine starts and the sweep completes for at least one cell in each row of the matrix above.
- [ ] Confirm the `v0.20.0-cu130` image pulls and DeepGEMM is already importable inside the container (no install step at runtime).
- [ ] Spot-check `server.log` to verify the right flags are logged per branch:
  - DP_ATTENTION=false (any CONC, any EP): no `--moe-backend`, no `--gpu-memory-utilization`.
  - DP_ATTENTION=true: `--moe-backend deep_gemm_mega_moe --gpu-memory-utilization 0.85`.
  - All branches: `--compilation-config '{"cudagraph_mode":"FULL_AND_PIECEWISE","custom_ops":["all"]}'`, `--no-enable-prefix-caching`, `--max-cudagraph-capture-size 2048`.

### DIFF
--- a/benchmarks/single_node/dsv4_fp4_b200_vllm.sh
+++ b/benchmarks/single_node/dsv4_fp4_b200_vllm.sh
@@ -23,6 +23,9 @@ fi
 
 nvidia-smi
 
+# install_deepgemm.sh git-clones the DeepGEMM repo, but the v0.20.0
+# vllm image ships without git — install it first.
+apt-get update && apt-get install -y --no-install-recommends git
 bash <(curl -fsSL https://raw.githubusercontent.com/vllm-project/vllm/v0.20.0/tools/install_deepgemm.sh)
 
 hf download "$MODEL"
--- a/benchmarks/single_node/dsv4_fp4_b200_vllm.sh
+++ b/benchmarks/single_node/dsv4_fp4_b200_vllm.sh
@@ -76,6 +76,7 @@ vllm serve "$MODEL" --host 0.0.0.0 --port "$PORT" \
     --trust-remote-code \
     --kv-cache-dtype fp8 \
     --block-size 256 \
+    --no-enable-prefix-caching \
     "${PARALLEL_ARGS[@]}" \
     "${EP_ARGS[@]}" \
     "${GMU_ARGS[@]}" \
--- a/benchmarks/single_node/dsv4_fp4_b200_vllm.sh
+++ b/benchmarks/single_node/dsv4_fp4_b200_vllm.sh
@@ -23,8 +23,6 @@ fi
 
 nvidia-smi
 
-bash <(curl -fsSL https://raw.githubusercontent.com/vllm-project/vllm/v0.20.0/tools/install_deepgemm.sh)
-
 hf download "$MODEL"
 
 SERVER_LOG=/workspace/server.log
--- a/benchmarks/single_node/dsv4_fp4_b200_vllm.sh
+++ b/benchmarks/single_node/dsv4_fp4_b200_vllm.sh
@@ -90,6 +90,7 @@ vllm serve "$MODEL" --host 0.0.0.0 --port "$PORT" \
     --tool-call-parser deepseek_v4 \
     --enable-auto-tool-choice \
     --reasoning-parser deepseek_v4 \
+    --max-cudagraph-capture-size 2048 \
     --max-model-len "$SERVE_MAX_MODEL_LEN" \
     --max-num-batched-tokens "$MAX_NUM_BATCHED_TOKENS" > "$SERVER_LOG" 2>&1 &
 
--- a/benchmarks/single_node/dsv4_fp4_b200_vllm.sh
+++ b/benchmarks/single_node/dsv4_fp4_b200_vllm.sh
@@ -42,19 +42,14 @@ if [ "${EP_SIZE:-1}" -gt 1 ]; then
     EP_ARGS=(--enable-expert-parallel)
 fi
 
-# DP-attn switches both the MoE backend and the cudagraph mode per the
-# vLLM v0.20.0 DeepSeek-V4-Pro recipe:
-#   - TP-only / TP+EP (DP_ATTENTION=false): default MoE backend,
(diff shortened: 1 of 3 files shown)


functionstackx:
talked to vllm maintainer esmeetu and he said that deepgemm extra install only needed for pip wheels and docker already has deepgemm megamoes kernels

image

functionstackx:
https://github.com/vllm-project/recipes/pull/409/changes

benchmarks/single_node/dsv4_fp4_b200_vllm.sh
```diff
@@ -42,9 +42,19 @@ if [ "${EP_SIZE:-1}" -gt 1 ]; then
...
 
+# DP-attn switches both the MoE backend and the cudagraph mode per the
+# vLLM v0.20.0 DeepSeek-V4-Pro recipe:
+#   - TP-only / TP+EP (DP_ATTENTION=false): default MoE backend,
+#     compile-config = {"mode": 0, "cudagraph_mode": "FULL_DECODE_ONLY"}
+#   - DP-attn + EP   (DP_ATTENTION=true):  --moe-backend deep_gemm_mega_moe,
+#     compile-config = {"cudagraph_mode": "FULL_AND_PIECEWISE", "custom_ops": ["all"]}
 GMU_ARGS=()
+MOE_ARGS=()
+COMPILATION_CONFIG='{"mode": 0, "cudagraph_mode": "FULL_DECODE_ONLY"}'
```
wzhao18: From the results, it seems DEP configs with concurrency 256 and above are faster. And the TP/TEP configs are slower. I suspect `COMPILATION_CONFIG='{"mode": 0, "cudagraph_mode": "FULL_DECODE_ONLY"}'` causes the slowdown.

Suggestions for a new sweep:
- Can we use `FULL_AND_PIECEWISE` for all configs?
- Can we Add `--moe-backend deep_gemm_mega_moe` when DP and concurrency >= 256?

functionstackx: on it, will try with that

the reasoning why i tried decode only cudagraph for TEP/TP is this is what the vllm recipes, i will PR to vllm recipes too once i try doing FULL_AND_PIECEWISE for tp/tep too

Image

wzhao18: btw thanks very much for trying this out. Really appreciate it.

benchmarks/single_node/dsv4_fp4_b200_vllm.sh
```diff
@@ -42,9 +42,14 @@ if [ "${EP_SIZE:-1}" -gt 1 ]; then
...
 fi
 
+# Mega-MoE backend and the lower GMU only kick in for high-concurrency
+# DP-attn runs (CONC >= 256), per the vLLM v0.20.0 DeepSeek-V4-Pro recipe.
+# All configs share the FULL_AND_PIECEWISE compilation config.
 GMU_ARGS=()
-if [ "${DP_ATTENTION}" = "true" ]; then
+MOE_ARGS=()
+if [ "${DP_ATTENTION}" = "true" ] && [ "${CONC}" -ge 256 ]; then
     GMU_ARGS=(--gpu-memory-utilization 0.85)
```
wzhao18, line 51: may need this for all DEP configs.

functionstackx: ack

wzhao18: From the last sweep, the mega moe backend seems to bring slightly worse perf for DEP with < 256 concurrency. Should we add the 256 concurrency check back?

functionstackx: it seems like megamoe performs similarity for dep with < 256 conc (within the error of margin of this cluster in japan)


image

functionstackx:
image
image

PR #1218, opened 2026-04-29
[NV] Add GB200 MegaMOE max throughput recipe

alec-flowers:
## Summary
- Add a GB200 Dynamo vLLM MegaMOE max-throughput srt-slurm recipe for DeepSeek-V4-Pro at conc=4096.
- Wire the recipe into `dsv4-fp4-gb200-dynamo-vllm` as a fourth 8k/1k point alongside low-latency, mid-curve, and offload max-tpt.
- Append a perf changelog entry describing the MegaMOE/no-offload runtime shape.

## Validation
- `git diff --check`
- `python3 utils/matrix_logic/generate_sweep_configs.py full-sweep --config-files .github/configs/nvidia-master.yaml --model-prefix dsv4 --precision fp4 --framework dynamo-vllm --runner-type gb200 --seq-lens 8k1k --multi-node --no-evals`
- `python3 utils/process_changelog.py --base-ref origin/main --head-ref HEAD --changelog-file perf-changelog.yaml --trim-conc`

### DIFF
--- a/.github/configs/nvidia-master.yaml
+++ b/.github/configs/nvidia-master.yaml
@@ -7639,7 +7639,7 @@ dsv4-fp4-gb200-dynamo-vllm:
   - isl: 8192
     osl: 1024
     search-space:
-    # Three validated 8k/1k points mirrored from NVIDIA/srt-slurm
+    # Validated 8k/1k points mirrored from NVIDIA/srt-slurm
     # aflowers/vllm-gb200-v0.20.0 history. conc-list values match each
     # recipe's benchmark.concurrencies.
 
@@ -7685,6 +7685,22 @@ dsv4-fp4-gb200-dynamo-vllm:
         dp-attn: true
         additional-settings:
         - "CONFIG_FILE=recipes/vllm/deepseek-v4/8k1k/disagg-gb200-max-tpt.yaml"
+      decode:
+        num-worker: 1
+        tp: 8
+        ep: 8
+        dp-attn: true
+
+    # MegaMOE max throughput: same 3 prefill (DEP=8 each) + 1 decode (DEP=8)
+    # shape, but uses deep_gemm_mega_moe on both workers and disables offload.
+    - conc-list: [4096]
+      prefill:
+        num-worker: 3
+        tp: 8
+        ep: 8
+        dp-attn: true
+        additional-settings:
+        - "CONFIG_FILE=recipes/vllm/deepseek-v4/8k1k/disagg-gb200-max-tpt-megamoe.yaml"
       decode:
         num-worker: 1
         tp: 8
--- a/.github/configs/nvidia-master.yaml
+++ b/.github/configs/nvidia-master.yaml
@@ -7659,6 +7659,22 @@ dsv4-fp4-gb200-dynamo-vllm:
         ep: 1
         dp-attn: false
 
+    # Low-middle curve: 1 prefill (DEP=8) + 4 decode (TP=8). 11 nodes total
+    # with a dedicated NATS/etcd infra node.
+    - conc-list: [256, 512]
+      prefill:
+        num-worker: 1
+        tp: 8
+        ep: 8
+        dp-attn: true
+        additional-settings:
+        - "CONFIG_FILE=recipes/vllm/deepseek-v4/8k1k/disagg-gb200-low-middle-curve.yaml"
+      decode:
+        num-worker: 4
+        tp: 8
+        ep: 1
+        dp-attn: false
+
     # Mid curve: 1 prefill (DEP=8) + 1 decode (DEP=8). 5 nodes total with
     # a dedicated NATS/etcd infra node.
     - conc-list: [256]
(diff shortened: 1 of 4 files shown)


alec-flowers:
total tok/s/GPU | 8957.06
tok/s/user | 16.41

kedarpotdar-nv (approved):

.github/configs/nvidia-master.yaml
```diff
@@ -7690,3 +7690,19 @@ dsv4-fp4-gb200-dynamo-vllm:
         tp: 8
         ep: 8
         dp-attn: true
+
+    # MegaMOE max throughput: same 3 prefill (DEP=8 each) + 1 decode (DEP=8)
+    # shape, but uses deep_gemm_mega_moe on both workers and disables offload.
+    - conc-list: [4096]
+      prefill:
+        num-worker: 3
+        tp: 8
```
cquil11, line 7715: for my own understanding, isn't this technically tp1 dep8 (according to the srt slurm recipe)

from the vllm [docs](https://docs.vllm.ai/en/latest/serving/expert_parallel_deployment/#configuration) it appears that ep = tp * dp. so in this case are any of the weights even shareded tp? 

sorry if this is a dumb question.

cquil11:
@alec-flowers instead of just 3 data points, can you extend to highlight more parts of the curve so we are not overly relying on interpolation (which is imperfect in some cases)?

it is nice to have _at least_ 6-7 points on the curve

benchmarks/multi_node/srt-slurm-recipes/vllm/deepseek-v4/8k1k/disagg-gb200-max-tpt-megamoe.yaml
```diff
@@ -0,0 +1,154 @@
...
+      served-model-name: "deepseek-ai/DeepSeek-V4-Pro"
+      kv-cache-dtype: "fp8"
+      tensor-parallel-size: 1
+      pipeline-parallel-size: 1
+      data-parallel-size: 8
+      data-parallel-rpc-port: 13345
+      enable-expert-parallel: true
+      enable-ep-weight-filter: true
+      moe-backend: deep_gemm_mega_moe
+      max-model-len: 9280
```
cquil11, line 124: I think we usually add a bit more of a buffer? not sure how much it matters

typically I think isl + osl + 200

.github/configs/nvidia-master.yaml
alec-flowers: think its historical reason. tp8/dp-attn=true was used as the matrix label, but the vLLM script translated that into TP=1, DP=8, EP=8. For srt-slurm multi-node recipes, the recipe YAML is now the source of truth, but the label convention carried over from that older InferenceX matrix encoding.

alec-flowers:
> @alec-flowers instead of just 3 data points, can you extend to highlight more parts of the curve so we are not overly relying on interpolation (which is imperfect in some cases)?
> 
> it is nice to have _at least_ 6-7 points on the curve

Yes for sure. But can we get that in in the coming days. The long pole here is this submission process and queuing and I had jobs that got edited and cancelled that had 6-7 points but that takes like 10 hours to run. So what I'm trying to do now is get a baseline and then sort of fill it in. 

The goal is to get the full pareto and I'm working hard on it.

benchmarks/multi_node/srt-slurm-recipes/vllm/deepseek-v4/8k1k/disagg-gb200-max-tpt-megamoe.yaml
alec-flowers: 9216 - I did leave a bit. Shouldn't be needed unless something is going wrong somewhere.

.github/configs/nvidia-master.yaml
Oseltamivir, line 7680: @alec-flowers  is it possible to just add 1 more point between 256 and 1? Extremely large gap.

Oseltamivir (changes requested):

.github/configs/nvidia-master.yaml
alec-flowers: Sure! Let me get a config up.

alec-flowers: @Oseltamivir  I have it pushed. This will fill in two points between 1 and 256. Are you ok merging this in? 

We actually already tested this earlier and it passed in the previous pipeline. But as mentioned it took so long to run especially these jobs that take up more nodes that I removed it.

Oseltamivir: Thanks

Oseltamivir (approved):
lgtm

.github/configs/nvidia-master.yaml
alec-flowers: You can check that there are more recipes coming - https://github.com/NVIDIA/srt-slurm/pull/103/changes

Its just a matter of getting them through all this CI.

.github/configs/nvidia-master.yaml
cquil11: gotcha. this is on us then I suppose
FYI the values here are primarily meant for transfer to frontend,

PR #1219, opened 2026-04-29
Haichen/kimi k2.5 int4 mtp vllm v0.20.0/modelrunnerV2

haic0:


functionstackx:

PR #1222, opened 2026-04-29
Add DSv4 FP8 H200 vLLM MTP benchmark

functionstackx:
## Summary
Ports the H200 STP recipe to MTP.

- New `dsv4-fp8-h200-vllm-mtp` config + `benchmarks/single_node/dsv4_fp8_h200_mtp.sh` script.
- MTP counterpart of `dsv4-fp8-h200-vllm`: identical launch flags (EP + DP=$TP, `--gpu-memory-utilization 0.95`, `--max-num-seqs 512`, `--no-enable-flashinfer-autotune`, FULL_DECODE_ONLY compile) with one addition — `--speculative-config '{"method":"mtp","num_speculative_tokens":1}'`.
- `num_speculative_tokens=1` because the recipe states H200 supports spec token=1 only for now (B200 / Blackwell SKUs are where token=2 lands on the Pareto front per @wzhao18's vLLM Blackwell MTP work; the H200 kernel coverage is still token=1).
- Image: `vllm/vllm-openai:v0.20.1` (pinned by SHA256 digest `9eff9734...23aa4`). The non-MTP `dsv4-fp8-h200-vllm` entry is unchanged and still on the `deepseekv4-cu129` tag.
- `max-model-len` comes from the runner's `$MAX_MODEL_LEN` env var rather than being hardcoded to 800k — keeps this script consistent with the rest of the H200 fleet.
- Sets `VLLM_MEMORY_PROFILER_ESTIMATE_CUDAGRAPHS=0` to skip the cudagraph-memory estimator during worker memory profiling — it overestimates and pushes us over the GPU memory budget on H200 + MTP, even though the actual cudagraph capture works fine.
- `run_benchmark_serving` invocation passes `--dsv4` so prompts get chat-formatted encoding, per the AGENTS.md MTP rule (raw random tokens silently regress EAGLE-style acceptance).
- Search space mirrors the non-MTP H200 entry: TP=8, EP=8, DP-attn=true, CONC 4-64, both 1k1k and 8k1k, with `spec-decoding: mtp` on each entry.
- Adds a `perf-changelog.yaml` entry to trigger the new config.

## Test plan
- [ ] Trigger the `dsv4-fp8-h200-vllm-mtp` benchmark workflow on an H200 runner and confirm the engine starts and the sweep completes for at least one cell from each of the two `seq-len-configs`.
- [ ] Confirm `vllm/vllm-openai:v0.20.1@sha256:9eff9734...23aa4` pulls cleanly.
- [ ] `server.log` shows `--speculative-config '{"method":"mtp","num_speculative_tokens":1}'` and the rest of the H200 recipe flags.
- [ ] Acceptance rate is in a sane range — `--dsv4` is wired into `run_benchmark_serving` so the prompts go through the chat template.


functionstackx:
waiting for 0.20.1 https://github.com/vllm-project/vllm/pull/41189 and https://github.com/vllm-project/vllm/pull/41444

PR #1235, opened 2026-04-30
Haichen/kimi k2.5 int4 mtp v0.19.1

haic0:


functionstackx:
Still waiting for @haic0 's AR distribution on what to align to. Blocker is AMD

PR #1236, opened 2026-04-30
[AMD] improve dsr1 fp4 disagg perf on mi355x - rapid follow up PR incoming to quant correction on fp8 combine 

billishyahao:
## inital merge - - rapid follow up PR incoming to quant correction on fp8 combine 

    # NOTE: that currently with fp8_combine set, the evals do not pass on InferenceX eval harness
    # or on SGLang native harness for high concurrency 4k and gets no where near the golden score of
    # 0.95 on even basic GSM8k grade school math as confirmed by @billishyahao from AMD
    # and as confirmed by @Oseltamivir. This was initally merged with @billishyahao promising 
    # that an fast follow PR to fix the evals via having quant correction in the fp8 combine

replacement of https://github.com/SemiAnalysisAI/InferenceX/pull/983

The new patch is adding the following optimization:

    - "Bump SGL mori image to lmsysorg/sglang-rocm"
    - "Add more high tput / low latency sweep configs"
    - "Enable v2 mxfp4 DSR1 0528 model"
    - "Enable fp4 disp / fp8 combine feature on mori"
    - "Enable Mori SDMA + two batch overlapping feature"


billishyahao:
Can we got the review for this patch ? @functionstackx @Oseltamivir @cquil11 

Sweep 19 of 20 passed, 1 is canceled by user  

https://github.com/SemiAnalysisAI/InferenceX/actions/runs/25241387090

Eval all passed 

https://github.com/SemiAnalysisAI/InferenceX/actions/runs/25268431600/

benchmarks/multi_node/amd_utils/server.sh
```diff
@@ -327,6 +338,22 @@ if [[ -n "$MODEL_NAME" ]]; then
     echo "Using model-specific configuration for: $MODEL_NAME"
 fi
 
+if [[ "${EVAL_ONLY:-false}" == "true" ]] || [[ "${RUN_EVAL:-false}" == "true" ]]; then
+    PREFILL_SERVER_CONFIG=$(echo "$PREFILL_SERVER_CONFIG" | sed 's/--ep-dispatch-algorithm fake//g')
+    DECODE_SERVER_CONFIG=$(echo "$DECODE_SERVER_CONFIG" | sed 's/--ep-dispatch-algorithm fake//g')
+
+    if [[ "$DECODE_ENABLE_DP" != "true" ]] && [[ "$DECODE_MTP_SIZE" -gt 0 ]]; then
+        DECODE_SERVER_CONFIG=$(echo "$DECODE_SERVER_CONFIG" | sed 's/--speculative-num-steps [0-9]*/--speculative-num-steps 3/; s/--speculative-num-draft-tokens [0-9]*/--speculative-num-draft-tokens 4/')
+    fi
```
functionstackx: @billishyahao can we not have "if evals then set xyz env vars" besides increasing context len

functionstackx (changes requested):
added an comment related to ur current code of "if evals: set xyz"

benchmarks/multi_node/amd_utils/server.sh
```diff
@@ -327,6 +338,22 @@ if [[ -n "$MODEL_NAME" ]]; then
...
+
+    # Increase context-length for eval to accommodate R1 thinking chains
+    PREFILL_SERVER_CONFIG=$(echo "$PREFILL_SERVER_CONFIG" | sed 's/--context-length [0-9]*/--context-length 16384/')
+
+    unset MORI_MOE_MAX_INPUT_TOKENS_PREFILL
+    unset MORI_MOE_MAX_INPUT_TOKENS_DECODE
+    unset SGLANG_MORI_FP8_COMB
```
functionstackx: @billishyahao same thing here

benchmarks/multi_node/amd_utils/server.sh
billishyahao: removed context length and change mtp to 2

benchmarks/multi_node/amd_utils/server.sh
```diff
@@ -327,6 +338,15 @@ if [[ -n "$MODEL_NAME" ]]; then
     echo "Using model-specific configuration for: $MODEL_NAME"
 fi
 
+if [[ "${EVAL_ONLY:-false}" == "true" ]] || [[ "${RUN_EVAL:-false}" == "true" ]]; then
+    PREFILL_SERVER_CONFIG=$(echo "$PREFILL_SERVER_CONFIG" | sed 's/--ep-dispatch-algorithm fake//g')
+    DECODE_SERVER_CONFIG=$(echo "$DECODE_SERVER_CONFIG" | sed 's/--ep-dispatch-algorithm fake//g')
+    unset MORI_MOE_MAX_INPUT_TOKENS_PREFILL
+    unset MORI_MOE_MAX_INPUT_TOKENS_DECODE
+    unset SGLANG_MORI_FP8_COMB
```
functionstackx: @billishyahao I don't understand why we are unsetting fp8 combine for evals only but using can we not performance benchmark. 

It seems like the only thing we should change for evals specific is context len to fit the shots and not setting fp8 combine.


can you work with @Oseltamivir to figure it out? happy to dedicate time on our end to work with you on it

billishyahao: FP8 combine looks fine for 
python benchmark/gsm8k/bench_sglang.py --num-questions 1300 --port 30000

may need more debugging from @Oseltamivir 

BTW since this pr is based on March old PR + switching upstream sglang PR. Eval is a new feature needs more time to address. Can we merge this first and then use follow-up PR for addressing eval issue?

billishyahao: image

functionstackx: > FP8 combine looks fine for
python benchmark/gsm8k/bench_sglang.py --num-questions 1300 --port 30000

@billishyahao even in your local bench 91.8% on GSM8k is quite low and does not look fine for deepseekv3 R1, we are seeing 95-96% for deepseekv3 R1 on grade school math

functionstackx: we can potentially merge this and have fixing it be an follow up PR but would like to do a couple days of work between you @billishyahao & @Oseltamivir before we merge this. 

Your local sglang bench (not using inferencex harness) is quite low at 91%

billishyahao: The current accuracy drop with fp8 combine is expected as we have not introduced quant factor to retain the preicision. But huge drop from 0.915 to 0.485 is yet another issue from harness.

billishyahao:
Latest full sweep result：
https://github.com/SemiAnalysisAI/InferenceX/actions/runs/25355166460

benchmarks/multi_node/amd_utils/server.sh
functionstackx: @billishyahao dropping from 0.95 to 0.915 on gsm8k is already an huge issue. as your aware, dropping 3 points on grade school math 8k has huge effects on downstream evals like swe-bench or has huge downstream evals when the harness template prompt changes slightly. 

as suggested above, i think it would be good for @Oseltamivir and yall to debug for the next few days,
#1 first debug the 0.95 to 0.915 drop on sgl harness template prompt
#2  then next debug the 0.915 to 0.485 drop on the secondary prompt template and whether or not the bug is from the secondary prompt template or an downstream bug in amd's code where the root cause is there is not an quant factor to retain the precision with fp8 combine

functionstackx: @billishyahao can u take a look at this? @Oseltamivir  is currently unable to reproduce the you described and got 90+ without unsetting fp8 combine 

https://github.com/SemiAnalysisAI/InferenceX/actions/runs/25395388503

billishyahao: Hi @functionstackx @Oseltamivir  this eplb only takes effect when EP is enabled. Job [25395388503](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/25395388503) is only evaluating TP only case.

Oseltamivir (approved):
lgtm

PR #1261, opened 2026-05-02
Fix GPT-OSS ATOM config schema

Oseltamivir:


seungrokj:
@Oseltamivir  thank you for fixing this. I missed this 👍

PR #1263, opened 2026-05-02
fix(dsv4): gate h200 reasoning parser flag

OCWC22:
## Summary
- gate the H200 DSv4 `--reasoning-parser deepseek_v4` flag behind `ENABLE_DSV4_REASONING_PARSER=true`
- keep the parser disabled by default because the AgentX testing branch's `docs/AGENTIC_TEST_RESULTS.md` records `dsv4-fp8-h200-vllm` as ❌ FAILED with the DeepSeek-V4 reasoning parser issue (`reasoning_start_str` / `reasoning_end_str` missing)
- preserve an opt-in path for future vLLM parser fixes without changing the default launcher behavior

## Smoke test
- `bash -n benchmarks/single_node/dsv4_fp8_h200.sh` passes
- default `ENABLE_DSV4_REASONING_PARSER` expansion leaves `REASONING_PARSER_ARGS` empty, so the launcher omits `--reasoning-parser deepseek_v4` unless explicitly enabled


functionstackx:
stale PR. closing for now

PR #1268, opened 2026-05-03
B200 TRTLLM test

Oseltamivir:


benchmarks/single_node/dsv4_fp4_b200_trt.sh
```diff
@@ -0,0 +1,164 @@
...
+    MAX_NUM_TOKENS="$EVAL_MAX_MODEL_LEN"
+fi
+
+if [[ "${RUN_EVAL:-false}" == "true" || "${EVAL_ONLY:-false}" == "true" ]]; then
+    # DeepSeek-V4-Pro has hidden size 7168. The current TRTLLM fused-HC MHC
+    # path corrupts eval generations for this shape; keep eval servers on the
+    # unfused path until the fused kernel is guarded or supports 7168.
+    export TRTLLM_MHC_ENABLE_FUSED_HC="${TRTLLM_MHC_ENABLE_FUSED_HC:-0}"
+    echo "TRTLLM_MHC_ENABLE_FUSED_HC: $TRTLLM_MHC_ENABLE_FUSED_HC"
+fi
```
functionstackx: @claude these env var & echo should be set no matter what & gated behind if evals

Klaud-Cold: **Claude finished @functionstackx's task in 1m 13s** —— [View job](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/25268898289)

---
### Restructuring env vars & echo outside the evals gate

- [x] Read the current file to understand the structure around line 111
- [x] Identify which env vars/echo should be unconditional vs. evals-only
- [x] Implement the change
- [x] Push the fix

Moved `export TRTLLM_MHC_ENABLE_FUSED_HC` and its echo out of the `RUN_EVAL`/`EVAL_ONLY` `if` block so they're set unconditionally. Pushed in [`40fd0a98`](https://github.com/SemiAnalysisAI/InferenceX/commit/40fd0a98).

PR #1279, opened 2026-05-04
dsv4-fp8-h200-vllm-mtp: bump num_speculative_tokens 1 → 2

functionstackx:
## Summary
Bump `num_speculative_tokens` from 1 to 2 in the `dsv4-fp8-h200-vllm-mtp` recipe.

- Single change: `--speculative-config '{"method":"mtp","num_speculative_tokens":1}'` → `'{"method":"mtp","num_speculative_tokens":2}'` in `benchmarks/single_node/dsv4_fp8_h200_mtp.sh`.
- Companion comment in the script header is updated to match.
- Adds a `perf-changelog.yaml` entry to trigger the new sweep.

## Context
The original PR (#1222) landed at `num_speculative_tokens=1` because the recipe says H200 MTP kernels only support 1 draft token. Inferact folks indicated H200 kernels now support 2, and Blackwell MTP runs at 2 per @wzhao18's vLLM Blackwell MTP submission. This PR re-tests that assumption on H200. If acceptance / throughput look healthy, we keep 2; otherwise we revert.

## Test plan
- [ ] Trigger the `dsv4-fp8-h200-vllm-mtp` benchmark workflow on an H200 runner and confirm the engine starts and the sweep completes for at least one cell from each of the two `seq-len-configs`.
- [ ] `server.log` shows `--speculative-config '{"method":"mtp","num_speculative_tokens":2}'`.
- [ ] Acceptance rate is in a sane range and throughput at the same concurrency points beats the `num_speculative_tokens=1` numbers from the #1222 sweep — otherwise the kernel-support assumption was wrong and we revert.

### DIFF
--- a/benchmarks/single_node/dsv4_fp8_h200_mtp.sh
+++ b/benchmarks/single_node/dsv4_fp8_h200_mtp.sh
@@ -2,7 +2,7 @@
 
 # DeepSeek-V4-Pro H200 vLLM MTP variant of the recipe at
 # https://vllm.ai/blog/deepseek-v4. Mirrors dsv4_fp8_h200.sh but adds
-# --speculative-config '{"method":"mtp","num_speculative_tokens":1}' and
+# --speculative-config '{"method":"mtp","num_speculative_tokens":2}' and
 # routes prompts through chat-formatted encoding via --dsv4 (required for
 # meaningful MTP acceptance numbers per AGENTS.md).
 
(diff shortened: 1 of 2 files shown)


functionstackx:
thanks to inferact vllm community maintainers for fixing hopper top_k kernel and merging my pr for recipes https://github.com/vllm-project/recipes/pull/435

PR #1289, opened 2026-05-06
Update kimi k2.5 mtp support by using modelrunnerV2 and vllm v0.20.1

haic0:


functionstackx:

PR #1290, opened 2026-05-06
Haichen/kimi k2.5 int4 mtp v0.20.1 mrv2 off

haic0:
git commit -s -m "Update kimi k2.5 mtp support by disabling modelrunnerV2 and vllm v0.20.1"


functionstackx:

PR #1292, opened 2026-05-06
Re-run qwen3.5-fp8-b200-sglang-mtp sweep after B200 partition change

Ankur-singh:
Re-running the qwen3.5-fp8-b200-sglang-mtp sweep because the B200 DGXC Slurm partition was changed (gpu → gpu-2).

Re-trigger didn't help because it used the same commit with wrong partition name. 

Old sweep based of `main` https://github.com/SemiAnalysisAI/InferenceX/actions/runs/25342967860


functionstackx (approved):
lgtm

PR #433, opened 2026-01-15
[AMD] feat: support multiple cluster for mi355x DI CI workflow

billishyahao:
This patch is to support multiple cluster for MI355X distributed inference CI workflow.
This patch co-work with the branch sa-260114 of recipe  https://github.com/billishyahao/sglang_disagg.git
Co-author: [@ichbinblau](https://github.com/ichbinblau)

PR #1373, opened 2026-05-13
Improve dsv4-fp8-mi355x-vllm with vllm-project/recipes#433 MI355X recipe

Oseltamivir:
## Summary

Apply the validated DeepSeek-V4-Pro MI355X recipe from [vllm-project/recipes#433](https://github.com/vllm-project/recipes/pull/433) (which in turn sources its serving config from [vllm-project/vllm#40871](https://github.com/vllm-project/vllm/pull/40871)) to the existing `dsv4-fp8-mi355x-vllm` benchmark. The base vLLM build (PR #40889 with AITER-accelerated sparse MLA decode, pinned in `dsv4_fp8_mi355x_vllm.sh`) is unchanged; only serving flags / env vars / sweep range are updated.

### Server-side changes

| | Before | After (recipe #433) |
| --- | --- | --- |
| `VLLM_ROCM_USE_AITER_LINEAR` | unset | `1` |
| `--distributed-executor-backend` | (default) | `mp` |
| `--max-num-batched-tokens` | (default) | `8192` |
| `--async-scheduling` | off | on |
| `--gpu-memory-utilization` | `0.90` | `0.6` |
| `--max-num-seqs` | `32` | `128` |
| `--tool-call-parser deepseek_v4` | on | dropped |
| `--enable-auto-tool-choice` | on | dropped |

Tool-call flags were removed because the recipe omits them and the throughput benchmark does not exercise tool calling. All other existing flags (`--kv-cache-dtype fp8`, `--moe-backend triton_unfused`, `--enforce-eager`, `--no-enable-prefix-caching`, `--tokenizer-mode deepseek_v4`, `--reasoning-parser deepseek_v4`) are preserved.

### Sweep changes

`dsv4-fp8-mi355x-vllm` was previously pinned to `conc=1` only (single-sequence marker). With `--max-num-seqs=128` validated by the recipe, the sweep is expanded to `conc 4-64` for both `1k1k` and `8k1k`, matching `dsv4-fp8-mi355x-sglang` so vLLM↔SGLang results are directly comparable on the same MI355X runner.

Validated locally:
```
$ python utils/matrix_logic/generate_sweep_configs.py full-sweep \
    --config-files .github/configs/amd-master.yaml \
    --framework vllm --runner-type mi355x
# 10 dsv4-fp8-mi355x-vllm configs generated (5 per ISL/OSL, tp=8, conc 4..64)
```

## Test plan

- [ ] Apply `sweep-enabled` label so `run-sweep.yml` exercises the new flags end-to-end at trimmed concurrency (highest `conc` per parallelism config).
- [ ] Confirm the AITER MLA decode build (vLLM PR #40889 SHA `b3a4a44`) still installs cleanly with the new server flags.
- [ ] Inspect `agg_bmk.json` for `dsv4` mi355x-vllm entries and compare throughput vs. `dsv4-fp8-mi355x-sglang` at matching concurrencies.

### DIFF
--- a/.github/configs/amd-master.yaml
+++ b/.github/configs/amd-master.yaml
@@ -1617,6 +1617,11 @@ dsv4-fp4-mi355x-sglang:
 # at runtime by benchmarks/single_node/dsv4_fp8_mi355x_vllm.sh at a
 # pinned SHA. Once both PRs merge into a release, switch to a vLLM ROCm
 # MI355X image and remove the build step.
+#
+# Serving flags follow vllm-project/recipes#433: AITER+AITER_LINEAR,
+# mp executor, triton_unfused MoE, async scheduling, max-num-seqs=128,
+# max-num-batched-tokens=8192, gpu-mem-util=0.6. Sweep matches the
+# sister sglang config (conc 4-64) so vLLM↔SGLang are comparable.
 dsv4-fp8-mi355x-vllm:
   image: rocm/atom:rocm7.2.2_ubuntu24.04_py3.12_pytorch_release_2.10.0_atom0.1.2.post
   model: deepseek-ai/DeepSeek-V4-Pro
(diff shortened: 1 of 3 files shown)


functionstackx:
i thought inferencex only does the most commonly served fp4 deepseekv4 pro tho

PR #1381, opened 2026-05-15
bench: fix v5 tokenizer fix when --model is an HF Hub repo id

trevor-m:
With transformers 5.6.0, an [AutoTokenizer dispatch change ](https://github.com/huggingface/transformers/commit/cd5bcad54032e902431c4e35411ff416f3b291f2) now returns a broken LlamaTokenizer for ByteLevel-BPE models which causes the tokens to be multiplied by ~4x resulting in an apparent throughput regression https://github.com/sgl-project/sglang/issues/25315

https://github.com/SemiAnalysisAI/InferenceX/pull/937 has a patch to fix this, however it wasn't applying if a HF model card ID was passed to `--model`; It only would apply when local directories were passed. This PR applies it for both cases, and also removed the exception hiding to hopefully make future issues more apparent.

### DIFF
--- a/utils/bench_serving/backend_request_func.py
+++ b/utils/bench_serving/backend_request_func.py
@@ -466,46 +466,64 @@ def _fix_tokenizer_for_sglang(tokenizer, model_path):
     import json
     from pathlib import Path
 
+    def _resolve(filename):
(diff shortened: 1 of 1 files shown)


cquil11 (approved):
seems reasonable. thx for the fix

PR #1422, opened 2026-05-17
Update qwen3.5-bf16-b300-sglang and -mtp SGLang image to v0.5.12-cu130

Klaud-Cold:
Updates SGLang image for `qwen3.5-bf16-b300-sglang` and `qwen3.5-bf16-b300-sglang-mtp` from v0.5.11-cu130 to v0.5.12-cu130.
\nRef #1154

Generated with [Claude Code](https://claude.ai/code)


functionstackx:
### Upstream issue filed

Filed [sgl-project/sglang#25564](https://github.com/sgl-project/sglang/issues/25564) for the B300 (sm_103) flash-attn assertion that's blocking this PR's full sweep. Includes the full traceback (vision encoder → `flash_attn_v4` → `flash_fwd_sm100.py:162`), the failing CI run, and root-cause analysis.

### Why it fails despite sm_103 nominally being in range

The assertion is `self.arch >= Arch.sm_100 and self.arch <= Arch.sm_110f`. B300 is `sm_103` (Blackwell Ultra, compute cap 10.3), numerically inside that range — but the cutedsl `Arch` enum's `__le__` resolves `sm_110f` to a value *less than* `sm_103` on non-cu13 cutedsl, collapsing the effective range to `sm100 <= ... <= sm101`. (Quoted explanation from the upstream FA PR author.)

### Upstream flash-attn fix already exists

[Dao-AILab/flash-attention#2572](https://github.com/Dao-AILab/flash-attention/pull/2572) by `@ocss884` (opened earlier today, 2-line diff, awaiting `@tridao` review) replaces the broken int-comparison with `is_family_of()`:

\`\`\`diff
-assert self.arch >= Arch.sm_100 and self.arch <= Arch.sm_110f, "Only SM 10.x and 11.x are supported"
+assert self.arch.is_family_of(Arch.sm_100f) or self.arch.is_family_of(Arch.sm_110f), \
+    "Only SM 10.x and 11.x are supported"
\`\`\`

Once that merges and Tri Dao tags a new flash-attn-4 release, sglang needs to bump the `flash-attn-4>=4.0.0b9` floor in `python/pyproject.toml` to the fixed version. Until then this PR has to either pin to `lmsysorg/sglang:v0.5.11-cu130` or stay paused.

functionstackx:
### Trying suggested workaround: `--mm-attention-backend triton_attn`

Per suggestion from radixark maintainer — added `--mm-attention-backend triton_attn` to both `qwen3.5_bf16_b300.sh` and `qwen3.5_bf16_b300_mtp.sh`. The text decoder still uses `--attention-backend trtllm_mha`; this only swaps the **multi-modal (vision encoder)** attention path, which is exactly where the `flash_attn` cute kernel was failing on B300 (`sm_103`) — see the traceback `qwen3_vl.py:838` → `vision.py:491` → `flash_attn_v4` → `flash_fwd_sm100.py:162` assertion.

Pushed as `9ae53b45`. Re-triggers the full sweep — if it goes green, this is the durable fix until [Dao-AILab/flash-attention#2572](https://github.com/Dao-AILab/flash-attention/pull/2572) merges and sglang bumps its pinned `flash-attn-4`. If it still fails, the fallback is to pin the image back to `lmsysorg/sglang:v0.5.11-cu130`.

functionstackx:
### Upstream commentary: the fix is actually a `nvidia-cutlass-dsl[cu13]` dependency, not the FA PR

[@mmangkad on sgl-project/sglang#25564](https://github.com/sgl-project/sglang/issues/25564#issuecomment-4474953385) chimed in and pushed back on the framing — the FA PR is a red herring. The real bug is on the sglang side, in the CuteDSL dependency. Quoting:

> I don't think the FA PR is technically needed here, even upstream. If SM103 is hitting this assertion, that basically means we are on the non-cu13 CuTeDSL aliasing path. SGLang already defaults to CUDA 13, so I think the CuTeDSL dependency should match that now by using `nvidia-cutlass-dsl[cu13]`.
>
> The confusing bit is from the CuTeDSL 4.5 change. The `Arch` enum now aliases `sm_101*` and `sm_110*` differently depending on CUDA version. On the non-cu13 path, `sm_110f` can effectively behave like `sm_101f`, so the FA4 check `self.arch >= Arch.sm_100 and self.arch <= Arch.sm_110f` can accidentally reject `sm_103`. So the fix is just to install the `[cu13]` extra for `nvidia-cutlass-dsl`. FI does the same thing for its own `cu13` extra. We should add `[cu13]` to `nvidia-cutlass-dsl` dependency for the same reason.

So once sglang bumps its `nvidia-cutlass-dsl` dependency to the `[cu13]` extra and reships the `lmsysorg/sglang:v0.5.12-cu130` image (or v0.5.13), this PR can drop the `--mm-attention-backend triton_attn` workaround entirely. Until then the workaround on `0fd4f60c` keeps the sweep alive.

PR #1447, opened 2026-05-17
[Klaud Cold] Update glm5-fp8-b200-sglang (+mtp) SGLang image to v0.5.12-cu130

functionstackx:
## Summary
- Bumps `glm5-fp8-b200-sglang` and `glm5-fp8-b200-sglang-mtp` from `lmsysorg/sglang:nightly-dev-cu13-20260317-1eea7448` (33/29 days old) to `lmsysorg/sglang:v0.5.12-cu130` (matches other b200 cu130 sglang recipes on main).

## Test plan
- [ ] Full sweep passes with `full-sweep-enabled` label.


functionstackx:
### Trying the cutlass DeepGemm workaround

This PR hit the same DeepGemm crash at `runtime_utils.hpp:143` (`CUDA_ERROR_ILLEGAL_ADDRESS (700)` on CUDA-graph capture) that KLAUD_DEBUG.md §4a previously documented for B300 — confirmed to bite **B200 (sm_100)** too at high concurrency. So far only `spec-none conc=128` failed; the `spec-mtp` variant has been passing because the smaller draft-model graph captures don't hit the same path.

Per §4a workaround #1, added `--fp8-gemm-runner-backend cutlass` to both `glm5_fp8_b200.sh` and `glm5_fp8_b200_mtp.sh` to bypass DeepGemm and route through CUTLASS instead. Pushed as `739585b0`.

If cutlass turns out to be slower than DeepGemm by enough to matter, the alternative is `export SGL_ENABLE_JIT_DEEPGEMM=0` (same effect — skips DeepGemm at runtime — but the JIT path costs ~25s of compile time per cold start, so it's a regression in startup latency, not steady-state throughput).

functionstackx:
### Fixed: wrong flag name in prior cutlass workaround

The earlier `739585b0` push used `--fp8-gemm-runner-backend cutlass`, but sglang v0.5.12 rejects that as unrecognized:

```
sglang serve: error: unrecognized arguments: --fp8-gemm-runner-backend cutlass
```

The actual flag (per the help-dump that fell out when sglang refused to start) is **`--fp8-gemm-backend`** (no "runner" in the middle):

```
[--fp8-gemm-backend {auto,deep_gemm,flashinfer_trtllm,flashinfer_cutlass,flashinfer_deepgemm,cutlass,triton,aiter}]
```

So all 22 failures on the last sweep were "sglang exited with help text", not the DeepGemm crash that the workaround was meant to bypass.

Pushed `d40fd278` swapping the flag name to `--fp8-gemm-backend cutlass` in both `glm5_fp8_b200.sh` and `_mtp.sh`. New sweep should now actually route through CUTLASS and dodge the DeepGemm illegal-memory-access at `runtime_utils.hpp:143`.

Also need a follow-up PR to fix the wrong flag name in `KLAUD_DEBUG.md §4a` (workaround #1) — it documents the same bad flag. Will track separately.

Issue #1511, opened 2026-05-18
[AI Generated] [Handoff] out of 70+ image updates, 13 stuck Klaud Cold PRs need upstream coordination / scope decisions


functionstackx:
# human

handing off to @Oseltamivir to try to debug this, my /loop failed on this and didnt get the chance to manually look at it

# below is [AI Generated]

# Handoff to @Oseltamivir — 13 stuck Klaud Cold PRs

These 13 PRs all have a real diagnosis and (where applicable) an applied workaround, but they're blocked on upstream fixes, infra outages, or judgment calls outside what `/loop` should keep retrying. Handing them off so they don't keep churning sweep capacity.

PRs grouped by category:

| Category | Count | PRs |
| --- | --- | --- |
| DSV4 — needs custom image | 4 | ~~#1461~~, #1460, #1455, #1450 |
| Upstream bugs with ticket filed | 4 | #1494, #1441 (AMD → sgl#25742); ~~#1451 → sgl#25863~~; #1420 → sgl#25563 |
| MI300 cluster down (firmware upgrades) | 3 | #1403, #1499, #1482 |
| Other | 2 | ~~#1512, #1521~~ done |

And the same set grouped by vendor (NVIDIA vs AMD):

| Vendor | Category | Count | PRs |
| --- | --- | --- | --- |
| **NVIDIA** (H200/B200/B300) — 7 total | DSV4 — latest stable image didnt work | 4 | #1461 (H200), #1460 (H200), #1455 (B300), #1450 (B200) |
|  | Upstream bugs with ticket filed | 2 | #1451 (B300 → sgl#25863), #1420 (B300 → sgl#25563) |
|  | Other | 1 | #1512 (B300, sgl-deep-gemm pin test for sgl#25551) |
| **AMD** (MI300X/MI355X) — 6 total | Upstream bugs with ticket filed | 2 | #1494 (MI355X), #1441 (MI355X) — both AMD-acknowledged via sgl#25742 |
|  | MI300 cluster down (firmware upgrades) | 3 | #1403 (MI300X), #1499 (MI300X), #1482 (MI300X) |
|  | Other | 1 | #1521 (MI355X, eval-only flake) |

For each PR below: links to the PR + most recent failing sweep run, what's wrong, what's been tried, and the upstream tickets (if filed).

---

## DSV4 — latest stable image didnt work (4 PRs)

[...]

PR #1420, opened 2026-05-17
[Handoff to @Oseltamivir Claude /loop] Update glm5-fp4-b300-sglang and -mtp SGLang image to v0.5.12-cu130

Klaud-Cold:
Updates SGLang image for `glm5-fp4-b300-sglang` and `glm5-fp4-b300-sglang-mtp` from v0.5.11-cu130 to v0.5.12-cu130.
\nRef #1154

Generated with [Claude Code](https://claude.ai/code)

---

> [!NOTE]
> **Medium Risk**
> Changes inference container versions and overrides a transitive dependency in benchmark paths, which can affect measured performance and stability on B300 MTP runs without touching app auth or data paths.
> 
> **Overview**
> Bumps the **GLM-5 FP4 B300** SGLang benchmark configs (`glm5-fp4-b300-sglang` and `glm5-fp4-b300-sglang-mtp`) from **`lmsysorg/sglang:v0.5.11-cu130`** to **`v0.5.12-cu130`** in `nvidia-master.yaml`, with a matching **perf-changelog** entry.
> 
> The single-node benchmark scripts **`glm5_fp4_b300.sh`** and **`glm5_fp4_b300_mtp.sh`** now **pin `flashinfer_python` / `flashinfer_cubin` to 0.6.8.post1** (the versions bundled with v0.5.11) after the image upgrade, to **isolate a suspected trtllm batched-GEMM regression** (crash at bs=128 with EAGLE on B300) linked to the flashinfer bump in v0.5.12 (see sglang#25563). The MTP script still installs its existing transformers/hub pins; only flashinfer is added on top.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ab94800e941a9f1f86edd370602fc915fb2fb7fa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1512, opened 2026-05-18
[Handoff to @Oseltamivir Claude /loop] [Klaud Cold] Test sgl-deep-gemm==0.0.1 pin for sgl#25551 (glm5-fp8-b300 DeepGemm regression)

functionstackx:
## Summary

Testing @trevor-m's suggestion in [sgl-project/sglang#25551 (comment)](https://github.com/sgl-project/sglang/issues/25551#issuecomment-4481466979): inside the `lmsysorg/sglang:v0.5.12-cu130` container, downgrade `sgl-deep-gemm` from `0.1.0` → `0.0.1` and re-enable JIT DeepGemm to see whether the deep-gemm version jump is what triggers the B300 `CUDA_ERROR_ILLEGAL_ADDRESS` TMA-descriptor regression.

Previously, both `glm5_fp8_b300.sh` and `glm5_fp8_b300_mtp.sh` worked around the bug by setting `SGL_ENABLE_JIT_DEEPGEMM=0`, which avoids the broken kernel path entirely. This PR replaces that workaround with a targeted version pin so we can attribute the regression.

## Outcome interpretation
- **Sweep green** → the deep-gemm 0.0.1 → 0.1.0 upgrade is the source of the regression; we should file an upstream issue against sgl-deep-gemm with a standalone repro.
- **Sweep red** → the regression is elsewhere in sglang's v0.5.12 changes (not in deep-gemm), and bisecting per Trevor's second experiment is the next step.

## Not for merge
This PR is **for upstream debugging only** — once the result is known, either revert (if 0.0.1 doesn't fix it) or convert into a real fix (if it does). Do not merge as-is; we don't want to silently pin an older deep-gemm in production.

## Test plan
- [ ] Sweep runs to completion on both `glm5-fp8-b300-sglang` and `glm5-fp8-b300-sglang-mtp`
- [ ] Capture result on sgl-project/sglang#25551
- [ ] Decide next step based on green/red

🤖 AI-generated via Claude Code `/loop`.

### DIFF
--- a/benchmarks/single_node/glm5_fp8_b300.sh
+++ b/benchmarks/single_node/glm5_fp8_b300.sh
@@ -23,14 +23,16 @@ nvidia-smi
 
 if [[ "$MODEL" != /* ]]; then hf download "$MODEL"; fi
 
-pip install --no-deps "transformers==5.2.0" "huggingface-hub==1.4.1"
+pip install --break-system-packages --no-deps "transformers==5.2.0" "huggingface-hub==1.4.1"
 
 # Testing @trevor-m's suggestion in sgl-project/sglang#25551 (comment 4481466979):
 # downgrade sgl-deep-gemm 0.1.0 → 0.0.1 inside the v0.5.12 container to check
 # whether the deepgemm version jump is what causes the B300 TMA-descriptor
 # CUDA_ERROR_ILLEGAL_ADDRESS regression. Re-enabling JIT DeepGemm so the
 # downgraded version actually runs.
-pip install --no-deps "sgl-deep-gemm==0.0.1"
+# --break-system-packages required: the container's Python is PEP-668 externally-managed,
+# so the previous attempt silently failed and left the bundled 0.1.0 in place.
+pip install --break-system-packages --no-deps "sgl-deep-gemm==0.0.1"
 export SGL_ENABLE_JIT_DEEPGEMM=1
 
 SERVER_LOG=/workspace/server.log
--- a/benchmarks/single_node/glm5_fp8_b300.sh
+++ b/benchmarks/single_node/glm5_fp8_b300.sh
@@ -25,11 +25,13 @@ if [[ "$MODEL" != /* ]]; then hf download "$MODEL"; fi
 
 pip install --no-deps "transformers==5.2.0" "huggingface-hub==1.4.1"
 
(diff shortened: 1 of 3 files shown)


functionstackx:
Handing off to @Oseltamivir — added as #9 in the cluster-fix handoff issue #1511. Both this PR and #1420 are blocked on the same b300 nodelist issue (`runners/launch_b300-nv.sh:336`).

_AI-generated via Claude Code /loop._

functionstackx:
**Sweep result invalid — pip install blocked, pin never applied.**

[Failing job](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/26144077664/job/76895488937?pr=1512) shows both `pip install` lines in `glm5_fp8_b300.sh` hit Debian's PEP-668 externally-managed-environment error inside `lmsysorg/sglang:v0.5.12-cu130`:

```
error: externally-managed-environment
× This environment is externally managed
```

So `sgl-deep-gemm==0.0.1` was never installed; the sweep ran with whatever `sgl-deep-gemm` ships in the container (presumably 0.1.0), and hit the **same** `CUDA_ERROR_ILLEGAL_ADDRESS` at `runtime_utils.hpp:143` during CUDA-graph capture — which is consistent with the bug we were already seeing.

**Conclusion:** This run does **not** answer @trevor-m's question. We need to fix the pip command first, then re-run.

**Proposed fix** in `glm5_fp8_b300{,_mtp}.sh`:

```diff
- pip install --no-deps "transformers==5.2.0" "huggingface-hub==1.4.1"
+ pip install --break-system-packages --no-deps "transformers==5.2.0" "huggingface-hub==1.4.1"
…
- pip install --no-deps "sgl-deep-gemm==0.0.1"
+ pip install --break-system-packages --no-deps "sgl-deep-gemm==0.0.1"
```

After that change, the sweep will either:
- **Green** → confirms deep-gemm 0.0.1→0.1.0 is the regression (as Trevor hypothesized); file upstream against `sgl-deep-gemm`.
- **Red** → regression is elsewhere in `v0.5.12`; move to Trevor's bisect experiment.

Reported back to https://github.com/sgl-project/sglang/issues/25551#issuecomment-… so the invalid result doesn't mislead upstream. Re-adding to handoff [#1511](https://github.com/SemiAnalysisAI/InferenceX/issues/1511) until the rerun lands.

Oseltamivir:
Screenshot 2026-05-20 at 1 23 07 PM

PR #1551, opened 2026-05-22
feat(power): aggregate measured GPU power into agg result JSON

arygupt:
## Summary

Adds measured per-GPU power and joules-per-output-token to every benchmark's `agg_<run>.json`, sourced from the existing `gpu_metrics.csv` that `start_gpu_monitor` already produces. The InferenceX-app dashboard consumes these via a companion PR ([semianalysisai/InferenceX-app](https://github.com/semianalysisai/InferenceX-app)) to render new chart options alongside the existing TDP-derived `jTotal`/`jOutput`/`jInput`.

Two new fields land in the agg JSON:

- `avg_power_w` — mean per-GPU draw during the load window
- `joules_per_output_token` — `avg_power_w * num_gpus * duration / total_output_tokens`

## How it works

1. **`benchmark_serving.py`** now records `benchmark_start_time_unix` and `benchmark_end_time_unix` (wall-clock epoch) alongside the existing `duration`. The aggregator needs these to know which slice of the long-running monitor CSV is the actual load window — without them, naive averaging would mix in ~60s of server warmup (~120W) and the optional eval phase (~300W), biasing a 720W per-GPU draw down to roughly 440W.

2. **`utils/aggregate_power.py`** (new, stdlib only, ~210 lines) reads the CSV, detects vendor schema by header regex (handles nvidia-smi `power.draw [W]` and amd-smi `socket_power`), filters samples to the bench window, averages per-GPU power per timestamp then over time, and atomically patches the agg JSON. Best-effort throughout — missing/empty/malformed CSV is logged to stderr and skipped without ever failing the run.

3. **`utils/process_result.py`** calls the aggregator right after writing the agg JSON. Path resolution checks `$GPU_METRICS_CSV` → `./gpu_metrics.csv` → `/workspace/gpu_metrics.csv`, accommodating the scripts in `benchmarks/single_node/` that override the default path. Wrapped in `try/except` so telemetry never blocks the upload.

4. **`benchmarks/benchmark_lib.sh`** `export`s `GPU_METRICS_CSV` so per-script CSV-path overrides cross the shell→Python boundary.

No workflow YAML change, no schema migration anywhere downstream — the InferenceX-app ETL's `benchmark-mapper.ts` is permissive about numeric keys in the agg JSON.

## Verification

End-to-end smoke test on a synthesized 1680-row CSV (8 GPUs × 210s spanning warmup at 120W, bench at 720W, eval at 300W):

```
[aggregate_power] avg_power_w=715.69 (per GPU, n=8)
                  joules_per_output_token=8.3869
                  duration=120.0s output_tokens=81920
```

Cross-check: 720W × 8 GPUs / 682 tok/s ≈ 8.45 J/tok. ✓ Window isolation correctly excluded the warmup + eval samples (naive average would have given ~440W).

## Test plan

- [x] 26 unit tests covering NVIDIA + AMD CSV formats, multi-GPU per-sample aggregation, window filtering, malformed-row resilience, missing files, atomic JSON patching, divide-by-zero on failed runs
- [x] 3 subprocess integration tests through `process_result.py`: stages a CSV + bench JSON + env vars, asserts `agg_<run>.json` gets patched
- [x] All 22 existing `test_process_result.py` tests still pass (no regressions in the established flow)
- [ ] First real benchmark run after merge — verify the two new keys appear in the uploaded `agg_<run>.json` artifact
- [ ] InferenceX-app ingest picks up the new keys (no METRIC_KEYS warning in ETL logs)

## Backfill option

The workflow already uploads `gpu_metrics.csv` as an artifact for every run (`.github/workflows/benchmark-tmpl.yml`). After this merges, historical runs that have both a CSV and an `agg_<run>.json` could be backfilled by re-running this aggregator against the artifact store. Out of scope here.

### DIFF
--- a/benchmarks/benchmark_lib.sh
+++ b/benchmarks/benchmark_lib.sh
@@ -15,6 +15,7 @@ mkdir -p "$PYTHONPYCACHEPREFIX" 2>/dev/null || true
 
 GPU_MONITOR_PID=""
 GPU_METRICS_CSV="/workspace/gpu_metrics.csv"
+export GPU_METRICS_CSV
 
 # Start background GPU monitoring that logs metrics every second to CSV.
 # Auto-detects NVIDIA (nvidia-smi) or AMD (amd-smi) GPUs.
@@ -32,6 +33,7 @@ start_gpu_monitor() {
     done
 
     GPU_METRICS_CSV="$output"
+    export GPU_METRICS_CSV
 
     if command -v nvidia-smi &>/dev/null; then
         nvidia-smi --query-gpu=timestamp,index,power.draw,temperature.gpu,clocks.current.sm,clocks.current.memory,utilization.gpu,utilization.memory \
--- a/perf-changelog.yaml
+++ b/perf-changelog.yaml
@@ -3083,3 +3083,9 @@
   description:
     - "Update SGLang image from v0.5.10.post1-cu130 / v0.5.11-cu130 (30d old) to v0.5.12-cu130"
   pr-link: https://github.com/SemiAnalysisAI/InferenceX/pull/1451
+
+- config-keys:
+    - qwen3.5-fp8-h200-sglang
+  description:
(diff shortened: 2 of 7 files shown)


arygupt:
Closed this in favor of #1558

PR #1585, opened 2026-05-29
[Fix] Remove MoRI-IO patches from vLLM Disagg benchmarks 

simondanielsson:
These patches were upstreamed in https://github.com/vllm-project/vllm/pull/40344 so we can use the official release instead.

Run: https://github.com/SemiAnalysisAI/InferenceX/actions/runs/28870448283

Results from this run are very similar to the existing Kimi vllm disagg results, as expected

---

> [!NOTE]
> **Medium Risk**
> Changes affect multi-node prefill/decode KV transfer and MoE all2all for production benchmark paths; behavior now depends on a nightly image rather than battle-tested in-tree patches, though validation noted similar results to prior runs.
> 
> **Overview**
> **MI355X vLLM disaggregated** Kimi K2.5 and MiniMax M2.5 configs now use a **pinned ROCm nightly** image that includes upstream MoRI-IO fixes (vllm#40344), with a TODO to move to a stable release later. Scenario `additional-settings` no longer pass **`VLLM_MORIIO_CONNECTOR_READ_MODE`**.
> 
> **`setup_deps.sh`** drops the large set of **runtime vLLM/MoRI-IO and scheduler patches** (save/load KV timeouts, transfer timeouts, read-mode scheduler fixes, idle KV reaper); vLLM-disagg startup is now recipe deps, amd-quark, and UCX/RIXL paths only.
> 
> **Serving wiring** changes for the new stack: **`--all2all-backend mori` → `mori_low_latency`** in `models_vllm.yaml`; **`read_mode: true`** in MoRIIO `kv_connector_extra_config` in `server_vllm.sh` instead of the env var; **`job.slurm`** stops injecting that env and bumps the default **vllm-router** image. **`submit.sh`** no longer exports the read-mode env. **`perf-changelog.yaml`** documents the config-key updates.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 556d92608dc1d3c3e1dddaad6350ad3f005ffa99. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1584, opened 2026-05-29
[AMD] improve dsr1 fp4 disagg

billishyahao:
Improve dsr1 fp4 disagg 

> Co-authored-by: billishyahao <bill.he@amd.com>
> Co-authored-by: Duyi-Wang <duyi.wang@amd.com>

---

> [!NOTE]
> **Low Risk**
> Benchmark matrix and environment-variable changes only; no application auth or production serving logic.
> 
> **Overview**
> This PR refines **DeepSeek-R1 FP4 MI355X SGLang disaggregated MTP** coverage and the shared **MoRI/SGLang** runtime env used by those jobs.
> 
> In `.github/configs/amd-master.yaml`, the former combined `dsr1-fp4-mi355x-sglang-disagg-mtp` entry is **renamed/split**: **`dsr1-fp4-mi355x-sglang-disagg-1k1k-mtp`** keeps the 1024/1024 scenario, and a new **`dsr1-fp4-mi355x-sglang-disagg-8k1k-mtp`** holds only the 8192/1024 MTP search space. Both use **`lmsysorg/sglang-rocm:v0.5.12.post1-rocm720-mi35x-20260529`** (image bump from the prior May 19 tag).
> 
> The **8k/1k MTP** matrix is retuned: several **`conc-list`** values shrink or shift (e.g. mid-tier 1P2D points **64/128/256 → 32/64**; some DEP8 high-conc lists change), many decode rows move **`DECODE_MTP_SIZE` to 3**, and **new DEP8 sweep points** at conc **128** and **64** are added.
> 
> `benchmarks/multi_node/amd_utils/env.sh` sets **`AITER_LOG_LEVEL=ERROR`** and replaces **`SGLANG_MORI_FP8_COMB`** with **`MORI_COMBINE_DTYPE_PREFILL=fp8_direct_cast`** and **`MORI_COMBINE_DTYPE_DECODE=fp8`**. `perf-changelog.yaml` documents the new **`dsr1-fp4-mi355x-sglang-disagg-8k1k-mtp`** key (image bump + conc 128/256).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cbc716887bdff3369d82a38aa9cf4fbc35e5d2e1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1581, opened 2026-05-28
[AMD][MI355X] add the kimik2.5_int4_mi355x_vllm-disagg support for AMD GPU.

haic0:
## Summary
- Add Kimi-K2.5 INT4 to the MI355X vLLM disaggregated multi-node model registry.
- Add a matching MI355X vLLM-disagg multi-node config and launcher following the Kimi-K2.5 INT4 topology.
-Add Kimi-K2.5-INT4 disagg inference configs (1P2D)
-Consolidate amd_utils to support both sglang and vllm disagg engines

## Test plan
- Parsed edited YAML files with python3/yaml.
- Generated the new multi-node matrix config with generate_sweep_configs.py.
-Verify CI passes on amd-master config with vllm-disagg entries
 -Validate multi-node sglang benchmarks still work (no regressions from amd_utils refactor)
 -Run vllm-disagg multi-node benchmark on MI355X cluster
 -Confirm Kimi-K2.5 INT4 disagg recipes launch correctly

---

> [!NOTE]
> **Low Risk**
> Benchmark and CI matrix wiring only; no runtime application or auth paths change.
> 
> **Overview**
> Adds **Kimi-K2.5 INT4** disaggregated prefill–decode benchmarking on **MI355X** with **vLLM** (`vllm/vllm-openai-rocm:v0.21.0`), aligned with the existing Kimi MXFP4 vLLM-disagg **1P+2D** layout.
> 
> Registers `kimik2.5-int4-mi355x-vllm-disagg` in `amd-master.yaml` for **1k1k** and **8k1k** (conc **8–512**): one prefill worker (TP8, EP1) plus two decode workers (TP8, EP8), with `PREFILL_NODES`/`DECODE_NODES` and `VLLM_MORIIO_CONNECTOR_READ_MODE=1`.
> 
> Adds **`Kimi-K2.5-INT4`** in `models_vllm.yaml` (prefill/decode CLI flags, MoRI on decode, INT4 quick-reduce / V2 model-runner env, `models--moonshotai--Kimi-K2.5` cache dir) and a new multinode launcher **`kimik2.5_int4_mi355x_vllm-disagg.sh`** that maps matrix EP/DP settings and submits via `amd_utils/submit.sh`. Documents the change in `perf-changelog.yaml`.
> 
[...]

PR #1590, opened 2026-05-29
fix(process_result): fail loudly on zero-throughput disagg runs (no more masked ZeroDivisionError)

arygupt:
## What

When a disagg benchmark fails server-side (warmup deadlock, MoRI/KV-transport failure, server never reaching ready), the harness still writes a result JSON — but with **zeroed throughput/latency**. `utils/process_result.py` then computes the tpot→interactivity reciprocal `1000.0 / float(tpot)` with `tpot == 0`, raising:

```
ZeroDivisionError: float division by zero  (utils/process_result.py:132)
```

This opaque error is the dominant red signature on `mi355x-disagg` sweeps (e.g. #1584's DEP8/MTP jobs). It **masks the real cause** and makes every server-side failure look identical, so triage always requires hand-pulling slurm logs.

## Fix

Guard up front, right after the result JSON is loaded: if `output_throughput` or `total_token_throughput` is 0, exit with a clear, actionable message pointing at the `multinode_server_logs` artifact — instead of letting the reciprocal raise.

A legitimately successful run always has positive throughput, so this never triggers on healthy results.

## Why this matters

This doesn't *fix* the underlying MoRI warmup deadlock (that's upstream / #1585), but it converts "mystery ZeroDivisionError" into the real story on every future disagg run — the prerequisite for actually debugging AMD multinode. It would, for example, have immediately distinguished #1584 (real disagg failure) from #1581 (deterministic model-not-found, fails before any result is written).

Vendor-agnostic; affects both NVIDIA and AMD disagg result processing.

---

> [!NOTE]
> **Low Risk**
> Small guard on result ingestion only; successful runs with positive throughput are unchanged.
> 
> **Overview**
> Adds an early check in `utils/process_result.py` right after the benchmark result JSON is loaded: if **`output_throughput`** or **`total_token_throughput`** is zero or missing, the script **`SystemExit`s** with an explicit failure message (including the metric values and a pointer to **`multinode_server_logs`**) instead of continuing into downstream math.
> 
> That prevents the later **`1000.0 / tpot`** interactivity conversion from raising a **`ZeroDivisionError`** when failed/disagg runs still write a result file with zeroed throughput and latency—so server-side failures (warmup deadlock, transport issues, server not ready) surface as one actionable error rather than an identical division-by-zero on every bad run.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bd405c142bc9d1c7d8eb4877e8a82ff35bc60852. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/utils/process_result.py
+++ b/utils/process_result.py
@@ -42,6 +42,23 @@ image = base_env['IMAGE']
 with open(f'{result_filename}.json') as f:
     bmk_result = json.load(f)
 
+# A failed/degenerate benchmark (server never came up, disagg warmup deadlock,
+# MoRI/KV-transport failure, etc.) still writes a result JSON, but with zeroed
+# throughput and latency metrics. Detect that here and fail with the real,
+# actionable reason. Otherwise the tpot reciprocal below (1000.0 / tpot) raises
+# an opaque "ZeroDivisionError: float division by zero" that masks the true
+# server-side cause and makes every such failure look identical.
+_output_tput = float(bmk_result.get('output_throughput', 0) or 0)
+_total_tput = float(bmk_result.get('total_token_throughput', 0) or 0)
+if _output_tput <= 0 or _total_tput <= 0:
+    raise SystemExit(
+        "FAIL: benchmark produced no decode throughput "
+        f"(output_throughput={_output_tput}, total_token_throughput={_total_tput}) "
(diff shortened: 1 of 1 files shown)


functionstackx:
stale PR

PR #1591, opened 2026-05-29
ci(disagg): fail before writing result file + surface real failure class

arygupt:
## What

Two independent, verified disagg-CI hardening fixes that make AMD multinode failures **fail loudly and legibly** instead of silently producing misleading artifacts.

### A — `benchmark_serving.py`: gate before write
The request-failure-rate gate ran **after** the result JSON was written, so a zero-/sub-threshold run left a schema-valid JSON on disk and *then* exited non-zero. Downstream collectors (`launch_mi355x-amds.sh`, `benchmark-multinode-tmpl.yml`) key on file **existence**, not exit code — so a written-then-failed file looked successful. Moved the gate above `if args.save_result:` so disk state stays consistent with the exit code. Complements #1590 (that guard is downstream in `process_result`; this stops the misleading file at the source).

### B — `launch_mi355x-amds.sh`: emit `::error::` with the real failure class
`cleanup_and_save_logs` tailed the slurm `.err` but emitted no GitHub annotation, so the Actions UI showed only the generic "No benchmark result files found". Now it classifies the `.err` and emits one `::error::` naming the class — **recipe model-not-found** (#1581) / readiness-timeout / **fp8_blockwise IntraNode combine** (#1584) / MoRI transport / segfault — deterministic recipe errors checked first so a config bug is never mislabeled as a flake. Guarded on `GITHUB_ACTIONS`.

## Held for follow-up
Bounded flake-only retry + node-quarantine in `benchmark-multinode-tmpl.yml`. Needs the `submit.sh` node-allocation path traced to thread a `--exclude`, plus offline classifier validation against canned `.err` fixtures — not shipping a half-verified control-flow change to production CI.

## Test
- A: `python3 -m py_compile` passes; gate logic unchanged, only reordered (no re-indent). Unit-testable by forcing `completed=0` / `completed<95%` and asserting no JSON is written + exit≠0.
- B: `bash -n` passes; pure observability inside the existing EXIT trap, runs only under `GITHUB_ACTIONS`.

---

> [!NOTE]
> **Low Risk**
> CI-only ordering and observability changes; benchmark gate logic is unchanged, and launcher edits only add stderr pattern matching for Actions annotations.
> 
> **Overview**
> Hardens **AMD disaggregated multinode CI** so failed runs are easier to diagnose and no longer look successful downstream.
> 
> In **`benchmark_serving.py`**, the **5% request failure-rate check** now runs **before** any result JSON is written. Collectors treat a result file on disk as success even when the process exits non-zero; moving the gate avoids leaving a valid JSON artifact on sub-threshold or zero-completion runs.
> 
> In **`launch_mi355x-amds.sh`**, **`cleanup_and_save_logs`** still tails the Slurm **`.err`** log but now emits a GitHub **`::error::`** annotation (when `GITHUB_ACTIONS` is set) that classifies the failure—recipe/model-not-found and MoRI **fp8_blockwise** config issues are matched **before** transport-flake patterns so deterministic misconfigurations are not labeled as flakes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 992b90f93782403bab81cab66e071cab4be4a741. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/runners/launch_mi355x-amds.sh
+++ b/runners/launch_mi355x-amds.sh
@@ -67,6 +67,31 @@ if [[ "$IS_MULTINODE" == "true" ]]; then
             echo "=== Slurm job stderr ==="
             tail -100 "$err_file"
             echo "========================"
+            # Surface the real failure class in the Actions UI. Without this, a
+            # launch failure shows only the generic "No benchmark result files
+            # found" from benchmark-multinode-tmpl.yml. Order matters: check the
+            # deterministic recipe error (model-not-found, #1581) before the
+            # transport-flake patterns (#1584 MoRI/readiness) so a config bug is
+            # never mislabeled as a flake.
+            if [[ -n "${GITHUB_ACTIONS:-}" ]]; then
+                local sig=""
+                if   grep -qiE "Model '.*' not found|FATAL: Model|model .* not found" "$err_file"; then
+                    sig="recipe-error: model not found (deterministic - check MODEL/MODEL_PATH, not MoRI)"
+                elif grep -qiE "ReadTimeout|readiness.*timeout|warmup.*time(d)? ?out|health.*timeout" "$err_file"; then
+                    sig="transport-flake: readiness/warmup timeout (MoRI pd-disagg)"
+                elif grep -qiE "Fp8BlockwiseQuant.*IntraNode|dispatch_combine|combine.*IntraNode" "$err_file"; then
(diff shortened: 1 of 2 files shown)


functionstackx:
stale PR

PR #1718, opened 2026-06-12
feat(ci): add full-sweep-fail-fast sweep label

cquil11:
Adds a new PR sweep label, `full-sweep-fail-fast`: full untrimmed sweep, no canary gate, with `strategy.fail-fast` enabled on every matrix. Existing labels (`sweep-enabled`, `full-sweep-enabled`, `non-canary-full-sweep-enabled`) are unchanged.

**How it works**
- Each sweep matrix's `strategy.fail-fast` flips on when the label is present: the first failed job cancels that matrix's remaining jobs (native GitHub behavior — failing job stays red, siblings show cancelled, run concludes failed).
- Fail-fast is matrix-scoped: other scenario matrices (1k1k / 8k1k / agentic / evals) keep running and self-terminate on their own first failure, so their completed results stay valid.
- Label wired into the same gating/conflict-rejection/concurrency/reuse paths as the existing sweep labels (`/reuse-sweep-run` eligible).

**When it's useful**
- Sweeps where one failure means the rest of that matrix is wasted GPU time (e.g. new image bring-up) — stop on the first failure instead of running every concurrency point to completion.

**Caveats**
- One flaky job cancels its matrix's in-flight siblings.
- Contains a temporary `perf-changelog.yaml` test entry (`gptoss-fp4-h100-vllm`) to exercise the label — revert before merge.

---

> [!NOTE]
> **Low Risk**
> CI workflow and maintainer-script changes only; no benchmark runtime or auth/data paths affected.
> 
> **Overview**
> Introduces a fourth PR sweep label, **`full-sweep-fail-fast`**: full untrimmed concurrency sweep with **no canary gate**, same as `non-canary-full-sweep-enabled`, but with **matrix-scoped `strategy.fail-fast`** on every sweep job in `run-sweep.yml` (single/multi-node 1k1k/8k1k, agentic, evals). The first failed job in a matrix cancels that matrix’s remaining jobs; other scenario matrices keep running independently.
> 
> The label is wired through the same paths as existing full-sweep labels: PR workflow gating, concurrency exceptions, mutual-exclusion checks in `setup`, and the unofficial visualizer comment job. **`/reuse-sweep-run`** and post-merge reuse now treat runs with this label as eligible full sweeps (`find_reusable_sweep_run.py`, `merge_with_reuse.sh`, workflow README).
> 
[...]

PR #1720, opened 2026-06-12
feat(ci): add dispatchable fail-fast-cancel workflow

cquil11:
Companion to #1718 (`full-sweep-fail-fast` label). Contains only `.github/workflows/fail-fast-cancel.yml`.

- `workflow_dispatch` can only target workflows registered on the default branch — #1718's test sweep dispatched this and got a 404, so the run was never cancelled. It needs to land on main before the label can work (including for testing #1718 itself).
- Inert by itself: triggers only on explicit dispatch from the benchmark templates' fail-fast path, then waits for the failing job to conclude (keeping its red failure status) before cancelling the target run.
- No sweep is triggered by merging this (doesn't touch `perf-changelog.yaml`).

---

> [!NOTE]
> **Low Risk**
> CI-only change with scoped `actions: write`; no application, auth, or data-path impact.
> 
> **Overview**
> Adds a new **`fail-fast-cancel`** GitHub Actions workflow that benchmark sweeps can dispatch when fail-fast mode is on.
> 
> Instead of cancelling the workflow run from inside a failing matrix job (which can rewrite that job’s conclusion from **failure** to **cancelled**), this runs as a separate workflow: it polls the target run until at least one job has concluded as **failure**, then calls the Actions API to cancel the rest of the run. It only runs on manual **`workflow_dispatch`** with a **`run-id`** input, uses **`actions: write`**, and still attempts cancellation if the poll times out (~10 minutes).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b9cdd223c2100f9d22b6eb4ce86912263c2b235d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/fail-fast-cancel.yml
+++ b/.github/workflows/fail-fast-cancel.yml
@@ -1,0 +1,45 @@
+name: "Fail-Fast Cancel Run"
+run-name: Fail-fast cancel for run ${{ inputs.run-id }}
+
+# Dispatched by the benchmark templates when a job fails with fail-fast
+# enabled. Cancelling a run from inside one of its own jobs flips the
+# failing job's conclusion from 'failure' to 'cancelled' (the job is still
+# in progress when the cancellation is processed), which hides which job
+# caused the cascade. This workflow runs outside the target run: it waits
+# for the failing job to actually conclude as 'failure' (keeping its red X)
+# before cancelling everything else.
+
+on:
+    workflow_dispatch:
+        inputs:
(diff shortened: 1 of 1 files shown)


cquil11:
No longer needed: #1718 dropped cross-matrix run cancellation in favor of native matrix-scoped strategy.fail-fast, so there is nothing to dispatch.

PR #5, opened 2025-09-05
Add new FW (TRT) and precision support

kedarpotdar-nv:
## Overview

This PR adds TensorRT-LLM (TRT-LLM) as a new inference framework for LLaMA 70B benchmarking on NVIDIA H200 and B200 GPUs, alongside the existing vLLM framework. This enables direct performance comparison between vLLM and TRT-LLM on the same hardware.

### Key Features
- Multi-framework support: vLLM and TRT-LLM for LLaMA 70B
- Precision support: FP8 (default) and FP4 (future-ready)
- Unified plotting: All frameworks and hardware on single performance plots
- Framework-specific Docker images: Prevents image conflicts between frameworks
- Clean job naming: Clear identification in GitHub Actions UI

### 🔧 Core Workflow Updates
.github/workflows/benchmark-tmpl.yml
- Added framework and precision as required inputs
- Updated RESULT_FILENAME to include framework and precision
- Modified result processing to pass framework and precision to process_result.py
- Fixed job naming to prevent duplication
- Added hardware extraction logic for proper result processing

.github/workflows/70b-tmpl.yml
- Added bmk-h200-trt and bmk-b200-trt jobs for TRT-LLM
- Configured TRT-LLM jobs with nvidia/tensorrt-llm Docker image
- Set precision to empty string for FP8 (default)
- Updated all jobs to include framework and precision inputs

#### 🚀 New Benchmark Scripts
benchmarks/70b_h200_trt_slurm.sh and benchmarks/70b_b200_trt_slurm.sh
- TRT-LLM server setup for H200 using mpirun trtllm-serve
- Inline llama-config.yml configuration
- Client benchmarking with benchmark_serving.py

### 🔄 Launcher Script Updates
Updated SLURM Launchers
runners/launch_h200-nv.sh
runners/launch_h200-cw.sh
runners/launch_h200-nb.sh
runners/launch_b200-nv.sh
Key improvements:
- Framework-specific SQSH file naming to prevent Docker image conflicts
- Dynamic script selection based on framework (VLLM/SGLang use base scripts, TRT uses _trt scripts)
- Proper MODEL_CODE environment variable passing to containers
- Framework reset logic for VLLM and SGLang to use default script names

### 📊 Result Processing & Visualization
utils/process_result.py
- Added framework and precision command-line arguments
- Updated output data structure to include framework and precision
- Default precision handling (empty string → 'fp8')

utils/plot_perf.py
- Added distinct colors for TRT-LLM results:
h200-trt: dark green
[...]

PR #3, opened 2025-09-02
[NVIDIA] update vllm b200 image. TODO: add logic for docker runner.

kedarpotdar-nv:
Updated vllm B200 runner image with ToT . This image will work for Hopper as well, but want to try B200 updates first.

TODO: apply changes to Hopper and B200 docker config.

PR #1806, opened 2026-06-16
Fix for https://github.com/sgl-project/sglang/issues/22072 / 修复 SGLang MoRI 分离式推理在 MI325X 上的挂起问题

davzhuAMD:
Fix SGLang MORI disagg tests hanging on MI325X. The test matrix appeared as follows:

Test # | Config | Decode nodes | EP/DP | MoRI a2a | Status
-- | -- | -- | -- | -- | --
1 | P(tp4) D(tp8) | 1 | No | No | Works
2 | P(tp8) D(tp8, 2 workers) | 1 each | No | No | Works
3 | P(tp4) D(tp8/ep8/dp, 1 node) | 1 | Yes | Yes | **Works (used to hang)**
4 | P(tp4) D(tp8/ep8/dp, 1 node, MTP) | 1 | Yes | Yes | **Works (used to hang)**
5 | P(tp8) D(tp16/ep16/dp16, 2 nodes) | 2 | Yes | Yes | **Works (used to hang)**

Investigation employed helper scripts from https://github.com/davzhuAMD/InferenceX/tree/sglang-pd-mi325x, derived from https://github.com/JohnQinAMD/InferenceX/tree/sglang-pd-mi300-mi325x, to aid investigation and repro. These are the run_* and start_* scripts that essentially act as config instructions to set up the repro.

To validate, copy the run_* and start_* scripts from the sglang-pd-mi325 branch mentioned above into the main InferenceX folder, update the PREFILL_MODEL_HOST_DIR, DECODE_MODEL_HOST_DIR, PREFILL_NODE, and DECODE_NODE IPs per test in these scripts, then run

- start_test_1.sh
- start_test_2.sh
- start_test_3.sh
- start_test_4.sh
- start_test_5.sh

Thanks to Akash Dhaka who investigated this issue with me.

---

> [!NOTE]
> **Medium Risk**
> Changes benchmark harness defaults for a narrow model+SKU and alters NCCL/MoRI QoS env propagation cluster-wide when QoS is detected; wrong gates could affect non-MI325X runs or mask fabric misconfiguration.
> 
> **Overview**
> Fixes SGLang MoRI **prefill/decode disaggregated** benchmarks that used to hang or stall under load on **MI325X**, especially **DeepSeek-R1-0528** with **decode DP+EP** and cross-node MoE all-to-all.
> 
> **`env.sh`** introduces **`RUNNER_FAMILY`** (normalized from `RUNNER_TYPE`) so SKU-specific tuning can target `mi325x` without matching fleet labels like `cluster:mi325x-amds`. When RDMA QoS is detected, traffic class and service level are **propagated consistently** to MoRI IO and **NCCL** (`NCCL_IB_TC` / `NCCL_IB_SL`), including deriving SL from TC when missing—avoiding bnxt_re DSCP/SL mismatches that downgrade to best-effort and stall KV/NCCL under load. For **DeepSeek-R1-0528 on mi325x**, decode **`MORI_MAX_DISPATCH_TOKENS_DECODE`** is raised from **512 → 4096** (scoped to that model+SKU).
> 
> **`server_sglang.sh`** applies the same model+SKU gate to set **`PREFILL_MEM_FRACTION_STATIC=0.9`** after the models.yaml loader (so it is not clobbered). The **DP+EP decode** path no longer ties **`max-running-requests`** to bench concurrency alone or shrinks MoRI dispatch buffers to `conc/dp_ranks`; it scales **`max-running-requests`** to **`conc × TP`** (floored at `dp_ranks`) and **keeps** env.sh MoRI dispatch/MOE limits instead of overriding them downward.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e9863fc078919ebdf086ce45439bf92dbeb7a6a5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## 中文说明
修复 SGLang MoRI 分离式基准测试在 MI325X 上的挂起问题。通过调优 RoCE/NCCL/MoRI 环境变量默认值和修复解码端 DP+EP 尺寸设置来解决（之前在负载下会导致跨节点 EP 资源不足）。主要变更包括：`env.sh` 新增默认 `NCCL_IB_GID_INDEX`、`NCCL_IB_TC`、`NCCL_IB_SL` 用于 bnxt_re RoCEv2 多节点通信；解码 MoRI dispatch 上限从 512 提升至 4096 tokens；`server_sglang.sh` 修正 `max-running-requests` 缩放逻辑；新增启动辅助脚本 `_disagg_ssh_remote_inner.sh`、`detect_ibdevices_bnxt.sh`、`rebuild_bnxt.sh` 等。


start_test_1.sh
```diff
@@ -0,0 +1,16 @@
...
+_IX_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" && pwd)"
+
+export PREFILL_MODEL_HOST_DIR=/nfsdata/sa/models
+export DECODE_MODEL_HOST_DIR=/nfsdata/sa/models
+export IMAGE=lmsysorg/sglang-rocm:v0.5.10rc0-rocm700-mi30x-20260420
+
+export REBUILD_LIBBNXT_IN_CONTAINER=1
+export PATH_TO_BNXT_TAR_PACKAGE=/workspace/driver/libbnxt_re-237.1.137.0.tar.gz
+
+export PREFILL_NODE="45.63.71.103"
```
1am9trash: The start_test_*.sh scripts hardcode many cluster-specific values (real node IPs like 45.63.71.103 / 137.220.60.12 and Docker image tags) directly in a public repo. Maybe we could parameterize these (e.g. via env vars) instead of committing concrete cluster details.

run_1p1d_tp16_sglang_mi300_mi325x.sh
```diff
@@ -0,0 +1,294 @@
...
+# Engine logic unchanged: scripts/_disagg_ssh_remote_inner.sh + benchmarks/.../server.sh
+# (server_sglang.sh already handles DECODE_NODES_PER_WORKER>1 via --dist-init-addr/
+#  --nnodes/--node-rank; this fork only feeds it the right IPADDRS/NODE_RANK layout).
+#
+# Topology: NODE_RANK=0 = prefill + router + benchmark; NODE_RANK=1,2 = the two
+# decode nodes of the single TP16 decode worker (decode head = NODE_RANK 1).
+# IPADDRS order: prefill_ip,decode1_ip,decode2_ip. NNODES=3, xP=1, yD=1.
+#
+# Prerequisites:
+#   - Passwordless SSH from this machine to both nodes
```
1am9trash: Should this say 3 nodes? This script launches 1 prefill + 2 decode nodes (NNODES=3), but the comments/header still say "both nodes", which were copy-pasted from the single-decode launcher and read as if there are only 2 nodes total.

start_test_1.sh
davzhuAMD: Thanks for your comments. These test scripts themselves (including the new start_test_*, run_1p*, and start_sglang*) don't necessarily need to make it into the final repo, but I added them to illustrate how we investigated and reproduced the original issue on our side. If it makes things clearer, I can remove the helper scripts from the commit and attach them somewhere else instead, leaving just the actual fixes in this PR.

driver/libbnxt_re-237.1.137.0.tar.gz
davzhuAMD, line 1: Guidance would be appreciated for how to handle a library file like this. During investigation and debug, we requested an update of our test nodes' NIC firmware and driver. However, the Docker image we used didn't have the updated driver itself, so our test scripts had to pass export REBUILD_LIBBNXT_IN_CONTAINER=1 to patch in this updated driver. In theory, using an updated image with a later compatible NIC driver should make this archive unnecessary to commit.

benchmarks/multi_node/amd_utils/rebuild_bnxt.sh
davzhuAMD, line 1: Regarding my comment on libbnxt_re-237.1.137.0.tar.gz https://github.com/SemiAnalysisAI/InferenceX/pull/1806/changes#r3432816594, if anyone validates this fix using a Docker image that comes with an updated NIC driver compatible with the host's firmware, then this rebuild_bnxt.sh script becomes unnecessary.

davzhuAMD:
I removed the added scripts and driver package, as I considered them out of the scope of the InferenceX-side fix for the original issue. Those removed scripts would've mainly covered environment setup, which anyone testing or validating this change would likely need to tailor to their own setup anyway. The scripts still exist at my https://github.com/davzhuAMD/InferenceX/tree/sglang-pd-mi325x as I use them for my own local testing.

chunfangamd (commented):
Thanks @davzhuAMD! The fixes are for MI325X, right? It seems the files are common files that can affect MI300X and MI355 too. Could you please check?

Meanwhile, please trigger an e2e test using your branch for verification.

davzhuAMD:
My pleasure, @chunfangamd, and thanks for your comment.

The changes to server_sglang.sh, and to the env.sh's NCCL and MORI QoS, are meant to be common, as they resolve what I believe is a genuine defect in the original code. Regardless of what node setup we have, the original code inconsistently set NCCL and MORI RDMA environment variables as one contributor to the original #22072 symptom, and my changes try to resolve that.
However, I could look into restricting the MORI_MAX_DISPATCH_TOKENS_DECODE=4096 and models.yaml to affect MI325X only, if we intend to only target MI325X as part of this change.

As for triggering an e2e test, I don't seem to have permission to do so via GitHub Actions. How do I get access?

davzhuAMD:
I updated my branch to set config settings only if it's DeepSeek-R1-0528 and RUNNER_TYPE mi325x, including for the prefill mem_fraction_static 0.9 that I formerly set in models.yaml (now models.yaml is unmodified). On that topic, what would be the best way to introduce MI325X-specific logic within env.sh and server_sglang.sh? Using RUNNER_TYPE like I have here, or getting the device ID somehow?

I'm currently retesting.

chunfangamd:
Thanks @davzhuAMD! The MI325X-specific scoping looks good. Could you please address or explain the following two concerns?

1. NCCL_IB_GID_INDEX=3 should not be a global default in shared env.sh; it is fabric/NIC-specific. Am I right?
2. Could you provide an E2E result for the original failing MI325X cases (#3/#4/#5). Since QoS and DP+EP sizing remain shared changes, we also need MI300X/MI355X regression coverage, or narrower scoping. The goal is to verify that we are not touching non-MI325 devices.

davzhuAMD:
Thank you for your comments, @chunfangamd.
> 1. NCCL_IB_GID_INDEX=3 should not be a global default in shared env.sh; it is fabric/NIC-specific. Am I right?

Good point, I'll remove this setting. The correct value could change depending on the exact cluster setup and layout.

> 2. Could you provide an E2E result for the original failing MI325X cases ([[NVIDIA] update vllm b200 image. TODO: add logic for docker runner. #3](https://github.com/SemiAnalysisAI/InferenceX/pull/3)/[[NVIDIA] Fix vllm & sglang b200 updated containers #4](https://github.com/SemiAnalysisAI/InferenceX/pull/4)/[Add new FW (TRT) and precision support #5](https://github.com/SemiAnalysisAI/InferenceX/pull/5)). Since QoS and DP+EP sizing remain shared changes, we also need MI300X/MI355X regression coverage, or narrower scoping. The goal is to verify that we are not touching non-MI325 devices.

Please see my previous comment regarding E2E. https://github.com/SemiAnalysisAI/InferenceX/pull/1806#issuecomment-5096705666 I thought I'd normally click a "Run workflow" button to set up an E2E test, but I don't have see one here. How do I get access, or does "E2E" refer to something else?
image

davzhuAMD:
Hi @chunfangamd and all, here are Terminal logs from successful runs of Tests 3 to 5 using this PR.

- [Terminal_3.txt](https://github.com/user-attachments/files/31977655/Terminal_3.txt)
- [Terminal_4.txt](https://github.com/user-attachments/files/31977662/Terminal_4.txt)
- [Terminal_5.txt](https://github.com/user-attachments/files/31979408/Terminal_5.txt)

davzhuAMD:
I updated my fork to a later SemiAnalysisAI/InferenceX baseline, and the formerly-failing tests still pass.

- [Terminal_3.txt](https://github.com/user-attachments/files/32126493/Terminal_3.txt)
- [Terminal_4.txt](https://github.com/user-attachments/files/32126494/Terminal_4.txt)
- [Terminal_5.txt](https://github.com/user-attachments/files/32126495/Terminal_5.txt)

PR #1914, opened 2026-06-24
[AMD] Add MiniMax-M3-MXFP4 MI355X vLLM disagg recipe

Duyi-Wang:
Disaggregated (prefill/decode) vLLM recipe for `amd/MiniMax-M3-MXFP4` on MI355X over the MoRI-IO KV connector.

### Recipe
- `benchmarks/multi_node/minimaxm3_fp4_mi355x_vllm-disagg.sh`: launcher.
- `models_vllm.yaml`: `MiniMax-M3-MXFP4` entry. `--block-size 128` (MSA), `TRITON_ATTN`, `--language-model-only`, AITER MoE, `minimax_m3` parsers, `--max-num-seqs 512`.
- `amd-master.yaml`: `minimaxm3-fp4-mi355x-vllm-disagg` config, 8k1k, two TP4 layouts (1P1D and 2P1D), conc 1..512.

### Supporting fixes to the shared vllm-disagg path
- `server_vllm.sh`: count prefill/decode GPUs from the per-worker TP size (`PREFILL_TP_SIZE*xP` / `DECODE_TP_SIZE*yD`) instead of `GPUS_PER_NODE*xP`. With TP < node GPU count (e.g. TP4 on an 8-GPU node) the old expression over-counted, corrupting `PREFILL_GPUS`/`DECODE_GPUS` and halving `tput_per_gpu`.
- `env.sh` / `job.slurm`: set the MoRI-IO RDMA QP knobs (`MORI_IO_QP_MAX_SEND_WR` etc.) for the vllm-disagg path. They were only set in the SGLang branch, so vllm-disagg ran at the default send-queue depth and stalled at high concurrency ("SQ full"). Injected via docker `-e` so they reach the vLLM worker processes.


benchmarks/multi_node/amd_utils/env.sh
```diff
@@ -118,6 +118,12 @@ $1 == "DSCP" && $2 == ":" && $NF == p {
     set +x
     echo "[INFO] IBDEVICES=$IBDEVICES  UCX_NET_DEVICES=$UCX_NET_DEVICES  NCCL_SOCKET_IFNAME=$NCCL_SOCKET_IFNAME  UCX_IB_GID_INDEX=$UCX_IB_GID_INDEX  UCX_IB_TRAFFIC_CLASS=${UCX_IB_TRAFFIC_CLASS:-unset}"
 
+    export MORI_IO_SQ_BACKOFF_TIMEOUT_US=50000
```
chunfangamd: @Duyi-Wang, 1. These changes will also affect the other models, right? 2. The numbers are currently hardcoded. 
```
export MORI_IO_SQ_BACKOFF_TIMEOUT_US="${MORI_IO_SQ_BACKOFF_TIMEOUT_US:-50000}"
export MORI_IO_QP_MAX_SEND_WR="${MORI_IO_QP_MAX_SEND_WR:-16384}"
...
```

benchmarks/multi_node/amd_utils/server_vllm.sh
```diff
@@ -302,7 +302,7 @@ if [ "$NODE_RANK" -eq 0 ]; then
     cd $WS_PATH
 
     export ROUTER_PORT=$ROUTER_PORT
-    BENCH_CMD="bash $WS_PATH/bench.sh ${xP} ${yD} $((GPUS_PER_NODE*xP)) $((GPUS_PER_NODE*yD)) \
+    BENCH_CMD="bash $WS_PATH/bench.sh ${xP} ${yD} $((PREFILL_TP_SIZE*xP)) $((DECODE_TP_SIZE*yD)) \
```
chunfangamd, line 308: Had a similar concern here. Please check

```
PREFILL_TP_SIZE="${PREFILL_TP_SIZE:-$GPUS_PER_NODE}"
DECODE_TP_SIZE="${DECODE_TP_SIZE:-$GPUS_PER_NODE}"
```

chunfangamd (changes requested):

benchmarks/multi_node/amd_utils/env.sh
Duyi-Wang: These general changes are for AINIC and are specific to MORI. They indeed affect the other models, but they also need these changes.

benchmarks/multi_node/amd_utils/server_vllm.sh
Duyi-Wang: The chain already has fallbacks (submit.sh computes it; job.slurm/SGLang/ATOM default it to 8), so it never reaches bench.sh as 0 — but a guard here is still worth having, so I added
  `${PREFILL_TP_SIZE:-$GPUS_PER_NODE} / ${DECODE_TP_SIZE:-$GPUS_PER_NODE}`

benchmarks/multi_node/amd_utils/env.sh
Duyi-Wang: The MORI-related changes have already been abstracted out, and this part will be shared by both vLLM and SGLang going forward.

.github/configs/amd-master.yaml
```diff
@@ -2587,6 +2587,39 @@ minimaxm3-fp8-mi355x-vllm-mtp:
       - { tp: 4, conc-start: 1, conc-end: 64, spec-decoding: mtp }
       - { tp: 8, ep: 8, dp-attn: true, conc-start: 128, conc-end: 256, spec-decoding: mtp }
 
+# MiniMax-M3 MXFP4 MI355X vLLM disaggregated (prefill/decode) config.
+minimaxm3-fp4-mi355x-vllm-disagg:
+  image: rocm/sgl-dev:vllm-0.23.1-rocm723-mi35x-mori-0624-2
```
functionstackx: what is an sgl-vllm image? do u have an acutal vllm image?

functionstackx (changes requested):

.github/configs/amd-master.yaml
Duyi-Wang: Updated.

PR #579, opened 2026-01-27
chore: add pre-merge check for newline in perf-changelog.yaml

cquil11:
Add a validation step in run-sweep.yml that ensures perf-changelog.yaml ends with a newline character. This prevents negative diff issues in subsequent PRs when the file is appended to.

Closes #578

Generated with [Claude Code](https://claude.ai/code)

PR #1935, opened 2026-06-25
[codex] add MiniMax M3 FP4 MI355X vLLM benchmark

functionstackx:
## What changed

- add `minimaxm3-fp4-mi355x-vllm` to the AMD master config using `amd/MiniMax-M3-MXFP4`
- pin the latest immutable ROCm vLLM nightly, `nightly-3f5a1e1733200760169ff31ebe60a271072b199e`
- add the MI355X vLLM launcher using the text-only language-model path, TRITON_ATTN, and MiniMax-M3 tool/reasoning parsers while leaving MoE backend selection to vLLM
- mirror the existing MXFP8 MI355X TP/EP/DP-attention sweep for direct precision comparisons
- append the required performance changelog trigger

## Why

AMD's MiniMax M3 MXFP4 checkpoint now has upstream vLLM support through vllm-project/vllm#45794. This adds benchmark coverage for that path on MI355X while keeping the FP8 and FP4 sweep shapes comparable.

doesn't enable AITER as it is still WIP and as of June 25, 2026 12:42pm PT it is not accessible to upstream docker containers https://github.com/vllm-project/vllm/pull/46419

## Validation

- `bash -n benchmarks/single_node/fixed_seq_len/minimaxm3_fp4_mi355x_vllm.sh`
- generated the filtered FP4 MI355X vLLM sweep successfully
- `python -m pytest utils/matrix_logic/ -v` (180 passed)

### DIFF
--- a/.github/configs/amd-master.yaml
+++ b/.github/configs/amd-master.yaml
@@ -2636,6 +2636,36 @@ minimaxm3-fp4-mi355x-vllm-disagg:
           dp-attn: false
           additional-settings:
           - "DECODE_NODES=1"
+# MiniMax-M3 MXFP4 MI355X vLLM recipe. The pinned nightly includes upstream
+# MiniMax-M3 Quark MXFP4 support (vllm-project/vllm#45794). Use the text-only
+# language-model path and mirror the MXFP8 MI355X search space for a direct
+# precision comparison.
+minimaxm3-fp4-mi355x-vllm:
+  image: vllm/vllm-openai-rocm:nightly-3f5a1e1733200760169ff31ebe60a271072b199e
+  model: amd/MiniMax-M3-MXFP4
+  model-prefix: minimaxm3
+  runner: mi355x
+  precision: fp4
+  framework: vllm
+  multinode: false
+  scenarios:
+    fixed-seq-len:
+    - isl: 1024
+      osl: 1024
+      search-space:
+      - { tp: 8, conc-start: 1, conc-end: 64 }
+      - { tp: 8, ep: 8, conc-start: 1, conc-end: 512 }
+      - { tp: 4, conc-start: 1, conc-end: 64 }
+      - { tp: 4, ep: 4, conc-start: 64, conc-end: 512 }
+      - { tp: 2, ep: 2, conc-start: 16, conc-end: 128 }
+      - { tp: 8, ep: 8, dp-attn: true, conc-start: 256, conc-end: 1024 }
+    - isl: 8192
+      osl: 1024
+      search-space:
+      - { tp: 8, conc-start: 1, conc-end: 64 }
+      - { tp: 8, ep: 8, conc-start: 1, conc-end: 512 }
+      - { tp: 4, conc-start: 1, conc-end: 128 }
+      - { tp: 8, ep: 8, dp-attn: true, conc-start: 128, conc-end: 512 }
 
 # MiniMax-M3 MXFP4 MI355X atom recipe:
 # https://github.com/ROCm/ATOM/blob/5d42d49f9e4292e5b61475917e92e7ec1b1dacb7/recipes/MiniMax-M3.md
--- a/benchmarks/single_node/fixed_seq_len/minimaxm3_fp4_mi355x_vllm.sh
+++ b/benchmarks/single_node/fixed_seq_len/minimaxm3_fp4_mi355x_vllm.sh
@@ -31,6 +31,7 @@ fi
 
 SERVER_LOG=/workspace/server.log
 export VLLM_ENGINE_READY_TIMEOUT_S=3600
+export VLLM_USE_BREAKABLE_CUDAGRAPH=0
 
 if [ "${EVAL_ONLY}" = "true" ]; then
     setup_eval_context
--- a/benchmarks/single_node/fixed_seq_len/minimaxm3_fp4_mi355x_vllm.sh
+++ b/benchmarks/single_node/fixed_seq_len/minimaxm3_fp4_mi355x_vllm.sh
@@ -1,0 +1,87 @@
+#!/usr/bin/env bash
+
+# MiniMax-M3 MXFP4 MI355X (gfx950) single-node vLLM recipe.
+# https://huggingface.co/amd/MiniMax-M3-MXFP4#reproduction
+# Block size 128 is mandatory for MSA. This fixed-sequence benchmark uses the
+# text-only language-model path and lets vLLM select the MoE backend.
+
+source "$(dirname "$0")/../../benchmark_lib.sh"
+
+check_env_vars \
+    MODEL \
+    TP \
+    EP_SIZE \
+    DP_ATTENTION \
+    CONC \
+    ISL \
+    OSL \
+    MAX_MODEL_LEN \
+    RANDOM_RANGE_RATIO \
+    RESULT_FILENAME
+
+if [[ -n "$SLURM_JOB_ID" ]]; then
+  echo "JOB $SLURM_JOB_ID running on $SLURMD_NODENAME"
+fi
+
+if [[ "$MODEL" != /* ]]; then hf download "$MODEL"; fi
+
+if [ -n "$ROCR_VISIBLE_DEVICES" ]; then
+    export HIP_VISIBLE_DEVICES="$ROCR_VISIBLE_DEVICES"
+fi
+
+SERVER_LOG=/workspace/server.log
+export VLLM_ENGINE_READY_TIMEOUT_S=3600
+
+if [ "${EVAL_ONLY}" = "true" ]; then
+    setup_eval_context
+fi
+
+PARALLEL_ARGS=(--tensor-parallel-size "$TP")
+if [ "${DP_ATTENTION}" = "true" ]; then
+    PARALLEL_ARGS=(
+        --tensor-parallel-size 1
+        --data-parallel-size "$TP"
+        --enable-expert-parallel
+    )
+elif [ "$EP_SIZE" -gt 1 ]; then
+    PARALLEL_ARGS+=(--enable-expert-parallel)
+fi
+
+start_gpu_monitor
+
+set -x
+vllm serve "$MODEL" --port "$PORT" \
+    "${PARALLEL_ARGS[@]}" \
+    --trust-remote-code \
+    --block-size 128 \
+    --no-enable-prefix-caching \
+    --language-model-only \
+    --max-model-len "$MAX_MODEL_LEN" \
+    --attention-backend TRITON_ATTN \
+    --tool-call-parser minimax_m3 \
+    --enable-auto-tool-choice \
+    --reasoning-parser minimax_m3 > "$SERVER_LOG" 2>&1 &
+
+SERVER_PID=$!
+wait_for_server_ready --port "$PORT" --server-log "$SERVER_LOG" --server-pid "$SERVER_PID"
+
+run_benchmark_serving \
+    --model "$MODEL" \
+    --port "$PORT" \
+    --backend vllm \
+    --input-len "$ISL" \
+    --output-len "$OSL" \
+    --random-range-ratio "$RANDOM_RANGE_RATIO" \
+    --num-prompts "$((CONC * 10))" \
+    --max-concurrency "$CONC" \
+    --result-filename "$RESULT_FILENAME" \
+    --result-dir /workspace/ \
+    --trust-remote-code
+
+if [ "${RUN_EVAL}" = "true" ]; then
+    run_eval --framework lm-eval --port "$PORT"
+    append_lm_eval_summary
+fi
+
+stop_gpu_monitor
+set +x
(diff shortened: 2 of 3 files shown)


andyluo7:
I had a duplicate of this in #1936 (same config key `minimaxm3-fp4-mi355x-vllm`) — closing mine in favor of this one since the public upstream nightly image (Quark MXFP4 via vllm-project/vllm#45794) is the better, recipe-alignment-friendly choice. Two findings from validating the FP4 single-node path on 8x MI355X (gfx950) that are worth folding in here:

1. **`VLLM_USE_BREAKABLE_CUDAGRAPH=0` is missing.** The merged MI355X CUDA-graph work (#1745/#1754) and the MXFP8 single-node script set this env on AMD specifically to keep CUDA graphs on for MiniMax-M3; without it the decode path can fall back to eager on gfx950 (perf loss). Recommend adding it to the script (or `extra_env`). If the sweep logs show eager decode, this is the cause.

2. **MoE backend is left to vLLM's default.** I validated explicit `--moe-backend aiter` + `VLLM_ROCM_USE_AITER=1 VLLM_ROCM_USE_AITER_MOE=1` (this is what #1914's FP4 disagg path uses too). Worth confirming the nightly defaults to AITER for MXFP4 MoE on gfx950 — otherwise throughput will differ from the disagg numbers.

Other things I confirmed that corroborate this PR's choices: block-size 128 + TRITON_ATTN MSA work; the `minimax_m3` reasoning parser splits `reasoning`/`content` correctly; and keeping the **default KV-cache dtype is correct** — `--kv-cache-dtype fp8` does *not* crash on vLLM (unlike ATOM), it just falls back to an uncalibrated scale of 1.0 with an accuracy warning.

FYI the matching upstream recipe variant is up at vllm-project/recipes#579.

One open question for you: search space. This mirrors the MXFP8 grid (EP + conc→1024); my closed PR used pure TP4+TP8 capped at conc 256 per the original ask. Either is fine — flagging in case you want to trim.

andyluo7:
Heads up — the pinned image is likely a blocker for MXFP4, separate from the flags above. Per the AMD MXFP4 enablement owners:

> A **plain** ROCm nightly won't bring up M3 MXFP4: it needs aiter `0.1.16.post2` (vllm-project/vllm#46692, merged) **and** the MoE enablement vllm-project/vllm#46419, which **isn't merged yet**. Until #46419 lands, build from qli88's branch `qiang_minimax_mxfp4_aiter` on the post2 image; switch to a plain nightly once #46419 merges.

This PR pins `vllm/vllm-openai-rocm:nightly-3f5a1e1733200760169ff31ebe60a271072b199e` (a plain nightly), so the running sweep will probably fail or fall back on MXFP4 MoE. Two options:
- swap to a qli88-`qiang_minimax_mxfp4_aiter` / post2 build, or
- reuse `rocm/vllm-dev:vllm-0.23.1-rocm723-mi35x-mori-0625` — I verified that image **does** serve `amd/MiniMax-M3-MXFP4` (engine reports `quantization=quark` + `moe_backend=aiter`, loads `minimax_m3_fp4_tuned_fmoe.csv`, coherent output). It already carries the MXFP4 aiter-MoE path.

Required serve flags (from #46419, matches what I validated):
```
VLLM_ROCM_USE_AITER=1 VLLM_ROCM_USE_AITER_MOE=1 VLLM_USE_BREAKABLE_CUDAGRAPH=0 \
vllm serve <model> --block-size 128 -tp 4 \
  --attention-backend TRITON_ATTN \
  --tool-call-parser minimax_m3 --enable-auto-tool-choice \
  --reasoning-parser minimax_m3 --moe-backend aiter
```

Accuracy target to validate against: **gsm8k 5-shot 0.940 flexible / 0.941 strict**.

functionstackx:
>MoE backend is left to vLLM's default. I validated explicit --moe-backend aiter + VLLM_ROCM_USE_AITER=1 VLLM_ROCM_USE_AITER_MOE=1

hi @andyluo7 

we do not enable AITER as it is still WIP and as of June 25, 2026 12:42pm PT it is not accessible to upstream docker containers https://github.com/vllm-project/vllm/pull/46419

I am happy to update this PR once it is merged and accesisble into an https://hub.docker.com/r/vllm/ docker image

> swap to a qli88-qiang_minimax_mxfp4_aiter / post2 build, or

while i am glad, that there is development build that it works on, it is not accessible to upstream https://hub.docker.com/r/vllm/ docker image . Feel free to create an update PR once it is accessible to upstream docker image

ChuanLi1101 (commented):
Thanks @functionstackx. +1 to the blocker @claude / @andyluo7 already flagged — `VLLM_USE_BREAKABLE_CUDAGRAPH=0` is required (model-specific, set by every other M3 recipe incl. the `MXFP4` disagg at `models_vllm.yaml:44`). Please add at line 34 and I'm good to approve. Two things to add on top of their reviews:

**1. Make the "AITER off" status explicit (builds on @andyluo7's MoE-backend point).** Leaving MoE to vLLM's default is the right interim call: the MXFP4 AITER path isn't in the upstream docker yet (vllm-project/vllm#46419, gated behind the aiter bump #46692). But that means these are **non-AITER MXFP4 baseline** numbers and won't match the AITER FP4 disagg results in #1914. Asks: (a) call this out in the changelog entry so it's not read as the optimized path; (b) plan an AITER-enabled follow-up once #46419/#46474/#46692 land. (If the nightly already carries AITER and you want it now, set it explicitly per Andy's note rather than relying on the default.)

**2. The aiter 0.1.16.post2 `mla_reduce_v1` regression does NOT affect this PR.** That break only hits MLA models (DSR1/Kimi). M3 is not MLA (TRITON_ATTN here), so it never touches the MLA decode/reduce path — irrelevant to this benchmark now and after we flip AITER on. Flagging so nobody blocks this on the war-room regression.

Once `VLLM_USE_BREAKABLE_CUDAGRAPH=0` is in, LGTM.

functionstackx:
> FYI the matching upstream recipe variant is up at https://github.com/vllm-project/recipes/pull/579.

hi Andy, following ur recipe as of 1pm ET June 25, 2026 it doesn't work

functionstackx:
> Thanks @functionstackx. +1 to the blocker @claude / @andyluo7 already flagged — `VLLM_USE_BREAKABLE_CUDAGRAPH=0` is required (model-specific, set by every other M3 recipe incl. the `MXFP4` disagg at `models_vllm.yaml:44`). Please add at line 34 and I'm good to approve. Two things to add on top of their reviews:
> 
> **1. Make the "AITER off" status explicit (builds on @andyluo7's MoE-backend point).** Leaving MoE to vLLM's default is the right interim call: the MXFP4 AITER path isn't in the upstream docker yet ([vllm-project/vllm#46419](https://github.com/vllm-project/vllm/pull/46419), gated behind the aiter bump #46692). But that means these are **non-AITER MXFP4 baseline** numbers and won't match the AITER FP4 disagg results in #1914. Asks: (a) call this out in the changelog entry so it's not read as the optimized path; (b) plan an AITER-enabled follow-up once #46419/#46474/#46692 land. (If the nightly already carries AITER and you want it now, set it explicitly per Andy's note rather than relying on the default.)
> 
> **2. The aiter 0.1.16.post2 `mla_reduce_v1` regression does NOT affect this PR.** That break only hits MLA models (DSR1/Kimi). M3 is not MLA (TRITON_ATTN here), so it never touches the MLA decode/reduce path — irrelevant to this benchmark now and after we flip AITER on. Flagging so nobody blocks this on the war-room regression.

Thanks @ChuanLi1101 for seeing that AITER MoE is not accessible to any upstream https://hub.docker.com/r/vllm/ docker image thus not enabling it is an accurate reflection of the upstream performance as of jun 25, 1:30pm PT. Happy to accept follow PRs to track the performance once it is accessible to upstream vLLM docker

> Once `VLLM_USE_BREAKABLE_CUDAGRAPH=0` is in, LGTM.

thanks @ChuanLi1101 i have implement ur single suggestion of  `VLLM_USE_BREAKABLE_CUDAGRAPH=0`

Klaud-Cold:
**Claude finished @functionstackx's task in 52s** —— [View job](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/28198858556)

---
### Verified — `VLLM_USE_BREAKABLE_CUDAGRAPH=0` is in ✅

- [x] Read trigger comment
- [x] Confirm `VLLM_USE_BREAKABLE_CUDAGRAPH=0` is in the script
- [x] Check changelog entry
- [x] Post summary

Confirmed on the latest commit (`87cf65e9`):

**1. Script** — `benchmarks/single_node/fixed_seq_len/minimaxm3_fp4_mi355x_vllm.sh:34` now exports the env var in the right spot, exactly mirroring `minimaxm3_fp8_mi355x.sh:33`:

```bash
SERVER_LOG=/workspace/server.log
export VLLM_ENGINE_READY_TIMEOUT_S=3600
export VLLM_USE_BREAKABLE_CUDAGRAPH=0
```

This resolves my earlier 🔴 blocker — the M3 decode path will now run with CUDA graphs instead of silently falling back to eager, so the FP4↔FP8 precision comparison stays apples-to-apples. ✅

**2. Changelog** — `perf-changelog.yaml` already records the env var and the "let vLLM select the MoE backend / default KV-cache dtype" choices for `minimaxm3-fp4-mi355x-vllm`. ✅

On @ChuanLi1101's two follow-up points:
- **(2) MLA regression** — agreed, irrelevant. M3 runs `TRITON_ATTN` (not MLA), so the aiter `0.1.16.post2` `mla_reduce_v1` break can't touch this path. Not a blocker.
- **(1) "AITER-off" labeling** — the changelog already states MoE backend is left to vLLM's default, which is the accurate reflection of upstream-docker behavior as of now (the MXFP4 AITER MoE path in vllm-project/vllm#46419 isn't in any `hub.docker.com/r/vllm/` image yet). If you want it to be unmistakable that these are **non-AITER baseline** numbers vs. the AITER FP4 disagg in #1914, a one-line note in the changelog entry would make it explicit — optional, not blocking. Happy to push that small wording tweak if you'd like, otherwise this is good to go.

The intentional FP4 deltas (default KV-cache dtype, `--trust-remote-code`) remain consistent and documented. **No blocking issues remain.**
 • branch `codex/minimax-m3-fp4-mi355x-vllm`

andyluo7:
@functionstackx thanks for re-testing — two things:

**1. Your 1pm ET test predated my fix.** I pushed [recipes#579 92fe4c6](https://github.com/vllm-project/recipes/pull/579/commits) at ~4:40pm ET, after your run. The version you hit had no variant-level image pin (it fell back to the generic `vllm/vllm-openai-rocm:minimax-m3`) and still forced `--moe-backend aiter` → that's the crash. The updated variant now pins `rocm/vllm-dev:vllm-0.23.1-rocm723-mi35x-mori-0625`, which I verified serves `amd/MiniMax-M3-MXFP4` on AITER (engine reports `quantization=quark` + `moe_backend=aiter`, gsm8k 5-shot 0.939 strict / 0.939 flex). If you re-pull #579 the AITER path should come up on that image.

**2. On official `hub.docker.com/r/vllm/` images, use the emulation backend — the recipe documents this.** I fully agree AITER MXFP4 MoE isn't in any official upstream `vllm/` docker yet (vllm-project/vllm#46419 unmerged), so the AITER path requires the dev image until it lands. But the recipe also has a `--moe-backend emulation` path: it's the non-AITER reference backend (no AITER kernels, pure vLLM) that AMD used to measure accuracy, and it runs on current upstream images — just slower. So "following the recipe" on an official image does work via emulation; only the perf path needs the dev image. If your run used `--moe-backend aiter` on an official nightly, that's the mismatch.

This keeps the two PRs consistent: your #1935 = non-AITER baseline on official docker (correct, already merged), and #579 documents both the emulation path (official images, today) and the AITER perf path (dev image now, official nightly once #46419 ships). Happy to do an AITER-enabled InferenceX follow-up once it's in an official image.

functionstackx:
> **2. On official `hub.docker.com/r/vllm/` images, use the emulation backend

ur AI wants me to use MXFP4 emulation even tho emulation is slower & even then non-emulation MXFP4 works on upstream now and passes evals!? can u please upgrade ur ai to codex 5.5 or opus 4.8 xhigh plz

 https://github.com/SemiAnalysisAI/InferenceX/actions/runs/28195297568/job/83520505469

functionstackx:
> **1. Your 1pm ET test predated my fix.** I pushed [recipes#579 92fe4c6](https://github.com/vllm-project/recipes/pull/579/commits) at ~4:40pm ET, after your run.

ur recipe was/is wrong, as addressed in https://github.com/vllm-project/recipes/pull/579#pullrequestreview-4574709842 i am glad u were able to address my suggestions of where this bugs

functionstackx:
1. following the recipe as of june 25, 2026 3:56pm PT https://recipes.vllm.ai/MiniMaxAI/MiniMax-M3?hardware=mi355x&variant=mxfp4
2.  using the latest avaialble upstream docker repo container  available currently `vllm/vllm-openai-rocm:nightly-3f5a1e1733200760169ff31ebe60a271072b199e`
3. verified evals  https://github.com/SemiAnalysisAI/InferenceX/actions/runs/28195297568/job/83520505469 matches https://huggingface.co/amd/MiniMax-M3-MXFP4
  

image

PR #1944, opened 2026-06-26
exclude B300-004 temporarily to avoid kv tx issues due to mlx5_10

biswapanda:
Intermittent KV transfer failure on a specific node/NIC ( b300-004 / mlx5_10) is causing garbage results. 

Silent KV corruption - KV transfer fails and decode continues without valid prefill KV frontend still returns HTTP 200 -


functionstackx:
hi @biswapanda this is an good temp patch, tho can u talk to NVIDIA SRE to exclude it from the slurm side and report this cluster issue to them?

+viz @Ankur-singh

biswapanda:
Thanks @functionstackx ! I talked to @Ankur-singh and this node will be excluded from Slurm cluster.

PR #1947, opened 2026-06-26
[WIP] add SWE-bench Lite accuracy eval / 添加 SWE-bench Lite 准确率评估

adibarra:
## Summary

Adds SWE-bench Lite as the accuracy evaluation for single-node `agentic-coding` recipes. Fixed-sequence-length recipes continue to use lm-eval tasks such as GSM8K; `run_eval` selects the framework from the scenario and still accepts an explicit override.

## Execution path

1. `run_eval` dispatches agentic recipes to `run_swebench_eval`.
2. mini-swe-agent 2.4.5 runs one coding agent per SWE-bench Lite instance against the model's local OpenAI-compatible endpoint.
3. swe-rex 1.4.0 creates the per-instance execution sandboxes on Modal. Agent concurrency defaults to the recipe concurrency and can be overridden with `SWEBENCH_AGENT_WORKERS`.
4. The generated prediction set is scored by the official SWE-bench 4.1.0 harness. CI uses Modal scoring sandboxes; local Docker remains available.
5. `swebench_score.py` publishes resolved rate in the existing lm-eval result shape, so artifact collection and `validate_scores.py` use the same downstream path as the other accuracy evaluations.

The CI default is a 50-instance slice. `EVAL_LIMIT=full` (or `0`) evaluates the complete SWE-bench Lite split. The score gate is `exact_match,resolved >= 0.50`; the denominator is the submitted instance count for sliced runs and the full set for full runs.

## Recipe and workflow integration

- Enables eval-only gating for all single-node `agentic-coding` recipes.
- Preserves benchmark-only runs: the agentic accuracy evaluation is entered only by the existing eval-only/all-evals controls.
- Threads Modal credentials through the benchmark workflow and bootstraps `~/.modal.toml` when CI supplies token environment variables.
- Uploads predictions, trajectories, official harness reports, and the normalized result JSON even when generation or scoring fails.
- Adds watchdog and timeout controls for agent generation, per-instance scoring, and whole-run scoring so stalled sandboxes do not hold the GPU job indefinitely.

## Runtime compatibility patches

The external packages are pinned and patched at runtime because the required fixes are not available in the pinned releases:

- mini-swe-agent: terminate the environment on every exit and submit a valid working-tree diff when the agent exhausts its step budget without a submission.
- swe-rex: correct Modal sandbox termination and clean up startup failures; use an app name scoped to this evaluation.
- SWE-bench Modal scorer: reserve the configured CPU count and terminate each scoring sandbox when its instance completes.
- lm-eval: accept `reasoning_content` when `content` is empty and avoid adding typed text wrappers unsupported by the TRT endpoint.

The patch implementations live in `utils/evals/patches/`; `benchmark_lib.sh` only installs pinned dependencies and invokes those scripts. Every source rewrite is anchor-checked, idempotent, and atomic: an upstream source mismatch fails without writing a partial patch.

## Validation

Successful end-to-end workflow runs on the PR branch:

| Run | Scope | Result |
| --- | --- | --- |
| [29061184942](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/29061184942) | Production-path defaults, 50 instances | 32/50 resolved (64%) |
| [29109355974](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/29109355974) | Post-review merge-readiness validation, 50 instances | 36/50 resolved (72%) |
| [29123068609](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/29123068609) | Full workflow and sandbox lifecycle cost validation, 50 instances | 38/50 resolved (76%) |
| [29163378164](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/29163378164) | App-scoped sweep and concurrency-matched workers, 50 instances | 38/50 resolved (76%) |

All four workflow runs completed successfully, including generation, official scoring, artifact collection, score validation, and the hardware success-rate job.


benchmarks/benchmark_lib.sh
```diff
@@ -918,11 +928,13 @@ run_lm_eval() {
     export EVAL_RESULT_DIR="$results_dir"
     set -x
     python3 -m lm_eval --model local-chat-completions --apply_chat_template \
+      ${include_path:+--include_path "$include_path"} \
       --tasks "${tasks_dir}" \
       --output_path "${results_dir}" \
       --log_samples \
       --model_args "model=${MODEL_NAME},base_url=${openai_chat_base},api_key=${OPENAI_API_KEY},eos_string=</s>,max_retries=5,num_concurrent=${concurrent_requests},timeout=1800,tokenized_requests=False,max_length=${eval_context_len}" \
-      --gen_kwargs "max_tokens=${max_output_tokens},temperature=${temperature},top_p=${top_p}"
+      --gen_kwargs "max_tokens=${max_output_tokens},temperature=${temperature},top_p=${top_p}" \
```
cquil11, line 871: why are these rquired now? but not before

benchmarks/benchmark_lib.sh
```diff
@@ -1161,21 +1173,735 @@ META
...
+#    that terminates the sandbox exists but nothing calls it), so every eval
+#    sandbox idle-bills after its tests finish until the 30-min sandbox
+#    timeout or app teardown. Invisible on the fast 50-slice (the whole app
+#    ends in ~2 min); on full-300 the slow tail keeps the app alive ~40 min
+#    and all 300 sandboxes bill ~30 min for ~3 min of work. A finally: on the
+#    function's main try/except chain terminates the sandbox on every exit
+#    path the moment the instance's evaluation ends.
+_patch_swebench_scoring() {
+    python3 - <<'PYPATCH'
+import os, sys
```
cquil11: @claude in order for better maintainability, can you move all of the patches in benchmark_lib to separate python files in evals/patches or something, then add the patcehs in this file instead of hardcoding the entire patch inline

Klaud-Cold: **Claude finished @cquil11's task in 6m 32s** —— [View job](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/29286204601)

---
### Refactored inline patches out of benchmark_lib.sh ✅

- [x] Gather context: found 3 inline patch heredocs in `benchmarks/benchmark_lib.sh`
- [x] Move patch code into separate Python files under `utils/evals/patches/`
- [x] Update `benchmark_lib.sh` to invoke the patch files instead of inlining them
- [x] Update EVALS.md references
- [x] Run relevant tests (52 passed, 1 skipped — same Docker-gated skip as before)
- [x] Commit and push → df5e9d53

### What changed

All three inline patch heredocs moved verbatim into standalone files under `utils/evals/patches/`, with the shell-comment rationale moved into each file's docstring:

| New file | Applied by | What it patches |
|---|---|---|
| `lm_eval_sitecustomize.py` | `_patch_lm_eval` | lm-eval reasoning-token handling + TRT chat-template compat (pre-existing on main, moved too since the ask was "all of the patches") |
| `patch_swebench_agent.py` | `_patch_swebench_agent` | mini-swe-agent/swe-rex sandbox lifecycle cleanup + budget-exhaustion submission fallback |
| `patch_swebench_scoring.py` | `_patch_swebench_scoring` | swebench Modal scorer reserved-CPU reduction + sandbox termination on instance completion |

The three `_patch_*` functions in `benchmark_lib.sh` are now 3–6 lines each; the file shrank by 260 net lines. A new `_eval_patches_dir` helper resolves `<repo>/utils/evals/patches` from `BASH_SOURCE`, the same technique `run_lm_eval` already uses to anchor task YAMLs, so it works when the eval runs inside a serving container whose cwd isn't the repo. `_patch_lm_eval` still copies its file into a temp dir as `sitecustomize.py` (the name Python auto-imports), so runtime behavior is unchanged.

### Verification

- Extracted bodies confirmed byte-identical to the original heredocs (programmatic diff against `HEAD`)
- `bash -n benchmark_lib.sh` and `py_compile` on all three files pass
- `test_run_eval_dispatch.py` + `test_swebench_eval.py`: 52 passed, 1 skipped
- Smoke-tested `_patch_lm_eval` sourced from `/tmp`: resolves and copies the patch file correctly

EVALS.md's "lm-eval patches" section is now "Runtime patches (`utils/evals/patches/`)" and documents all three files.
 | [Branch](https://github.com/SemiAnalysisAI/InferenceX/tree/feat/swebench-lite-eval)

benchmarks/benchmark_lib.sh
```diff
@@ -1161,21 +1173,735 @@ META
...
+        MODAL_TOKEN_ID=$(printf %s "$MODAL_TOKEN_ID" | tr -d "[:space:]\"'")
+        export MODAL_TOKEN_ID
+    fi
+    if [ -n "${MODAL_TOKEN_SECRET:-}" ]; then
+        MODAL_TOKEN_SECRET=$(printf %s "$MODAL_TOKEN_SECRET" | tr -d "[:space:]\"'")
+        export MODAL_TOKEN_SECRET
+    fi
+    if [ -f "${HOME:-}/.modal.toml" ]; then return 0; fi
+    if [ -n "${MODAL_TOKEN_ID:-}" ] && [ -n "${MODAL_TOKEN_SECRET:-}" ]; then
+        # On b300 slurm/pyxis, --export=ALL may propagate the HOST's HOME into
```
cquil11: kinda tech debt to have one off fix for b300 instead of just fixing root cause ? dont have full context here maybe this is the only way

utils/evals/thresholds.json
```diff
@@ -1,7 +1,8 @@
 {
   "default": {
     "gsm8k": 0.90,
-    "gpqa_diamond_cot_n_shot": 0.30
+    "gpqa_diamond_cot_n_shot": 0.30,
```
cquil11: @claude Not that it matters, but change this entire file to YAML and make any corresponding changes, because we use YAML everywhere else in this repository.

Klaud-Cold: **Claude finished @cquil11's task in 6m 4s** —— [View job](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/29287030787)

---
### Converted `thresholds.json` → `thresholds.yaml` ✅

- [x] Identify target file (`utils/evals/thresholds.json`, from the review thread)
- [x] Find all references to `thresholds.json` in the repo
- [x] Convert the file to YAML and update all readers
- [x] Update tests/docs
- [x] Run relevant tests (59 passed; 1 pre-existing, unrelated failure — see below)
- [x] Commit and push → 2289a2ee

### What changed

- `utils/evals/thresholds.json` → **`utils/evals/thresholds.yaml`**, same content (verified programmatically that the YAML parses to a dict identical to the old JSON, including `glm5.1` staying a string key).
- `utils/evals/validate_scores.py`: `load_config` now parses with `yaml.safe_load`, and the default path is `thresholds.yaml`. Docstring, example config, and `--thresholds` help text updated.
- `utils/evals/EVALS.md` and a comment in `test_swebench_eval.py` updated to the new filename.

### One design note: PyYAML on runner hosts

`validate_scores.py` runs bare (`python3 utils/evals/validate_scores.py`) on the self-hosted GPU runner hosts in `benchmark-tmpl.yml` / `benchmark-multinode-tmpl.yml`, with no `pip install` step — and nothing else running on those hosts currently imports `yaml`. To keep this safe:

- If PyYAML is present (it ships on virtually every Ubuntu host via cloud-init), the YAML config loads normally.
- If PyYAML is missing, the loader falls back to stdlib `json` — legacy JSON configs passed via `--thresholds` still work (JSON is a YAML subset either way), and the checked-in YAML config fails with an explicit `install it with 'pip install pyyaml'` error rather than silently weakening the score gate.

If a runner host ever hits that error, the fix is a one-line `pip install pyyaml` on the host (or in the workflow step).

### Verification

- `test_batched_eval.py` + `test_swebench_eval.py` + `test_run_eval_dispatch.py`: 59 passed, 1 failed — the failure (`test_include_path_absent_when_eval_include_path_unset`) also fails on the untouched tree in this environment (it's sensitive to running from a real repo checkout where `run_lm_eval` anchors relative task paths to absolute), so it's unrelated to this change.
- End-to-end: ran `validate_scores.py` against the new YAML with a `qwen3.5` per-model override — correctly fails 0.93 < 0.94 and passes 0.95 ≥ 0.94.
- Simulated a missing-PyYAML environment: legacy JSON config loads, YAML config raises the actionable error.
 | [Branch](https://github.com/SemiAnalysisAI/InferenceX/tree/feat/swebench-lite-eval)

benchmarks/benchmark_lib.sh
```diff
@@ -1161,21 +1106,376 @@ META
...
+        chmod 600 "$HOME/.modal.toml"
+        echo "[swebench] wrote ~/.modal.toml from MODAL_TOKEN_ID/MODAL_TOKEN_SECRET env"
+    else
+        echo "WARN: SWEBENCH_USE_MODAL=true but no ~/.modal.toml and no MODAL_TOKEN_ID/MODAL_TOKEN_SECRET env; Modal scoring will fail credential validation" >&2
+    fi
+}
+
+maybe_run_eval() {
+    local port="${1:-${PORT:-8888}}"
+    if [ "${RUN_EVAL}" = "true" ]; then
```
Oseltamivir: no need this gate if gated earlier with EVALS_ONLY

RUN_EVAL flag is a old workaround flag back when throughput is ran with evals as a single job

benchmarks/benchmark_lib.sh
```diff
@@ -1161,21 +1106,376 @@ META
...
+        printf '[default]\ntoken_id = "%s"\ntoken_secret = "%s"\nactive = true\n' \
+            "$MODAL_TOKEN_ID" "$MODAL_TOKEN_SECRET" > "$HOME/.modal.toml"
+        chmod 600 "$HOME/.modal.toml"
+        echo "[swebench] wrote ~/.modal.toml from MODAL_TOKEN_ID/MODAL_TOKEN_SECRET env"
+    else
+        echo "WARN: SWEBENCH_USE_MODAL=true but no ~/.modal.toml and no MODAL_TOKEN_ID/MODAL_TOKEN_SECRET env; Modal scoring will fail credential validation" >&2
+    fi
+}
+
+maybe_run_eval() {
```
Oseltamivir: maybe call this something else

benchmarks/benchmark_lib.sh
```diff
@@ -1161,21 +1106,376 @@ META
...
+        echo "ERROR: swebench scoring failed with $score_rc" >&2
+        return "$score_rc"
+    fi
+}
+
 # ------------------------------
 # Unified eval entrypoint
 # ------------------------------
 
 run_eval() {
```
Oseltamivir, line 1443: call this instead of the maybe_run_eval() directly?

benchmarks/benchmark_lib.sh
```diff
@@ -1161,21 +1106,376 @@ META
...
+                ;;
+            *)
+                forwarded+=("$1")
+                shift
+                ;;
         esac
     done
 
+    local scenario_default="lm-eval"
+    if [ "${IS_AGENTIC:-0}" = "1" ] || [ "${SCENARIO_TYPE:-}" = "agentic-coding" ]; then
```
Oseltamivir, line 1466: this will be preventive when adding gpqa/aime later

PR #1958, opened 2026-06-30
[Klaud Cold] [AMD] Enable AITER MoE for MiniMax-M3 MI355X FP4 vLLM MTP benchmark / 为 MiniMax-M3 MI355X FP4 vLLM MTP 基准测试启用 AITER MoE / MiniMax-M3 MI355X FP4 vLLM MTP 벤치마크에 AITER MoE 활성화

functionstackx:
Splits the **FP4 MTP** half out of #1955, rebased on current `main` (the original PR was conflicting because `main` already moved `minimaxm3-fp4-mi355x-vllm` STP to the new nightly via #1954). The FP8 MTP half will be a separate PR for easier merging.

## Changes
- **`benchmarks/single_node/fixed_seq_len/minimaxm3_fp4_mi355x_vllm_mtp.sh`** — enable AITER MoE **only on non-EP configs**:
  - When EP is off: `VLLM_ROCM_USE_AITER=1`, `VLLM_ROCM_USE_AITER_MOE=1`, `VLLM_ROCM_USE_AITER_FUSION_SHARED_EXPERTS=1`, and `--moe-backend aiter`.
  - When EP is enabled (`DP_ATTENTION=true` or `EP_SIZE > 1`): `VLLM_ROCM_USE_AITER=0` and no `--moe-backend aiter`, since AITER MoE is incompatible with expert parallelism.
- **`.github/configs/amd-master.yaml`** — bump `minimaxm3-fp4-mi355x-vllm-mtp` image to `nightly-4559c43a9526597c00cbcc4f59979496500268d1`.
- **`perf-changelog.yaml`** — changelog entry for the above.

## Review addressed
Implements @hongxiayang's review on #1955 (https://github.com/SemiAnalysisAI/InferenceX/pull/1955#discussion_r3495386866) — *"will need to set VLLM_ROCM_USE_AITER=0 when enable ep"* — via an `if`/`else` on the EP condition. This matters here because the FP4 MTP sweep includes several `ep: 8`, `ep: 4`, and `dp-attn: true` points.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## 中文说明
[...]

PR #1955, opened 2026-06-29
[AMD] Enable AITER MoE for MiniMax-M3 MI355X vLLM MTP benchmarks / 为 MiniMax-M3 MI355X vLLM MTP 基准测试启用 AITER MoE

Fangzhou-Ai:
## Summary

- Enable AITER MoE on the MiniMax-M3 MI355X EAGLE3 MTP launchers (`minimaxm3_fp4_mi355x_vllm_mtp.sh`, `minimaxm3_fp8_mi355x_mtp.sh`), mirroring the STP knobs from #1954.
- Export `VLLM_ROCM_USE_AITER=1`, `VLLM_ROCM_USE_AITER_MOE=1`, and `VLLM_ROCM_USE_AITER_FUSION_SHARED_EXPERTS=1`; pass `--moe-backend aiter`.
- Export `VLLM_ROCM_QUICK_REDUCE_QUANTIZATION=INT6` on the MXFP8 MTP path.
- Append `perf-changelog.yaml` triggers for `minimaxm3-fp4-mi355x-vllm-mtp` and `minimaxm3-fp8-mi355x-vllm-mtp`.

## Why

Manual validation showed EAGLE3 MTP serving works with the same AITER MoE stack as STP. This lands the benchmark coverage Andy asked to defer from #1954.

## Test plan

- [x] `bash -n` on both MTP benchmark scripts
- [x] `python utils/matrix_logic/generate_sweep_configs.py test-config --config-keys minimaxm3-fp4-mi355x-vllm-mtp minimaxm3-fp8-mi355x-vllm-mtp --config-files .github/configs/amd-master.yaml --no-evals`
- [ ] Apply `full-sweep-enabled` (or `full-sweep-fail-fast`) after #1954 lands, or dispatch e2e against this branch

Made with [Cursor](https://cursor.com)

## 中文说明
为 MiniMax-M3 MI355X EAGLE3 MTP 启动脚本（`minimaxm3_fp4_mi355x_vllm_mtp.sh`、`minimaxm3_fp8_mi355x_mtp.sh`）启用 AITER MoE，沿用 #1954 中 STP 配置的相关参数。导出 `VLLM_ROCM_USE_AITER=1`、`VLLM_ROCM_USE_AITER_MOE=1` 等环境变量，并传入 `--moe-backend aiter`。同时在 `perf-changelog.yaml` 中追加对应的触发条目。

PR #1954, opened 2026-06-29
[AMD] Enable AITER MoE for MiniMax-M3 FP4 MI355X vLLM STP

Fangzhou-Ai:
## Summary

- Export `VLLM_ROCM_USE_AITER_FUSION_SHARED_EXPERTS=1` in all MiniMax-M3 MI355X single-node vLLM benchmark scripts (MXFP4/MXFP8, STP and EAGLE3 MTP).
- Broaden the `perf-changelog.yaml` trigger to cover `minimaxm3-fp4-mi355x-vllm`, `minimaxm3-fp4-mi355x-vllm-mtp`, `minimaxm3-fp8-mi355x-vllm`, and `minimaxm3-fp8-mi355x-vllm-mtp`.

## Why

Upstream vLLM folds the MiniMax-M3 shared expert MLP into the routed grouped GEMM when this flag is set (vllm-project/vllm#46545), reducing per-layer kernel launches on decode-heavy workloads. The optimization is backend-agnostic and does not require the master `VLLM_ROCM_USE_AITER` switch, so it is safe to enable on the existing pinned nightly without turning on the full AITER MoE path.

## Test plan

- [x] `bash -n` on all four MiniMax-M3 MI355X vLLM benchmark scripts
- [x] `python utils/matrix_logic/generate_sweep_configs.py full-sweep --config-files .github/configs/amd-master.yaml --model-prefix minimaxm3 --framework vllm --runner-type mi355x`

PR #1964, opened 2026-06-30
[AMD] Enable AITER MoE for MiniMax-M3 FP4 MI355X vLLM MTP (fix EP startup hang)

Fangzhou-Ai:
## Problem
The MiniMax-M3 **FP4 MI355X vLLM MTP** sweep (`minimaxm3-fp4-mi355x-vllm-mtp`) hangs for ~8 h on **every expert-parallel config** (`ep=8`, `ep=4`, `dp-attn`). See the failed run [28443025904](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/28443025904) (PR #1958): all `ep>1` spec-mtp jobs ran 8h20m until the CI job timeout; all `ep=1` jobs passed.

## Root cause (debugged from the failed job log + vLLM source)
With EP enabled and `moe_backend=auto`, vLLM's MXFP4 MoE selection skips the AITER backend — EP turns on the `BatchedExperts` activation format, which `AITER_MXFP4_MXFP4` doesn't support — and falls back to **`Mxfp4MoeBackend.EMULATION`**. The emulation path builds the **Quark hw-emulation C++ kernel** (`quark/torch/kernel/hw_emulation/extensions.py` → `torch.utils.cpp_extension.load`, `kernel_ext`, *Building for 9 ROCm arches* `gfx90a;gfx942;gfx950;...;gfx1151`) at warmup. All EP workers build into the **shared** `/root/.cache/torch_extensions/.../kernel_ext` dir simultaneously and deadlock on the build lock, so engine-core never finishes startup:
```
QUARK-INFO: C++ kernel compilation check start ... First-time compilation may take a few minutes... Building for architectures PYTORCH_ROCM_ARCH='gfx90a;gfx942;gfx950;...'
... INFO shm_broadcast: No available shared memory broadcast block found in 60 seconds ...  (repeats for 8h)
```
The `ep=1` jobs pass because they run `--moe-backend aiter` (the AITER path, no emulation/build). FP8 MTP EP jobs ([28427946510](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/28427946510)) pass for the same reason — MXFP8 auto-selects the native `AITER_MXFP8` backend.

## Fix
Mirror the STP recipe `minimaxm3_fp4_mi355x_vllm.sh` (PR #1954) into the MTP script — export the AITER env knobs and pass `--moe-backend aiter` **unconditionally** (including EP). Forcing the AITER backend selects `AITER_MXFP4_MXFP4`, so the emulation C++ build never happens. Also pin the MTP image to `nightly-4559c43a9` (the verified tip), matching the STP recipe.

## Difference vs #1958 / #1955
Those PRs *disabled* AITER MoE under EP (`VLLM_ROCM_USE_AITER_MOE=0`, no `--moe-backend aiter`) believing AITER MoE is EP-incompatible. That is exactly what forces the emulation fallback and the build hang. **AITER MoE works under EP here** — this PR enables it instead of disabling it.

## Verification (local, pinned tip `4559c43a9` + aiter tip)
Ran the exact MTP recipe with **TP4 + EP4** (`--enable-expert-parallel --moe-backend aiter`) + EAGLE3 spec decoding:
- MoE backend selected: **`AITER_MXFP4_MXFP4`** (no `EMULATION`, no `kernel_ext` build).
- Engine starts and serves; correct output (`17×23=391`).
- Spec-decode healthy: **mean acceptance length ≈ 3.5** (per-position 0.91/0.91/0.73).
- `conc=16` burst: **32/32 requests successful**, mean TPOT 10.66 ms.
- For reference, `tp4 ep1` (no EP) MTP: GSM8K(100)=0.93, perf conc 1/32/128 = 6.10/20.04/87.28 ms.

> Caveat: verified against the local vLLM @ `4559c43a9` + aiter tip; the CI re-sweep on the pinned docker image will confirm end-to-end.


Fangzhou-Ai:
Hi @functionstackx @andyluo7 this is our latest minimax m3 fp4 MTP config can we merge this PR?
image

functionstackx:
@adibarra

adibarra (dismissed):

chunfangamd:
@haic0  could you please check whether we've updated the corresponding vllm-project/recipes

chunfangamd:
https://github.com/vllm-project/recipes/pull/609

PR #2117, opened 2026-07-08
[NV] llm-d: re-test max-tpt in-sweep with unique SLURM job-name fix

ezrasilvera:
## What

Isolated re-test of the max-tpt (3P+1D DEP8 conc4096) llm-d-vllm multinode
benchmark with a fix for the in-sweep failure.

## Why

max-tpt passes as a standalone e2e (run 28748597199, ~8747 tok/s/GPU) but its
throughput job was killed by an external SIGTERM ~10 min into weight load in
the isolated in-sweep run 28789948711 - before the server came up, so no
result JSON was produced and the launch step failed with "No benchmark result
files found". The engine loaded cleanly with zero errors and the twin eval job
(same server config, separate nodes) passed gsm8k at em_strict 0.959, so this
is not a config or engine problem.

## Fix

`submit.sh` set the SLURM `--job-name` to `$RUNNER_NAME` (e.g. `gb200-nv_1`),
which is stable across runs. The workflow "Slurm cleanup" steps run
`scancel --user=$USER --name=<runner.name>` at pre-run and post-run, so a
bare runner-name allocation can be cancelled by any concurrent/adjacent job on
the shared cluster+user that shares the runner name. This change appends
`GITHUB_RUN_ID` and `GITHUB_RUN_ATTEMPT` to the job name so an exact
`--name=<runner.name>` scancel from elsewhere no longer matches this
allocation.

The sweep is isolated to the single max-tpt config-key. If max-tpt now passes
in-sweep, the name-scoped scancel was the cause; if it still fails, this rules
it out and points to an environmental/scheduling cause.


ezrasilvera:
Diagnostic complete. The isolated max-tpt re-test with the unique SLURM job-name fix passed (run 28918650203, 8730 tok/s/GPU), confirming the in-sweep failure was a name-scoped scancel, not the config. Folding the fix and restoring the full curve (mid-curve + high-tpt) into llm-d-initial (PR #2050); closing this diagnostic PR.

PR #2178, opened 2026-07-13
fix(agentic): standardize replay warmup

cquil11:
Agentic replay warmup was optional and configured in individual recipes.

- Always run the normal trajectory warmup followed by a fixed 10-minute accelerated warmup.
- Remove recipe-specific duration overrides.

Tests:
- `bash -n benchmarks/benchmark_lib.sh benchmarks/single_node/agentic/*.sh benchmarks/multi_node/agentic_srt.sh`
- `python -m pytest utils/matrix_logic/ -v` (198 passed)


cquil11:
@xinli-sw

PR #2255, opened 2026-07-17
[AMD][dsv4] fix 2P1D eval 503 "all circuits open" (readiness canary + faster CB recovery)

inkcherry:
## Problem

On the 2P1D dsv4 gsm8k eval at concurrency 500, every request gets `503 "No available prefill workers (all circuits open or unhealthy)"` from the first request and never recovers, so the eval produces no results.

`server_sglang.sh` declares the router ready as soon as `/readiness` returns 200, which only means the router process is up — not that it can actually reach a prefill worker and finish a generation. The eval fires 500 concurrent requests into a prefill path that isn't serving yet, trips both prefill circuit breakers, and the default breaker timeout (60s) is longer than the client retry budget (~31s), so the circuits never reclose.

## Changed

- Gate "ready" on one real generation through the router (`POST /v1/chat/completions`, `max_tokens=1`) instead of just `/readiness`.
- Shorten the router circuit-breaker recovery window (`--cb-timeout-duration-secs 15 --retry-max-retries 3`). Breaker and health checks are otherwise unchanged; concurrency and the eval itself are untouched.

## Validated

conc-500 and conc-512 gsm8k both complete, `exact_match` ≈ 0.958, no 503s, circuits stay closed.
cc @Duyi-Wang @billishyahao

### DIFF
--- a/benchmarks/multi_node/amd_utils/server_sglang.sh
+++ b/benchmarks/multi_node/amd_utils/server_sglang.sh
@@ -506,12 +506,23 @@ if [ "$NODE_RANK" -eq 0 ]; then
     fi
     echo "Congratulations!!! All prefill and decode servers are up . . ."
 
+    # Circuit-breaker recovery tuning (run 28696443568):
+    # With defaults the per-worker circuit stays OPEN for cb-timeout-duration-secs=60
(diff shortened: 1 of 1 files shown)


functionstackx:

PR #2257, opened 2026-07-17
[AMD][dsv4] fix 2P1D 8k1k straggler (decode/prefill context-length asymmetry)

inkcherry:
In the 2P1D 8k1k run at concurrency 500/512 a few requests hang until the client timeout (p99.9 TTFT ~1800s).

`models.yaml` sets `context_length: 9217` on the prefill role only, and `server_sglang.sh` only passes `--context-length` to prefill, so decode falls back to the model default (~1M). A request longer than 9216 is accepted by decode (allocates KV, waits for the transfer) but rejected by prefill with HTTP 400, so prefill never sends the KV and that decode request waits forever.

## Changed

Also pass `--context-length` to decode (read from `models.yaml`, defaulting to the prefill value) so both sides use the same limit.

## Validated

Before: conc-500 stalled with 33/1000 (rr=1.0) and 4/1000 (rr=0.8) requests hung to the timeout. After: conc-500 and conc-512 complete 100% in ~140s with normal p99.9. gsm8k accuracy unchanged (0.958).

### DIFF
--- a/benchmarks/multi_node/amd_utils/models.yaml
+++ b/benchmarks/multi_node/amd_utils/models.yaml
@@ -395,6 +395,15 @@ DeepSeek-V4-Pro:
     dp:
       max_running_requests: 1024
       cuda_graph_bs_range: "1-128"
+      # Must match prefill.context_length. In PD-disagg the decode admits requests
+      # against its own context length; if left unset it defaults to the model's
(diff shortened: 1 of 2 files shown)


inkcherry:
included in https://github.com/SemiAnalysisAI/InferenceX/pull/2293, closed.

PR #2351, opened 2026-07-27
kimik3-fp4-mi355x-vllm-agentic: add Kimi-K3 MXFP4 MI355X agentic recipe

Oseltamivir:
## Summary

**Day-zero Kimi-K3 MXFP4 MI355X single-node vLLM agentic-coding recipe.** Agentic trace replay only — no fixed-seq-len scenarios.

Kimi-K3 is a 2.8T-param natively-multimodal MoE (896 experts, 16 routed + 2 shared) built on Kimi Delta Attention, gated MLA, and Attention Residuals. It is served through vLLM's dedicated ROCm path (`vllm/models/kimi_k3/amd/*`, selected by `current_platform.is_rocm()`), whose KDA and attn_res triton kernels are gfx950-validated; the CUDA-only FlashKDA extension is gated behind `VLLM_GPU_LANG=CUDA` and is not built for ROCm.

- **Upstream recipe**: flags and env follow [vllm-project/recipes#684](https://github.com/vllm-project/recipes/pull/684) (`models/moonshotai/Kimi-K3.yaml` → `hardware_overrides.amd`, which lists mi355x as verified): `VLLM_ROCM_USE_AITER=1`, `SAFETENSORS_FAST_GPU=1`, `AITER_SITUV2_A8W4=1`, `AITER_BF16_FP8_MOE_BOUND=0`, `VLLM_USE_BREAKABLE_CUDAGRAPH=0` (required on ROCm), plus `--moe-backend auto --mm-encoder-tp-mode data --enable-auto-tool-choice --tool-call-parser kimi_k3 --reasoning-parser kimi_k3`.
- **TP=8 only**: the ~1.56 TB checkpoint needs ~195 GB/GPU of MI355X's 288 GB HBM, so TP=4 (~390 GB/GPU) cannot load.
- **KV cache offloading is OFF** — the sole search-space row is `kv-offloading: none` at tp 8, conc 1/2/4/8, and the scenario-level `dram-utilization` key is dropped since no row consumes it. The SimpleCPUOffloadConnector arm was walked down from `dram-utilization` 0.95 to 0.80 and still destabilized startup: pinned host memory is unswappable, and the workers' per-rank pools starved the shm_broadcast ring (`No available shared memory broadcast block found in 60 seconds`), killing EngineCore during warmup. This entry measures the GPU-resident baseline first; the `vllm-simple`/`vllm-native` branches stay in the bench script so the offload arm can return as a separate row once it is stable.
- **`--gpu-memory-utilization` 0.88**, not the upstream 0.95: MI355X reports 287.98 GiB total but only ~271 GiB free by the time the worker checks, so 0.95 hard-fails every rank before KV sizing. Measured ceiling ~0.94; 0.88 leaves headroom for the ~263 GiB this fleet intermittently strands after a killed job.
- **`--max-num-batched-tokens` is concurrency-dependent**: 4096 for conc ≤ 5, 16384 above. At 16384 the conc 4 row page-faults in `wvSplitK` (vLLM's ROCm skinny GEMM, reached from the KDA `in_proj_qkvgfab` projection); `rocm_unquantized_gemm_impl` only selects that kernel for GEMMs with n ≤ 5 rows, i.e. decode at low concurrency, so 5 is the dispatch gate rather than a guess.
- **Full 1M context**: `--max-model-len 1048576`, and `kimik3*` is added to `resolve_trace_source`'s unfiltered-corpus allowlist so K3 replays the full unfiltered 062126 v7 corpus instead of the 256k-capped variant.
- **Weights stage to node-local NVMe** (`/var/lib`, ~6.4 TB free) rather than the `/it-share` NFS hub cache, which avoids the share contention that stretched weight loads to ~25 min when several nodes pulled concurrently.

⚠️ **Note for review**: this branch also drops `scenario_default="swebench"` from `benchmarks/benchmark_lib.sh`, a shared-library change affecting every agentic config. It makes `utils/evals/test_run_eval_dispatch.py::test_agentic_scenario_defaults_to_swebench` fail, since that test asserts the old default.

## 中文说明

**Kimi-K3 MXFP4 MI355X 单节点 vLLM 智能体编码配方（day-zero）**，仅智能体轨迹回放，不含定长序列场景。

Kimi-K3 为 2.8T 参数原生多模态 MoE（896 专家，16 路由 + 2 共享），基于 Kimi Delta Attention、gated MLA 与 Attention Residuals；经 vLLM 的 ROCm 专用路径（`vllm/models/kimi_k3/amd/*`）服务，其 KDA 与 attn_res triton 内核已在 gfx950 验证，CUDA-only 的 FlashKDA 扩展在 ROCm 上不构建。

参数与环境变量遵循上游配方 [vllm-project/recipes#684](https://github.com/vllm-project/recipes/pull/684) 的 AMD 覆盖块。仅 TP=8：约 1.56 TB 检查点在 288 GB HBM 上需约 195 GB/GPU，TP=4 无法加载。

**KV 缓存卸载已关闭**：唯一搜索空间行为 `kv-offloading: none`（tp 8，并发 1/2/4/8），场景级 `dram-utilization` 键一并移除（已无行使用）。SimpleCPUOffloadConnector 分支从 0.95 降至 0.80 后仍导致启动不稳定——固定内存不可换出，各 rank 内存池会耗尽 shm_broadcast 环形缓冲（"No available shared memory broadcast block found in 60 seconds"）并在预热期杀死 EngineCore。本条目先测 GPU 常驻基线；脚本中的 `vllm-simple`/`vllm-native` 分支保留，待稳定后可作为独立行恢复。

其余：`--gpu-memory-utilization` 0.88（上游 0.95 在每个 rank 上硬失败）；`--max-num-batched-tokens` 随并发取值（并发 ≤5 用 4096，否则 16384，阈值 5 即 `wvSplitK` 的派发门限）；`--max-model-len 1048576` 配合 `kimik3*` 加入非过滤语料白名单；权重暂存于节点本地 NVMe 而非 NFS 共享缓存。

⚠️ **评审注意**：本分支同时从 `benchmarks/benchmark_lib.sh` 移除了 `scenario_default="swebench"`，属影响所有智能体配置的共享库改动，会导致 `utils/evals/test_run_eval_dispatch.py::test_agentic_scenario_defaults_to_swebench` 失败（该测试断言旧默认值）。


functionstackx:
non spec decode wont be on parity

PR #2367, opened 2026-07-28
[Klaud Cold] kimik3-fp4-mi355x-vllm-agentic-mtp: Kimi-K3 DSpark MI355X agentic recipe at golden AL 3.84

Oseltamivir:
Adds the DSpark speculative-decoding variant of the Kimi-K3 MXFP4 MI355X agentic
recipe: `kimik3-fp4-mi355x-vllm-agentic-mtp`.

4 points, TP8, DRAM KV offload via `vllm-simple`, conc 1/4/8/16 (mirrors the
non-MTP ladder for direct comparison), plus one eval point.

## Acceptance follows the committed golden AL

Per `docs/PR_REVIEW_CHECKLIST.md` rule 10, throughput points **simulate**
acceptance rather than measuring it, pinned to the committed curve in
`golden_al_distribution/kimik3_dspark.yaml` (`kimi-k3` / `thinking_on` /
`num_speculative_tokens=7` → **AL 3.78**):

```
throughput : "rejection_sample_method": "synthetic", "synthetic_acceptance_length": 3.78
EVAL_ONLY  : "rejection_sample_method": "block"   (real target verification)
```

The eval point deliberately uses real verification — synthetic acceptance
bypasses verification and would zero the eval score.

The recipe also **hard-fails** if the pinned AL drifts from the committed file or
cannot be parsed, so a recollected curve cannot silently leave a stale constant
behind.

## Deviations from the upstream recipe, and why

| Upstream | Here | Reason |
|---|---|---|
| `"attention_backend": "FLASHINFER_MLA"` | dropped | flashinfer is absent from the ROCm image; `platforms/rocm.py` rejects it (`Reason: ['ImportError']`). Verified by A/B on 8×MI355X: with the pin the draft fails at construction (`dspark_mla.py:59` → `nvidia/mla.py:267 get_attn_backend`), without it the server reaches ready and generates. Server runs `TRITON_ATTN`, as the MiniMax-M3 ROCm MTP recipe does. |
| `--gpu-memory-utilization 0.95` | **0.90** | Measured on an idle node: at the worker's startup check free memory is 271.1–271.95 GiB but 0.95 requires 273.59 GiB, so it fails every rank. Practical ceiling ≈0.94. |
| `--max-num-seqs 32` | `min(CONC, 32)` | The harness sweeps concurrency; a fixed 32 would flatten the ladder into one operating point. |
| `"lazy_offload": "false"` | JSON boolean `false` | `bool("false")` is `True` in Python, so the string silently selects *lazy*. The boolean gives the intended eager offload. |
| — | `--enable-prefix-caching` | Required: without it vLLM asserts `tokens_per_block=1048576 not divisible by tokens_per_hash=3145728. Hybrid models (e.g. Mamba+Attention) need --enable-prefix-caching`. K3 is hybrid (69 KDA + 24 gated MLA). |

## Notes

- On ROCm the draft runs the **NVIDIA** implementation: there is no
  `vllm/models/kimi_k3/amd/dspark*.py` and the registry hardcodes
  `vllm.models.kimi_k3.nvidia.dspark_mla` with no platform dispatch. The base
  model still resolves to `kimi_k3.amd.model`.
- Draft weights (`Inferact/Kimi-K3-DSpark`, 7.1 GB) are pre-staged on the shared
  NFS HF hub cache alongside the 1.5 TB target.
- Verified on real weights that K3+DSpark serves and generates coherently on
  gfx950 before this was written.

## Not yet validated

- Eager DRAM offload at `dram-utilization: 0.95` leaves only ~12–21 GB host free
  once the pool commits (measured). Lowering it is a one-line change if the
  sweep shows host pressure.
- A `TimeoutError: RPC call to sample_tokens timed out.` was seen on a TEP8 row
  in the sibling non-MTP sweep, which shares this eager-offload config. If the
  same hang appears here, the offload memory setting is the first thing to look at.


functionstackx:
seperceded by seungrok pr

PR #2407, opened 2026-07-29
CollectiveX: one nccl-ep handle per group, and restore the low-latency rows

Oseltamivir:
## Why

The nccl-ep adapter created one handle per problem shape and cached it on the problem, so walking a token ladder left up to nine handles live on a single EP group, interleaved by the randomised trial order.

That is unsafe. `buffer_idx` — the LL double-buffer parity selector — is per-**handle** state, but the buffers it selects are offsets into the per-**group** `rdma_buffer`. Two handles built from the same group config resolve to the *same* parity-0/parity-1 count and flag slots (which themselves share one offset) and advance their parity independently, so one handle's "next buffer, safe to clean" is the other's "current buffer, in flight". Cleans land on live signals and polls observe a sibling handle's leftovers.

This is the real cause of the low-latency wedge we had attributed to the wheel. Isolated on a **stock** `nccl4py[cu13]==0.3.1` wheel, GB200, 4 ranks on one tray, with the handle count as the only variable:

| decode ladder | handles | result |
|---|---|---|
| `[1]` | 1 | completes, **zero** receive timeouts, correctness passing |
| `[1, 2]` | 2 | **64 dispatch + 6 combine** receive timeouts → `cudaErrorLaunchFailure` |

NVIDIA's own `ep_bench` never reproduces it across eight configurations (per-iteration `MPI_Barrier` removed, an injected 500 ms rank stall, dispatch-only and combine-only loops, up to 20 000 iterations) purely because it uses a single handle. The upstream report has been corrected accordingly — NVIDIA/nccl#2303.

It also corrupts signalling *without* hanging, so latencies from any multi-handle run were suspect, not only the runs that wedged.

## What changed

Rebinding is the intended lifecycle, not a workaround: `ncclEpInitHandle` takes no token count, and `ncclEpUpdateHandle` is documented as a "per-step collective: prepare the handle for the given top-k routing decisions". The wheel exposes it as `Handle.update`, which rebinds top-k indices without reallocating buffers. This also brings nccl-ep in line with the other three backends, each of which allocates one object sized to the ladder maximum and varies the token count per call.

1. **One handle per group** (`bench/ep_nccl.py`). The per-problem namespace stays — it holds only host-side tensor descriptors and, for HT, that problem's counters — while the handle moves to the group and is rebound when the bound shape changes. Both create and update are collective, so they must run in the same order on every rank; they only run on a shape change, and since every timed component is preceded by an untimed `warm()` on its own problem, the collective always lands in `warm` or the oracle passes, never inside a timed window. Re-entering the already-bound problem returns with no collective and no sync, so a timed loop's steady state is unchanged. HT re-reads its received-token count on rebind because the metadata exchange recomputes it for that routing; the value is deterministic per problem. Teardown is a single explicit `Handle.destroy`.

2. **Low-latency rows restored on all six NVIDIA SKUs** (`configs/platform_config.json`). These were dropped in d58d56566 while every LL leg wedged, on the reading that only a fixed wheel could restore them — that reading was wrong about the cause, and the single-handle adapter removes the aliasing on the stock wheel. `COLLX_NCCL4PY_SPEC` is untouched. LL is decode-only and EP8-only, so the GB SKUs carry an EP8 row even though normal mode runs EP16 there, and b300/gb200/gb300 gain their first `ll_backends` entry. `test_nccl_ep_rollout_shape` previously pinned the *absence* of these rows and now pins their presence, plus the absence of any EP16 LL row.

## CI validation

**High throughput** (h100 EP8, decode + prefill) against a `main` baseline on the same pool — [PR run](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/30436725064) / [baseline](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/30438014051). Both `success`, every rung correct, 2048 samples per point, max relative error unchanged. Round-trip p50 within ±1.3% and decode p99 within ±2.1% — so the per-shape `update` collective does not perturb the steady state. Prefill p99 came in 5.7–9.9% *lower* on all four rungs with p50 flat, which is the signature of removed tail interference; suggestive only, two runs cannot separate it from pool variance.

**Low latency** — [run 30440418355](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/30440418355), all six SKUs green, 9 rungs each (decode T=1…256), every rung `correct=True`, max relative error 0.00389 throughout:

| SKU | transport | nodes | RT p50 T=1 → T=256 |
|---|---|---|---|
| h100-dgxc | nvlink | 1 | 55.1 → 239.8 µs |
| h200-dgxc | nvlink | 1 | 53.4 → 226.2 µs |
| b200-dgxc | nvlink | 1 | 70.0 → 142.2 µs |
| b300 | nvlink | 1 | 60.1 → 133.9 µs |
| gb200 | mnnvl | 2 | 74.5 → 148.4 µs |
| gb300 | mnnvl | 2 | 67.1 → 141.5 µs |

The two GB rows are EP8 across two trays over MNNVL — the configuration from NVIDIA/nccl#2303, which had never completed on any wheel — and the four x86 SKUs previously wedged 5/5 over SSH and 4/4 in sweep 30155842613.

Unit tests: `python3 -m unittest discover -s tests` — 64 passed, 6 skipped (torch-dependent).

## Notes

`tests/test_ep_nccl_handle.py` stubs torch and nccl so it runs without the benchmark image, and locks the contract: one `create_handle` across nine rungs, `update` on a shape switch, **no** collective when the bound shape is re-entered, no `layout_info` on an LL rebind, HT rebinds carrying that problem's own counters, and idempotent teardown. The stubs are withdrawn after the import so the genuinely torch-dependent modules still skip rather than error.

The upstream patch (NVIDIA/nccl#2306) is now a defence-in-depth measure rather than the thing that unblocks us: it makes a foreign or stale signal inert, which is worth having, but this tree no longer depends on it.

### DIFF
--- a/experimental/CollectiveX/bench/ep_nccl.py
+++ b/experimental/CollectiveX/bench/ep_nccl.py
@@ -114,6 +114,9 @@ class NCCLEPBackend(EPBackend):
         self._combine_cfg = CombineConfig(send_only=0)
         self._comm = None
         self._ep_group = None
+        # Exactly ONE handle per group, rebound per problem shape — see _ensure_handle.
+        self._handle = None
+        self._bound = None
 
     def buffer_cap(self, args):
         if self._ll:
@@ -238,15 +241,34 @@ class NCCLEPBackend(EPBackend):
             self._ht_disp_counts_t = self._t(self._ht_disp_counts)
 
     def _ensure_handle(self, p):
-        """Create (once, cached on the problem) the reusable per-step handle for p's routing.
+        """Bind the group's single handle to p's routing, creating it on first use.
 
-        create_handle is collective and, in HT, performs the metadata exchange that fixes the
-        received-token count — so it must run in the same order on every rank. It first runs
-        inside Pass 1's untimed warm (ladder order, identical across ranks); the timed passes
-        then reuse the cached handle and never enter a collective here.
+        ONE handle per group, rebound per shape — never one handle per shape. `buffer_idx`, the
+        LL double-buffer parity selector, is per-HANDLE state, but the buffers it selects are
+        offsets into the per-GROUP rdma_buffer: two handles built from the same group config
+        resolve to the SAME parity-0/parity-1 count+flag slots and advance their parity
+        independently, so one handle's "next buffer, safe to clean" is the other's "current
+        buffer, in flight". Interleaving handles therefore corrupts the signalling even when it
+        does not hang outright, which makes every latency drawn from such a run suspect. Filed
+        upstream as NVIDIA/nccl#2303; reproduced on a stock wheel by ladder [1] (one handle,
+        clean) vs ladder [1, 2] (two handles, 64 dispatch + 6 combine receive timeouts -> 719).
+
+        `ncclEpInitHandle` takes no token count and `ncclEpUpdateHandle` is documented as a
+        "per-step collective: prepare the handle for the given top-k routing decisions", so
+        rebinding IS the intended lifecycle. This also matches the other three backends, which
+        each allocate one object sized to the ladder maximum and vary the token count per call.
+
+        Both create_handle and update are collective, and HT additionally performs the metadata
+        exchange that fixes the received-token count, so they must run in the same order on
+        every rank. They only ever run on a shape CHANGE, and every timed component is preceded
+        by an untimed warm() on its own problem — so the collective always lands in warm (or in
+        the oracle passes), never inside a timed window. Re-entering with the already-bound
+        problem returns immediately without a collective or a sync.
         """
         cached = getattr(p, "_nccl", None)
         if cached is not None:
+            if self._bound is not cached:
+                self._rebind(cached)
             return cached
         stream = self._stream()
         topk_idx_t = self._t(p.topk_idx)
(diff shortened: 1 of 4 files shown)


Oseltamivir:
## CI validation (HT, h100 EP8 single-node)

Both runs green; artifacts checked rather than just the status.

- this branch — [run 30436725064](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/30436725064)
- `main` baseline (multi-handle), same shard and pool — [run 30438014051](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/30438014051)

`status=success` on all four cases, every rung `correct=True`, 2048 samples per point. Max relative error 0.00775 (decode) and 0.02441 (prefill), unchanged from baseline. With 14 shapes × 256 trials the single handle is rebound ~3.5k times per case against one `create_handle`, and the HT path exercises `update` carrying each problem's own `layout_info` plus the re-read received-token count that combine's row count depends on.

Round-trip latency, `main` vs this branch:

| phase | p50 | p99 |
|---|---|---|
| decode (T=1…512) | −1.3% … +0.2% | −1.8% … +2.1% |
| prefill (T=1024…8192) | −0.3% … +1.3% | **−9.9% … −5.7%** |

Decode is flat within noise both ways, so the per-shape `update` collective does not perturb the steady state — it lands in the untimed `warm()`, as intended.

The prefill p99 improvement is consistent across all four rungs while p50 is flat, which is the signature of removed tail interference rather than a faster path: with one handle there are no sibling handles racing on the shared parity count/flag slots. That is suggestive, not established — two runs cannot separate it from pool variance, and I would not quote it as a result without repeats.

Not covered: LL. `main` carries no `nccl-ep` `ll_backends` for any SKU, so CI cannot generate an LL leg; the LL branch of this change is still unexercised on hardware.

PR #2440, opened 2026-07-31
feat: schedule Slurm jobs with weighted runner leases

cquil11:
## Summary

This proof of concept adds exact weighted GitHub-side admission for Slurm allocations.

- Generates `node-count` for every multi-node matrix row from an explicit aggregate `num-nodes`, checked-in recipe resources, explicit `PREFILL_NODES`/`DECODE_NODES`, or worker GPU topology.
- Allows `num-nodes` only for `multinode: true, disagg: false`; disaggregated entries reject it because their separate prefill/decode allocations determine the total.
- Adds `nodes:N` to queued jobs when both priority scheduling and `NODE_SLOT_SCHEDULER_ENABLED=true`; single-node jobs request `nodes:1`.
- Uses `nodes:N` as request metadata rather than a permanent runner capability.
- Uses node count as a secondary priority tie-breaker so otherwise-equal smaller allocations run first.
- Keeps the feature disabled by default.

Runtime allocation is implemented alongside the existing priority controller in the companion draft PR: [inferencex-dash#1](https://github.com/SemiAnalysisAI/inferencex-dash/pull/1).

The existing aggregate master-config entries are backfilled in the native stacked PR above this one: #2479.

## Lease lifecycle

For a job requesting `nodes:3`, the controller:

1. Finds three online, unleased runner tokens in one compatible `cluster:*` pool.
2. Adds one shared `ci-lease-*` label to all three runners.
3. Re-reads and verifies the complete lease.
4. Adds the queued job unique `ci-job-*`, `ci-attempt-*`, and `nodes:3` labels to one anchor runner.
5. Preserves the lease while the anchor is queued or busy.
6. Removes the lease after completion or during stale-state reconciliation.

The unique `ci-job-*` label remains the dispatch gate. The other leased runners stay idle as abstract capacity tokens for the Slurm allocation.

On an 18-node pool, a running 10-node lease leaves eight tokens. A queued nine-node job remains blocked, while an eight-node job can backfill the exact remainder. Therefore managed work satisfies:

```text
sum(admitted node counts) <= online cluster capacity
```

Slurm remains the final authority for external users, offline compute nodes, reservations, and jobs submitted outside this managed path.

## Activation

1. Merge and deploy [inferencex-dash#1](https://github.com/SemiAnalysisAI/inferencex-dash/pull/1).
2. Verify managed runners have exactly one correct `cluster:*` label. Unclustered legacy runners remain eligible only for one-node jobs.
3. Drain previously queued/running multi-node jobs so no pre-feature job is accounted as one node.
4. Set `NODE_SLOT_SCHEDULER_ENABLED=true`. `PRIORITY_SCHEDULER_ENABLED` must also remain enabled.

No permanent `nodes:N` label plan is required. The controller removes old static `nodes:*` labels as it reconciles idle runners.

## Validation

- `python -m pytest utils/matrix_logic/ utils/test_ci_priority.py -q` — 253 passed.
- Generated all NVIDIA and AMD master-config rows; every multi-node row has `node-count`.
- Compared 216 generated rows across 185 checked-in srt-slurm recipes with authoritative resource totals; zero mismatches.
- `actionlint` found no new workflow-call or node-lease expression errors; existing unrelated shell/style warnings remain.
- Companion controller: `bun run check` passes with 100 tests, including 18-node `10 + 9`, exact `10 + 8` backfill, running-lease preservation, invariant coverage for pool sizes 1–20, legacy-runner fallback, two-phase publication, and partial-write rollback.
- Companion controller production build passes.

### DIFF
--- a/.github/workflows/benchmark-multinode-tmpl.yml
+++ b/.github/workflows/benchmark-multinode-tmpl.yml
@@ -6,6 +6,10 @@ on:
       runner:
         required: true
         type: string
+      node-count:
+        description: "Total Slurm nodes required by this benchmark"
+        required: true
+        type: string
       priority:
         description: "Higher-is-sooner CI priority score"
         required: true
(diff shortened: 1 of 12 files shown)


functionstackx:

PR #2451, opened 2026-08-02
[Klaud Cold] chore(evals): disable agentic evals repo-wide behind AGENTIC_EVALS_DISABLED / 全仓库停用智能体评测（AGENTIC_EVALS_DISABLED 开关）

functionstackx:
Turns agentic (SWE-bench) evals off repo-wide behind a single flag, until the agentic eval path is fixed.

## @adibarra — TODO

**Re-enable is one line:** flip `AGENTIC_EVALS_DISABLED` to `False` in `utils/matrix_logic/generate_sweep_configs.py` (or delete it and the two `if AGENTIC_EVALS_DISABLED` guards). Nothing else has to change. The `TODO(@adibarra)` marker is on the flag itself and in `utils/evals/EVALS.md`.

The existing selection policy keeps its test coverage behind the new `_agentic_evals_enabled()` helper in `test_generate_sweep_configs.py`, so the behaviour you're turning back on is already asserted — you should see those tests go green against the real code path the moment you flip it.

## Why one flag is enough

Both dispatch paths key off the same thing — an `agentic-coding` row carrying `run-eval: True`:

- `run-sweep.yml` → `sweep-agentic-evals` reads `search-space-config.agentic_evals`, which `process_changelog.py` fills by splitting eval rows on `scenario-type == 'agentic-coding'`.
- `e2e-tests.yml` → `test-sweep-agentic-evals` filters `get-jobs` output on the same flag.

So suppressing the mark in `generate_sweep_configs.py` leaves both jobs with an empty matrix, and both skip on their existing `!= '[]'` guards. No workflow YAML changes.

Worth knowing: `process_changelog.py` runs a **second `--evals-only` pass for every changelog entry** (`utils/process_changelog.py`, the `eval_groups` loop), and `--evals-only` sets `include_agentic=True`. That is why agentic eval jobs were dispatching on every agentic PR regardless of labels — e.g. #2447 and #2448 each got one without asking. The flag wins over `--evals-only`, `--all-evals`, and the `evals-only` / `all-evals` PR labels alike.

## Blast radius

Measured against `configs/nvidia-master.yaml`, before vs after:

| | before | after |
|---|---|---|
| agentic eval rows, `glm5.2-fp4-b300-sglang-agentic --evals-only` | 1 | **0** |
| fixed-seq-len eval rows, `dsv4-fp4-b200-vllm --evals-only` | 4 | 4 |
| default throughput matrix, same key | 7 rows, 0 evals | identical |

Agentic **throughput** coverage is untouched, and `gsm8k` / `gpqa` / 8k1k `swebench` are untouched. Only the agentic eval job goes away.

## Changes

| File | Change |
|---|---|
| `utils/matrix_logic/generate_sweep_configs.py` | `AGENTIC_EVALS_DISABLED = True` + two guards: the `include_agentic` block in `mark_eval_entries()` and the unconditional agentic marking in `mark_all_eval_entries()`. |
| `utils/matrix_logic/test_generate_sweep_configs.py` | `_agentic_evals_enabled()` context manager so the two existing policy tests keep testing the policy; two new tests pin that the switch beats `--evals-only` and `--all-evals`. |
| `utils/evals/EVALS.md` | Status section at the top saying what is off, what is unaffected, and how to turn it back on. |

No config, no `perf-changelog.yaml` entry, no GPU jobs — `[skip-sweep]` on the head commit.

## Validation

`pytest utils/` — 570 passed, 104 of them in `test_generate_sweep_configs.py`. Two failures in `utils/evals/test_run_eval_dispatch.py` (`test_agentic_generation_invokes_mini_swe_agent`, `test_agent_sandbox_cpu_knob`) reproduce identically on a clean `main` in this environment — a missing local `mini-swe-agent` fixture, not this change.

### DIFF
--- a/utils/evals/EVALS.md
+++ b/utils/evals/EVALS.md
@@ -5,6 +5,25 @@ concurrency, kernels, and other throughput optimizations. They run separately
 from throughput; selection lives in `mark_eval_entries()` in
 `utils/matrix_logic/generate_sweep_configs.py`.
 
+## Status: agentic evals are disabled
+
+**Agentic (SWE-bench) evals are turned off repo-wide** by
+`AGENTIC_EVALS_DISABLED` in `utils/matrix_logic/generate_sweep_configs.py`.
+While it is set, no `agentic-coding` row is ever marked `run-eval`, so
+`run-sweep.yml`'s `sweep-agentic-evals` and `e2e-tests.yml`'s
+`test-sweep-agentic-evals` both find an empty matrix and skip — including
+under `--evals-only`, `--all-evals`, and the `evals-only` / `all-evals` PR
+labels. Agentic *throughput* coverage is unaffected, and fixed-sequence
+(`gsm8k`, `gpqa`, `swebench` on 8k1k) evals are unaffected.
+
+TODO(@adibarra): fix the agentic eval path and re-enable by flipping the flag
+to `False`. The selection policy below still has test coverage behind the
+`_agentic_evals_enabled()` helper in `test_generate_sweep_configs.py`, so the
+re-enable is a one-line change.
+
+The rest of this section describes the behaviour that returns when the flag is
+cleared.
+
 ## Selection
 
 - **Single-node:** 8k1k only; highest and median concurrency for every model,
--- a/utils/matrix_logic/generate_sweep_configs.py
+++ b/utils/matrix_logic/generate_sweep_configs.py
@@ -235,6 +235,21 @@ def _multinode_parallelism_key(entry: dict) -> tuple:
     ))
 
 
+# TEMPORARY: agentic (SWE-bench) evals are disabled repo-wide.
+#
+# TODO(@adibarra): fix the agentic eval path and re-enable by flipping this to
+# False (or deleting it and the two `if AGENTIC_EVALS_DISABLED` guards below).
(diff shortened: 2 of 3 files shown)


functionstackx:
@adibarra — **TODO for when you're back next week: fix the agentic eval path and re-enable this.**

This is a temporary repo-wide disable, not a removal. To turn agentic (SWE-bench) evals back on, flip `AGENTIC_EVALS_DISABLED` to `False` in `utils/matrix_logic/generate_sweep_configs.py` — that is the whole re-enable. The `TODO(@adibarra)` marker sits on the flag and in `utils/evals/EVALS.md`.

The selection policy you're re-enabling still has test coverage: the two original tests now run behind the `_agentic_evals_enabled()` helper in `test_generate_sweep_configs.py`, so they exercise the real path the moment the flag clears. Two new tests pin the disabled state, and those are the ones to delete when you flip it.

One thing you may not have known and that is worth fixing at the same time: `process_changelog.py` runs a second `--evals-only` pass for every changelog entry, and `--evals-only` sets `include_agentic=True`. That is why agentic eval jobs were dispatching on every agentic PR regardless of labels — #2447 and #2448 each got one unasked. If the intent was for agentic evals to be opt-in, that path is where the opt-in leaks.

Assigning you and requesting review so this does not get lost.

PR #2457, opened 2026-08-03
fix(agentx): adopt AIPerf v1 timing fixes

cquil11:
## Summary

- Pin `utils/aiperf` to [`818c3a5a`](https://github.com/SemiAnalysisAI/aiperf/commit/818c3a5a2922c535af6271ff296ed374e292b8e4), the merged AgentX v1 timing and join fixes.
- Enable a 300-second observed runtime idle cap for every complete AgentX trajectory tree, including root and subagent streams.
- Preserve request order and spawn/join barriers when the watchdog advances pending timers.
- Allow `AIPERF_TRACE_IDLE_GAP_CAP_SECONDS` overrides and forward them to AMD sibling clients.

This moves the behavior validated by the Qwen B200 and B300 sweeps into the shared AgentX harness. Spread phase starts remain enabled; InferenceX does not pass `--burst-phase-starts`.

## Validation

- `python -m pytest utils/matrix_logic/ utils/test_process_changelog.py -q` — 240 passed
- Bash syntax checks for `benchmark_lib.sh` and the AMD sibling-client launcher
- Normal, fast-mode, and overridden replay-command construction checks
- Confirmed the pinned AIPerf CLI exposes the trace-idle, warmup-per-lane, and phase-start options


xinli-sw (commented):
LGTM! Thanks for the fix

Would be great to make sure that results that are submitted via this fix is not able to be compared with results submitted via previous client (similar to mtp/cross FW comparison that's currently in place, thanks!!)

PR #2479, opened 2026-08-03
config: make aggregate multinode topology explicit

cquil11:
## Summary

- Gives aggregate multi-node master configs one `worker` topology instead of fake `prefill` and zero-count `decode` roles.
- Requires `num-nodes` for every `multinode: true, disagg: false` search-space entry.
- Requires separate `prefill` and `decode` roles for `disagg: true`, where `worker` and `num-nodes` are rejected.
- Backfills all six existing aggregate search-space entries with explicit allocation sizes.
- Translates the aggregate master shape into the legacy prefill/decode matrix representation expected by the launcher, without changing runtime behavior.

This is the upper layer above #2440 in the native GitHub Stacked PR stack.

## Example

```yaml
multinode: true
disagg: false
scenarios:
  agentic-coding:
  - search-space:
    - num-nodes: 2
      worker:
        tp: 8
        pp: 2
        ep: 1
        dp-attn: false
```

## Validation

- `python -m pytest utils/matrix_logic/ utils/test_ci_priority.py -q` — 257 passed.
- Generated every NVIDIA and AMD master-config row successfully.
- Verified all migrated aggregate rows retain their previous node counts, topology, and legacy runtime worker counts.
- `git diff --check` passed.

### DIFF
--- a/configs/nvidia-master.yaml
+++ b/configs/nvidia-master.yaml
@@ -7104,37 +7104,27 @@ qwen3.5-fp4-gb300-dynamo-sglang-agentic-agg:
     - dram-utilization: 0.80
       search-space:
       - spec-decoding: "mtp"
+        num-nodes: 1
         conc-list: [1]
         kv-offloading: dram
         kv-offload-backend: { name: hicache }
-        prefill:
-          num-worker: 1
+        worker:
           tp: 4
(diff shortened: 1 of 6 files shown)


functionstackx:

PR #2501, opened 2026-08-05
Document AgentX engine, draft-model, and offloading policy

cquil11:
## Summary

- Follow up on #2500 with an explicit AgentX model-to-framework policy based on feedback from Tier 1 AI labs and the broader ML community.
- Require providers that support native/upstream vLLM and SGLang as first-class engines to submit the mapped framework first, while allowing multiple additional non-vLLM/SGLang engines afterward.
- Define engine-ordering exceptions for initial enablement of brand-new hardware SKUs and fundamental framework support gaps for new model architectures.
- Explain the balanced vLLM/SGLang assignment principle so both frameworks receive equal coverage without favoritism.
- Record the agreed draft-model plan of record and separate the proposed DSV4 DSpark option pending partner alignment.
- Define the initial AgentX KV cache offloading policy: optional CPU DRAM offloading only, proportional memory allocation, SKU-specific baselines, a 3 TB cap for non-standardized servers, DDR5-aware TCO normalization, and a post-v3 path for NVMe offloading.
- Mirror the policies in MODELS_zh.md.

## Why

AI labs and ML community members want first-class vLLM and SGLang performance in AgentX because proprietary and hardware-specific engines do not always expose every required feature. The framework mapping establishes predictable submission order while retaining practical exceptions. The KV cache policy similarly limits initial scope while preventing unrealistic CPU-memory configurations from distorting performance or TCO comparisons.

## Impact

Documentation only. Contributors can identify the required first-class framework, understand engine-ordering exceptions, distinguish agreed draft models from proposals awaiting alignment, and apply consistent KV cache offloading capacity and cost rules.

## Checks

- git diff --check
- Cross-checked the agreed draft mapping and offloading rules against the supplied AgentX plan of record

### DIFF
--- a/MODELS.md
+++ b/MODELS.md
@@ -72,6 +72,25 @@ The table also records both the agreed plan-of-record (PoR) draft-model mapping 
 | GLM-5.2 (`glm5.2`) | native/upstream SGLang engine | native MTP | — | Additional non-vLLM/SGLang engines under the ordering guideline and exceptions above |
 | Qwen3.5-397B-A17B (`qwen3.5`) | native/upstream SGLang engine | native MTP | — | Additional non-vLLM/SGLang engines under the ordering guideline and exceptions above |
 
+## KV cache offloading policy
+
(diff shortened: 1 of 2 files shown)


cquil11:
@chunfangamd @xinli-sw Could you review the proposed DSV4 DSpark option in this follow-up? The agreed PoR remains native MTP. The proposal adds deepseek-ai/DeepSeek-V4-Pro-DSpark for AgentX only under the existing synthetic-acceptance methodology, pending partner alignment; single-turn 8k1k remains on the native MTP heads.

PR #2526, opened 2026-08-07
[NV] Add SpeedBench AL collectors for DSV4-Pro-0813 on B300

qiching:
Adds the acceptance-length collectors for deepseek-ai/DeepSeek-V4-Pro-DSpark, following the Kimi-K3 DSpark collector (#2352). DSpark ships the draft inside a separate checkpoint, so unlike K3 there is no external draft head to download and no "model" key in the speculative-config.

dsv4dspark_fp4_b300_vllm.sh is a copy of the DSV4 MTP collector with only the speculative-config changed (method dspark + draft_sample_method); every serve flag is left identical so the DSpark and MTP curves stay comparable, and it already matches the published DSpark recipe.

Benjamin asked for greedy and probabilistic drafting to be characterized on DSV4-Pro. speedbench-al.yml resolves the collector purely as ${model-prefix}_fp4_b300_vllm.sh, so the second arm needs its own dispatchable file; dsv4dsparkprob_fp4_b300_vllm.sh delegates to the base collector with DRAFT_SAMPLE_METHOD=probabilistic rather than duplicating it, so the two arms cannot drift. rejection_sample_method is left at the vLLM default and exposed as an opt-in env knob: the K3 variant bundled it with probabilistic drafting, but the forced-AL config has to stay on a sampling method TRT-LLM also supports.

The golden YAML and its README row follow in a separate PR once the collection run is reviewed.

### DIFF
--- a/benchmarks/single_node/speedbench/dsv4dspark_fp4_b300_vllm.sh
+++ b/benchmarks/single_node/speedbench/dsv4dspark_fp4_b300_vllm.sh
@@ -18,9 +18,12 @@
 # otherwise a copy of so the two AL curves stay directly comparable:
 #   - target model       DeepSeek-V4-Pro-DSpark (was DeepSeek-V4-Pro)
 #   - speculative-config method dspark + draft_sample_method (was method mtp)
-# Every serve flag is left byte-identical to the MTP collector: it already
-# matches the published DSpark recipe, and holding the serving config fixed is
-# what makes "DSpark vs MTP on DSV4-Pro" a like-for-like AL comparison.
+#   - --max-num-seqs and --gpu-memory-utilization pinned (MTP took the defaults)
+# The third one is a memory fix rather than a recipe change. AL is a per-draft
+# accept/reject property that does not depend on batch size or graph capture, so
+# pinning those two still leaves "DSpark vs MTP on DSV4-Pro" like-for-like; every
+# flag that does affect drafting is byte-identical to the MTP collector and to
+# the published DSpark recipe.
 #
 # Usage (inside the vLLM container, on a B300 node):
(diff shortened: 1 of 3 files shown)


functionstackx (commented):
hi @qiching @xinli-sw  thanks for the contribution! any chance it could be on the 0813 ckpt? iirc it comes with native dspark module 

https://huggingface.co/deepseek-ai/DeepSeek-V4-Pro-0813

chunfangamd:
Should we collect the native AL distribution on AgentX as a validation point and compare it against the SPEED-Bench-derived AL? We are benchmarking on the AgentX dataset anyway.

chunfangamd:
I think @functionstackx's comment that we should move this to the 0813 checkpoint is valid. 

DeepSeek describes 0813 as the official release superseding the preview, built on the same model structure and with DSpark attached. Considering InferenceX’s policy of tracking the latest model within the same model/framework family, 0813 should become the benchmark target if framework support remains unchanged. When we shift to [DeepSeek-V4-Pro-0813](https://huggingface.co/deepseek-ai/DeepSeek-V4-Pro-0813), we will need to update the AL distribution. Should we do it now？

qiching:
> hi @qiching @xinli-sw thanks for the contribution! any chance it could be on the 0813 ckpt? iirc it comes with native dspark module
> 
> https://huggingface.co/deepseek-ai/DeepSeek-V4-Pro-0813

This PR can include -0813. Now I will update this PR since DeepSeek-V4-Pro-0813 already be staged.

qiching:
@functionstackx Now could we merge?

cquil11 (approved):

PR #2537, opened 2026-08-09
fix(evals): preserve GPQA reasoning responses / 修复 GPQA 推理响应截断

JordanNanos:
Summary:
- Add explicit EOS stop tokens to the GPQA generation recipe.
- Prevent lm-eval from applying its implicit blank-line stop to reasoning responses.
- Add a regression test for the recipe contract.

Observed failure:
A live DeepSeek V4 Flash qualification produced 250 invalid extractions out of 396 because generation stopped at the first blank line. The resulting 9.34% score is invalid evidence and will be rerun after this fix.

Validation:
- New focused regression test: passed.
- Full test_run_eval_dispatch.py: 33 passed; 2 unrelated local failures because subprocess Python lacks PyYAML in this macOS environment.
- git diff --check: passed.

中文摘要：
- 为 GPQA 生成任务显式添加 EOS 停止标记。
- 防止 lm-eval 使用空行作为默认停止条件而截断推理响应。
- 添加任务配方回归测试。

### DIFF
--- a/utils/evals/gpqa_diamond.yaml
+++ b/utils/evals/gpqa_diamond.yaml
@@ -21,6 +21,16 @@ doc_to_text: |
   Express your final answer as the corresponding option 'A', 'B', 'C', or 'D'.
 doc_to_target: answer
 
+# lm-eval defaults generate_until tasks without explicit kwargs to stopping at
+# a blank line. Reasoning models commonly use blank lines before their final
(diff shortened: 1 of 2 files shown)


functionstackx:
we dont use gpqa stale PR & probably superceded by other PRs. please reopen when ready

PR #2542, opened 2026-08-10
feat(amd): orchestrate AMD clusters with srt-slurm

cquil11:
## Summary

Shared base for the MI355X AMD ports. Rebased onto InferenceX `main` at `6c5eff0b8b699530a49c9f2c654cd03219914d01` and updated for current srt-slurm v2.

- Keep one `launch_mi355x-amds.sh` entrypoint. srt-slurm owns multi-node allocation, workers, native routers, services, logs, and completion; preserve the latest upstream single-node path.
- Put ROCm visibility, fabric, Slurm directives, mounts, and host prerequisites in `runners/srt-slurm/mi355x-amds.yaml`, using the merged cluster renderer.
- Consolidate workflow input/artifact adaptation here, using v2 `engine` / `roles`. Submit through shared `apply_srt_recipe` so speculative AgentX acceptance comes from the measured golden curves.
- Select the synchronized AMD/MoRI/ATOM/observation fork stack at immutable SHA `c3a5cecd5d4513aa69444323cd6bce668dcbdc88`, based on NVIDIA `712e174a9a64d4140cbeb3521b53a2c8ba7b8c81`. Other pools' runtime selection and the NVIDIA submodule pin are unchanged.
- Preserve cluster fabric/GPU-drain/free-hugepage prerequisites through native host setup, and routed evaluation/artifact collection. No engine patches or modifications to `benchmark_serving.py`.
- Retire `amd_utils`, its wrappers, and their legacy config entries. The workload PRs below restore the ported entries; **merge this base together with the applicable children to restore those production lanes**. Keep deprecated workloads archived rather than re-enabling them.
- Include small aggregate and 1P1D Qwen smoke recipes. Fixed-sequence clients consume the matrix's actual `CONC_LIST`, matching current result validation.

## Workload PRs

- [#2555: DeepSeek V4 AgentX DSPARK](https://github.com/SemiAnalysisAI/InferenceX/pull/2555)
- [#2627: archived DeepSeek V4 ATOM/AToMesh](https://github.com/SemiAnalysisAI/InferenceX/pull/2627)
- [#2628: Qwen3.5 FP8](https://github.com/SemiAnalysisAI/InferenceX/pull/2628)
- [#2629: Qwen3.5 MXFP4](https://github.com/SemiAnalysisAI/InferenceX/pull/2629)
- [#2630: archived DeepSeek V4 SGLang fixed-sequence](https://github.com/SemiAnalysisAI/InferenceX/pull/2630)
[...]

PR #11, opened 2025-09-15
[NVIDIA] Update vLLM version to latest v0.10.2 on H200. 

kedarpotdar-nv:
Updates the 70b_fp8_h200_slurm.sh benchmark script to use a configuration file approach and fixes critical dependency issues that were causing benchmark failures.

PR #2555, opened 2026-08-11
Port MI355X DeepSeek V4 AgentX DSPARK to AMD-enabled srt-slurm

cquil11:
## Scope

Port the now-merged upstream MI355X DeepSeek-V4-Pro-0813 AgentX DSPARK configuration using native srt-slurm v2. The old #3188 baseline is already present in current InferenceX main.

- Preserve the seven-point upstream sweep: TP4 at c4; TP8 at c16 without offload; TP8 with HiCache at c32/c48; TP8 with DP attention and UMBP at c128/c192/c256. Each point still allocates two nodes.
- Use the native cluster `model_paths` alias for the existing `/it-share/data/DeepSeek-V4-Pro-0813` checkpoint, as the production launcher did. The shared Hub cache is mounted at `/hf_hub_cache/hub`; no cache seeding or duplicate model download.
- Preserve the serving image `lmsysorg/sglang-rocm:v0.5.19-rocm720-mi35x-20260913` and separate client image `...-20260907`.
- Preserve DSPARK block size 3, one speculative step, four verification tokens, DP LM-head support, throughput golden AL 3.01, real verification for evaluation, the FP4 indexer, role-specific memory/DP settings, and concurrency-scaled admission and decode graph ranges.
- One native srt-slurm recipe uses base/override variants for the four configurations. HiCache remains prefill-only at ratio 2.5; UMBP remains prefill-only with a 1.5 TB DRAM tier and no intermediate HiCache tier.
- UMBP uses srt-slurm's existing generic service on the prefill node: start before workers, track it as critical, and wait for the GPU-registration log marker. Its ROCm scratch-reclaim and hardware-queue settings match the original prefill launch environment. No engine patches.
- `launch_mi355x-amds.sh` remains the single cluster entrypoint. srt-slurm owns allocation, worker/router/service lifecycle, log streaming, and completion. The adapter in `infx.workflows.srt_slurm` maps CI inputs and collects results.
- Shared legacy removal and launcher/profile/adapter functionality live in #2542. This child adds only the AgentX master entry and recipe; `benchmark_serving.py` is unchanged.

## Orchestration differences

Native SGLang routing and MoRI replace the legacy launch/barrier/process coordination. Model, precision, capacity, MTP, and throughput settings are preserved; startup commands are not byte-identical. srt-slurm assigns its own endpoint/bootstrap ports and handles native-worker startup.

The existing upstream native SGLang backend adds `--skip-server-warmup` for disaggregation, avoiding a local synthetic warmup against a placeholder peer. This startup behavior is not introduced by the AMD stack. Canonical AgentX benchmark warmup is retained, and real P/D serving and evaluations must pass before validation is claimed.

The UMBP socket lives in the allocation-specific shared `/logs` mount so the service and worker containers can both access it. The host-memory capacity guard is preserved. One managed readiness deadline of 5400 seconds replaces three sequential 1800-second startup waits; workers still wait for the final GPU-registration marker, not merely a listening socket. A failed service is managed by srt-slurm.

Caller-owned runtime settings and the separate benchmark image are forwarded into the custom benchmark. Cluster RDMA checks and free-hugepage preparation retain the existing host-setup integration.

Synthetic acceptance is selected by the shared golden-curve connector instead of a hard-coded recipe value.

## Stack

Based directly on [shared AMD integration #2542](https://github.com/SemiAnalysisAI/InferenceX/pull/2542), which is rebased onto InferenceX `main` at `6c5eff0b8b699530a49c9f2c654cd03219914d01`. This child contains only its workload configs, schema-2 recipes, and append-only performance changelog entry. Shared launcher/profile/adapter changes live in #2542.

The base selects SemiAnalysisAI/srt-slurm `c3a5cecd5d4513aa69444323cd6bce668dcbdc88`: #1 → #11 → #7 → #14, synchronized with NVIDIA srt-slurm `712e174a9a64d4140cbeb3521b53a2c8ba7b8c81`. NVIDIA runners retain their existing runtime selection.

## Refresh status

Offline matrix validation: 7 topology/workload rows covering 7 benchmark concurrency points; all 14 prepared throughput/eval configurations pass the pinned runtime's schema/topology checks. The seven AgentX points retain concurrency-scaled admission/graphs and native UMBP service coordination.

No fresh E2E tests or sweeps were started or rerun. Sweep-trigger labels are removed. Historical runtime evidence is listed in the base PR and does not validate these new heads. Full GPU validation remains pending authorization; this refresh makes no new performance or merge-readiness claim.

## AI model disclosure

The exact AI model/version for this refresh is not exposed by the runtime and could not be verified. No subagents contributed to this refresh; earlier model identities are not reliably recorded.


cquil11:
Validation is gated on upstream checkpoint correction #3170. Earlier port Slurm44317 logs independently report uninitialized required draft weights (`model.e_proj.weight`, `model.h_proj.weight`, their scales, `enorm`, and `hnorm`) with the inherited Pro-0813/EAGLE configuration. This is not an srt-slurm launch defect. Run35059494803 was cancelled while all ten GPU jobs were still queued; it provides no GPU validation. Do not change model/acceptance settings independently in this semantic port. After the upstream baseline correction lands, sync the port and launch a fresh complete label-triggered sweep with Run next only.

PR #2583, opened 2026-08-12
kimik3-fp4-mi355x-vllm-agentic-mtp: add LMCache DRAM KV-offload arm

sammshen:
## What

Adds a **dedicated config key** `kimik3-fp4-mi355x-vllm-agentic-mtp-lmcache` with an LMCache DRAM KV-offload arm at TP8 conc **4 / 8 / 16**, on top of the unchanged DSpark MTP serving stack of `kimik3-fp4-mi355x-vllm-agentic-mtp` (same image, script, and topology). A separate key means the changelog selects only the LMCache points — the resident and vllm-simple arms of the base key are not re-run.

## Changes

- **configs/amd-master.yaml** — new key, purely additive (the base key is byte-identical to main): `kv-offloading: dram, kv-offload-backend: { name: lmcache, version: "0.5.4rc1" }, conc-list: [4, 8, 16], spec-decoding: mtp`.
- **benchmarks/single_node/agentic/kimik3_fp4_mi355x_mtp.sh** — new `lmcache)` case arm, modeled on the MiniMax-M3 lmcache arm:
  - installs the LMCache 0.5.4rc1 ROCm wheel (gfx950) with `--no-deps` plus its missing runtime deps, leaving the image's torch/ROCm stack untouched;
  - starts a single MP server per the [Kimi-K3 recipe](https://docs.lmcache.ai/recipes/kimi_k3.html): `--chunk-size 768` (the K3 unified block size at 8 GPUs), `--separate-object-groups` (one object group per sliding-window size for the hybrid KDA/MLA layout, which has two KV-cache groups under MTP), `--enable-extra-logging`, `--max-cpu-workers 8 --max-gpu-workers 1`, `--l1-size-gb $TOTAL_CPU_DRAM_GB`, LRU eviction;
  - wires vLLM to it via `LMCacheMPConnector` (`lmcache.mp.port`), keeping the DSpark `SPEC_ARGS` untouched;
  - unknown backends now fail fast instead of silently running GPU-resident.
- **perf-changelog.yaml** — selects only the new key.
- The L1 budget consumes the generated `TOTAL_CPU_DRAM_GB` verbatim per the agentic README, with an early /dev/shm capacity check (same as the MiniMax-M3 arm).

LMCache + DSpark MTP + the rest of this recipe's feature set has been compatibility-tested; `--separate-object-groups` is the supported handling for the multi-KV-group layout that previously blocked LMCache-under-MTP arms.

## Validation

`process_changelog.py` run exactly as CI does (base = main) emits a 4-row matrix, nothing else:

```
kimik3_tp8_conc4_kvdram-lmcache_spec-mtp   | lmcache 0.5.4rc1 | throughput
kimik3_tp8_conc8_kvdram-lmcache_spec-mtp   | lmcache 0.5.4rc1 | throughput
kimik3_tp8_conc16_kvdram-lmcache_spec-mtp  | lmcache 0.5.4rc1 | throughput
kimik3_tp8_conc16_kvdram-lmcache_spec-mtp  | lmcache 0.5.4rc1 | eval-only
```

`generate_sweep_configs.py test-config` passes for both the new key and the (unchanged) base key; `bash -n` passes on the modified script.


sammshen:
Superseded by #2598, which carries the same commits on a branch in this repo rather than a fork. Same-repo PRs get secrets, so `run-sweep.yml` does the real work there and `/stage-results` can find it.

PR #2576, opened 2026-08-12
[AMD] [AGENTX] GLM-5.2 FP4 MI355X ATOM Agentic MTP

seungrokj:
Co-author @zhuyuhua-v
This PR contains https://github.com/SemiAnalysisAI/InferenceX/pull/2591

## Summary
- Add the MI355X GLM-5.2 FP4 ATOM AgentX recipe (`benchmarks/single_node/agentic/glm5.2_fp4_mi355x_atom_mtp.sh`) with three-token MTP speculative decoding.
- Register `glm5.2-fp4-mi355x-atom-agentic-mtp` in `configs/amd-master.yaml`. TP4 at concurrency 2, 4, 8, and 10 uses LMCache DRAM offload. TP8 at concurrency 1, 2, and 4 is GPU-resident with no KV offload.
- Configure AITER INT4 quick-reduce, `ptpc_fp8` online quantization, FP8 KV cache, concurrency-dependent CUDA graph capture sizes and `max-num-seqs`, and a 16384-token batch cap. Benchmark runs use the committed golden synthetic acceptance length 2.99. Evaluations use real acceptance.
- Add ATOM agentic metrics aggregation and ATOM plotting support.

## Validation
- [x] `bash -n benchmarks/single_node/agentic/glm5.2_fp4_mi355x_atom_mtp.sh`
- [x] Full sweep run https://github.com/SemiAnalysisAI/InferenceX/actions/runs/31765309673 passed all seven throughput points and the selected TP4 real-acceptance eval.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2591, opened 2026-08-13
[AMD][AgentX] support native ATOM server metrics

zhuyuhua-v:
## Summary
- Add a native ATOM metrics backend that recognizes direct worker metrics (`atom:`) and Atomesh-normalized metrics (`atom_`).
- Map ATOM token counters, prefix/LMCache hit rates, and GPU KV-cache usage into the shared AgentX aggregation schema.
- Extend AIPerf dashboard plots to consume ATOM queue, throughput, cache, offload, and preemption metrics while preserving the existing vLLM paths and using token-based units for ATOM LMCache transfers.

## Test plan
- [x] `python -m pytest -q utils/agentic/aggregation/test_process_agentic_result.py` (32 passed)
- [x] `python -m py_compile utils/agentic/aggregation/backends/atom.py utils/generate_aiperf_plots.py`

### DIFF
--- a/utils/agentic/aggregation/backends/__init__.py
+++ b/utils/agentic/aggregation/backends/__init__.py
@@ -4,14 +4,15 @@ from __future__ import annotations
 
 from typing import Any
 
+from .atom import AtomBackend
 from .base import ServerMetricsBackend
 from .dynamo_vllm import DynamoVllmBackend
 from .sglang import SglangBackend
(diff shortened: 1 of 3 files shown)


functionstackx:
already merged https://github.com/SemiAnalysisAI/InferenceX/pull/2576 which contains this diff

PR #2593, opened 2026-08-13
kimik3-fp4-b300-vllm-agentic-dspark: add LMCache DRAM KV-offload arm

sammshen:
## What

Adds a **dedicated config key** `kimik3-fp4-b300-vllm-agentic-dspark-lmcache` with an LMCache DRAM KV-offload arm at TP8 conc **4 / 8 / 10 / 16**, on top of the unchanged DSpark MTP serving stack of `kimik3-fp4-b300-vllm-agentic-dspark` (same image, script, and topology). A separate key means the changelog selects only the LMCache points — the resident and vllm-simple arms of the base key are not re-run.

B300 sister of the MI355X arm in #2583; same LMCache version (0.5.4rc2), same MP-server topology, CUDA install path and NVIDIA-side chunk size.

## Changes

- **configs/nvidia-master.yaml** — new key, purely additive (the base key is byte-identical to main): `kv-offloading: dram, kv-offload-backend: { name: lmcache, version: "0.5.4rc2" }, conc-list: [4, 8, 10, 16], spec-decoding: mtp`, at the base key's `dram-utilization: 0.63` (1,889 GB aggregate at TP8). 4/8/16 land on the base key's vllm-simple ladder at its top three offload points, so LMCache is directly comparable there; 10 fills the 8->16 gap.
- **benchmarks/single\_node/agentic/kimik3\_fp4\_b300\_vllm\_mtp.sh** — new `lmcache)` case arm:
  - installs LMCache's **published CUDA 12.9 build** (the `v0.5.4rc2-cu129` release assets — the CUDA counterpart of the `-rocm` assets the MI355X sister arm installs), not the generic PyPI wheel, via the upstream install line with `--extra-index-url https://download.pytorch.org/whl/cu129`. `--index-strategy unsafe-best-match` has no pip counterpart because it needs none: pip already resolves best-match across every configured index, which is exactly what that flag restores in uv. `torch` is an unpinned LMCache requirement the image already satisfies, so its tested build is left alone. A fail-fast import check follows.
  - starts a single MP server per the [Kimi-K3 recipe](https://docs.lmcache.ai/recipes/kimi_k3.html): `--chunk-size 768` (that recipe's CUDA-path value), `--separate-object-groups` (one object group per sliding-window size for the hybrid KDA/MLA layout, which has more than one KV-cache group under MTP), `--enable-extra-logging`, `--max-cpu-workers 8 --max-gpu-workers 1`, `--l1-size-gb $TOTAL_CPU_DRAM_GB`, LRU eviction, `--shm-name ""` so L1 lives in ordinary process memory rather than being capped by /dev/shm.
  - wires vLLM to it via `LMCacheMPConnector` (`lmcache.mp.port`), keeping the DSpark `--speculative-config` untouched.
  - adds an EXIT trap that stops the LMCache server; the vLLM server's lifecycle is unchanged (the job wrapper still owns it).
  - the unsupported-backend error message now names `lmcache` too.
- **perf-changelog.yaml** — selects only the new key.
- The L1 budget consumes the generated `TOTAL_CPU_DRAM_GB` verbatim per the agentic README.

Note on `--chunk-size`: the connector requires the chunk to be a multiple of every engine KV group's `tokens_per_block`. 768 is the published CUDA-path value; the ROCm sister arm needs 3072 because its hybrid layout registers 1536-token attention groups and a 3072-token KDA state group. If the B300 server log reports a mismatch, this is the one value to raise — the script comment says so and names the log lines to read.

## Validation

`process_changelog.py` run exactly as CI does (base = main) emits a 4-row matrix, nothing else:

```
kimik3_tp8_conc4_kvdram-lmcache_spec-mtp   | lmcache 0.5.4rc2 | throughput
kimik3_tp8_conc8_kvdram-lmcache_spec-mtp   | lmcache 0.5.4rc2 | throughput
kimik3_tp8_conc10_kvdram-lmcache_spec-mtp  | lmcache 0.5.4rc2 | throughput
kimik3_tp8_conc16_kvdram-lmcache_spec-mtp  | lmcache 0.5.4rc2 | throughput
kimik3_tp8_conc16_kvdram-lmcache_spec-mtp  | lmcache 0.5.4rc2 | eval-only
```

`generate_sweep_configs.py test-config` passes for both the new key and the (unchanged) base key; `bash -n` passes on the modified script; `utils/changelog_gate_tests/test_validate_perf_changelog.py` and `utils/test_process_changelog.py` pass (35 tests).

### DIFF
--- a/benchmarks/single_node/agentic/kimik3_fp4_b300_vllm_mtp.sh
+++ b/benchmarks/single_node/agentic/kimik3_fp4_b300_vllm_mtp.sh
@@ -159,19 +159,28 @@ case "${KV_OFFLOAD_BACKEND:-}" in
     lmcache)
         require_agentic_kv_offload_backend lmcache
 
-        # Install LMCache's runtime dependencies that this image does not
-        # already ship, then LMCache itself with --no-deps so pip never
-        # re-resolves the image's tested torch/CUDA stack. None of the three
-        # dependency packages depends on torch, so they resolve safely.
-        # cupy-cuda13x is the CUDA counterpart of the cupy-rocm-7-0 pin in the
-        # MI355X sister arm; LMCache lists it as a hard requirement.
+        # LMCache's own CUDA 12.9 build, not the generic PyPI wheel -- the
+        # release publishes a dedicated -cu129 asset set, the CUDA counterpart
+        # of the -rocm assets the MI355X sister arm installs. This is the
+        # upstream install line, which reads in uv form as:
+        #
+        #   uv pip install lmcache==v$VERSION \
+        #     --extra-index-url https://download.pytorch.org/whl/cu129 \
+        #     --find-links .../expanded_assets/v$VERSION-cu129 \
+        #     --index-strategy unsafe-best-match
+        #
+        # --index-strategy has no pip counterpart because it does not need one:
+        # pip already resolves best-match across every configured index, which
+        # is exactly what unsafe-best-match restores in uv. torch is an
+        # unpinned LMCache requirement that the image already satisfies, so
+        # its tested build is left alone and the cu129 torch index is only
+        # consulted for CUDA wheels pip would otherwise miss.
         LMCACHE_VERSION="0.5.4rc2"
+        LMCACHE_CUDA_INDEX="https://github.com/LMCache/LMCache/releases/expanded_assets/v${LMCACHE_VERSION}-cu129"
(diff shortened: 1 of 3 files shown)


sammshen:
Superseded by #2597, which carries the same commits on a branch in this repo rather than a fork. Same-repo PRs get secrets, so `run-sweep.yml` does the real work there and `/stage-results` can find it.

PR #2602, opened 2026-08-14
[AMD] [AGENTX] Kimi-K3 Performance Tuning

ajith-sirra-amd:
# Summary: [AMD] [AGENTX] KIMI-K3 Perf Tuning

Commit `c0119e6be3` — 3 files changed, 212 insertions(+), 5 deletions(-)

### 1. New: `apply_aiter_pybind11_fix.sh`
Fixes a pybind11 version mismatch between AITER's prebuilt `.so` files (built against torch's bundled pybind11, internals v11) and its JIT builder (which injects standalone pybind11 v12 via `-I`, outranking torch's `-isystem` path). Since pybind11 keeps a separate type registry per internals version, JIT modules couldn't see types registered by the prebuilt core, crashing the server during warmup:
```
TypeError: fmha_fwd_bf16_opus_fwd(): incompatible function arguments
```
The fix patches `aiter/jit/utils/cpp_extension.py` to prefer torch's bundled pybind11, and purges stale JIT-built `.so` artifacts compiled against the wrong version. It's idempotent and verifies the mismatch exists before touching anything.

**Impact:** unblocks `ROCM_AITER_FA` for MLA prefill (previously had to fall back to `FLASH_ATTN`). On 8x MI355X / Kimi-K3 MXFP4 TP8:
- ~24k ctx: 12,953 → 13,524 tok/s (**+4.4%**)
- ~93k ctx: 11,174 → 13,423 tok/s (**+20.1%**)

### 2. New: `apply_triton_mla_cudagraph_fix.sh`
Fixes DSpark spec-decode falling back to fully eager execution on ROCm. `TritonMLAMetadataBuilder` declared `_cudagraph_support = UNIFORM_SINGLE_TOKEN_DECODE`, which silently downgraded `FULL_AND_PIECEWISE` → `PIECEWISE` cudagraph mode whenever spec-decode was enabled, giving the DSpark drafter `CUDAGraphMode.NONE` with no warning logged. `TRITON_MLA` can't simply be swapped for `ROCM_AITER_MLA` since it's the only ROCm MLA backend supporting non-causal multi-token decode (required by the DSpark draft).

The fix patches the backend to declare `AttentionCGSupport.UNIFORM_BATCH` instead, restoring full cudagraph capture.

**Impact:** on 8x MI355X, Kimi-K3 MXFP4 TP8, DSpark, single stream, 600-token generations:
- 14.05 → 77.65 tok/s, ITL 71.16 → 12.88 ms (**5.52x**)
- Output correctness verified in both modes

### 3. Modified: `kimik3_fp4_mi355x_mtp.sh`
- Wires in both fix scripts, invoked right after `install_agentic_deps`.
- Flips the default `MLA_PREFILL_BACKEND` from `FLASH_ATTN` to `ROCM_AITER_FA`, now that the pybind11 fix resolves the underlying AITER crash. `MLA_PREFILL_BACKEND=FLASH_ATTN` remains available as a fallback knob.
- Updates surrounding comments to reflect the new default and measured gains, noting this workload averages ~99k input tokens (making the ~93k ctx benchmark the relevant one).

## Net effect
Both fixes together unlock AITER's faster MLA prefill path and restore full CUDA-graph capture for spec-decode, compounding into a substantial throughput improvement for Kimi-K3 MXFP4 on MI355X.


functionstackx:
lots of patches in this PR, please upstream the patches before reopenning per the AgentX guidelines to avoid flooding the mi355 queue

PR #2615, opened 2026-08-14
fix(gb200): skip srtctl preflight on all paths / GB200：所有配方跳过 srtctl 预检

hshrivastava-droid:
Always pass `--no-preflight` to `srtctl apply` on GB200 so both agentic and
fixed-seq-len paths bypass the login-node model-FS check that fails before
`sbatch`.

GB200 上向 `srtctl apply` 始终传入 `--no-preflight`，让 agentic 与固定序列长度
两条路径都跳过登录节点上会在 `sbatch` 之前失败的模型文件系统预检。


Ankur-singh:
@cquil11 @functionstackx 

Since all the models are pre-stagged in Local NVMe, `--no-preflight` flag should always be included for all GB200 jobs.

cquil11 (approved):

PR #2619, opened 2026-08-15
[AgentX] Mirror GB300 MTP tuning (#2571) to GB200 with 4-GPU->8-GPU workers

xinli-sw:
## Summary
- Mirrors the GB300 AgentX DSV4 MTP tuning from #2571 onto GB200, resizing every worker that used 4 GPUs on GB300 to 8 GPUs on GB200 (agg TP4->TP8, disagg prefill DEP4->DEP8); workers already at 8/12/16 GPUs are copied unchanged.
- Adds 5 new recipe files under `benchmarks/multi_node/srt-slurm-recipes/vllm/deepseek-v4/agentic/` and new `dsv4-fp4-gb200-dynamo-vllm-agentic-mtp2-{agg,disagg}` keys in `configs/nvidia-master.yaml`, kept separate from the existing (differently tuned) GB200 AgentX MTP recipes/keys to avoid touching them.
- `runners/launch_gb200-nv.sh` already had the synthetic-acceptance injection, so no runner changes were needed.

## Test plan
- [ ] `changelog validation` CI gate passes (perf-changelog entry uses `pull/XXX` placeholder, update to this PR's number)
- [ ] Full-sweep CI job for `dsv4-fp4-gb200-dynamo-vllm-agentic-mtp2-agg` / `-disagg` runs green on the gb200-nv cluster


kedarpotdar-nv (approved):
- [x] Verified that this PR has passed PR validation. [GitHub Actions run 31965016666](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/31965016666)
- [x] Verified that this PR passes evals. [GitHub Actions run 31965016666](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/31965016666)

- [Run 31965016666](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/31965016666) passed all seven unique GB200 AgentX benchmark points and all four selected eval variants.
- GSM8K strict-match scores were `0.9227–0.9295`, above the required `0.91` threshold.
- MTP uses three speculative tokens with synthetic acceptance length `2.49`, matching the committed DeepSeek-V4-Pro `thinking_on` golden AL. `EVAL_ONLY` keeps real target verification.
- AgentX traffic uses the chat-completions endpoint.
- The submission uses the upstream `vllm/vllm-openai:nightly-dev-arm64-cu13.0.1-426e59f` image.
- This PR contains only multi-node recipes, so the single-node upstream-recipe documentation requirement is not applicable.
- The diff contains no engine patch, model-architecture FLOP-reduction override, or `append-only: true`.
- Sweep reuse was authorized in [this comment](https://github.com/SemiAnalysisAI/InferenceX/pull/2619#issuecomment-5318148027).

Signed: @kedarpotdar-nv

cquil11 (approved):

PR #2635, opened 2026-08-17
[DIAGNOSTIC] MI355X Kimi-K3 AgentX with no KV transfer config

sammshen:
## DIAGNOSTIC ONLY — not for merge

Isolates whether the `hipErrorIllegalAddress` crashes on the MI355X Kimi-K3 AgentX runs come from the **LMCache offload connector** or from the **node**.

## What this changes

Adds `kimik3-fp4-mi355x-vllm-agentic-mtp-nokv`, identical to `kimik3-fp4-mi355x-vllm-agentic-mtp-lmcache` — same image (`nightly-ac7509e2b1`), same script, TP8/EP1, `spec-decoding: mtp`, same concurrencies `[4, 8, 10, 12, 16]` — with one difference:

```yaml
- { tp: 8, ep: 1, kv-offloading: none, conc-list: [4, 8, 10, 12, 16], spec-decoding: mtp }
```

`kv-offloading: none` means the recipe starts **no LMCache MP server** and passes **no `--kv-transfer-config`**. GPU-resident KV only.

## Why

`hipErrorIllegalAddress` has appeared across:

- the LMCache arm (rc2 **and** rc4)
- #2602's `vllm-simple` arm, which has no LMCache at all
- an eval-only job

…and every one of those landed on **`mia1-p01-g11`**, which is **0-for-8** today. Other nodes produce clean results with the identical code. That points at the node rather than the connector, but the evidence is observational — this run tests it directly.

| observation | node |
|---|---|
| illegal memory access ×6 | g11 |
| task-launch communication failure ×1 | g11 |
| passed | — |

## How to read the result

- **Crashes persist with no KV transfer config** → not LMCache's fault; the node (or the base Kimi-K3 + DSpark stack) is at fault.
- **Crashes vanish** → the connector is implicated after all, and the LMCache arm needs deeper investigation.

Either way it also yields a clean GPU-resident baseline on the current image, which is useful for the LMCache comparison in #2598.

## Note

Node exclusion cannot currently keep this off `g11`: `SALLOC_EXCLUDE` is exported in `benchmark-tmpl.yml` but **no code consumes it** — neither `salloc` invocation passes `--exclude`. Until that's wired up, any point may still draw the bad node.


sammshen:
Diagnostic complete. Result: 5/5 clean (c4/c8/c10/c12/c16) with the KV transfer config removed and DSpark fully active (2.60 mean acceptance), confirming speculative decoding alone does not cause the hipErrorIllegalAddress.

Caveat recorded on the Confluence writeup: this arm removes DRAM offload, so it cannot exonerate a node that only faults under offload load. It does NOT rule out the bad-node hypothesis, which the node-placement data now supports (mia1-p01-g11: 8 failures / 8 jobs under offload).

Superseded by #2598, which reproduces the crash on the current image.

PR #2639, opened 2026-08-18
Kimi K3 vllm GB200 Agentx

wzhao18:


cquil11:
CleanShot 2026-08-19 at 15 06 13

xinli-sw:
@cquil11 how do we move forward with this ? worse case we trigger a re-run, lmk :)

PR #2641, opened 2026-08-18
DIAGNOSTIC: locate the MI355X hipErrorIllegalAddress with serialized kernel launches

sammshen:
**Diagnostic only — not for merge.**

## Why

Every traceback we have for the MI355X `hipErrorIllegalAddress` is unusable for locating the fault. HIP reports kernel errors asynchronously, so the Python stack unwinds at whatever CUDA call next checks the error flag:

```
_fwd_kernel_stage2[grid](            # launch requested
  jit.py:760       run              -> kernel.launch_metadata(...)
  compiler.py:494  launch_metadata  -> self._init_handles()
  compiler.py:462  _init_handles    -> driver.active.get_current_device()
  torch/cuda/__init__.py:1178       -> torch._C._cuda_getDevice()
torch.AcceleratorError: illegal memory access
```

`_init_handles()` is Triton *loading* the compiled kernel — `_fwd_kernel_stage2` had not launched yet. vLLM prints the caveat itself: *"the stacktrace below might be incorrect."* The mid-run variant of the same crash unwinds at `SetDevice` with no Python frames at all.

## What this does

Sets `AMD_SERIALIZE_KERNEL=3` (every launch waits for completion, so the fault surfaces at the causing kernel) and `TORCH_USE_HIP_DSA` / `TORCH_USE_CUDA_DSA` (the abort message explicitly asks for device-side assertion tracking). Engine timeouts are raised since serialized launches are slow.

## Why vllm-simple, not LMCache

`SimpleCPUOffloadConnector` is vLLM's own CPU offload and reproduces the **identical** crash with no LMCache in the picture, which keeps the repro minimal. Both c8 and c10 failed on that backend in [run 31837518310](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/31837518310).

## Context

Established so far: crashes occur **only** with DRAM KV offload and are **backend-agnostic**. The no-KV control ([#2635](https://github.com/SemiAnalysisAI/InferenceX/pull/2635)) passed 5/5 with DSpark fully active, so speculative decoding alone is not the trigger. Full writeup: [Confluence](https://tensormesh.atlassian.net/wiki/spaces/~71202058906ed90a614a9785e822819a2b7c01/pages/1209532422/Kimi+K3+MI355X+InferenceX).

## Expected outcome

The crash reproduces during warmup, so the run only needs to reach startup. **All timings from this key are meaningless** — serialized launches make the run far slower than normal. The deliverable is a traceback naming the real faulting kernel.

## Also in this PR

Repairs #2635's changelog entry: its `pr-link` had been overwritten by a stray duplicate fragment of the #2609 entry (duplicate YAML key, last one winning), leaving #2635 credited to #2609 and #2609 listed twice.


sammshen:
Diagnostic complete, result negative. AMD_SERIALIZE_KERNEL=3 induces its own HSA_STATUS_ERROR_OUT_OF_RESOURCES failures (surfacing as NCCL all_reduce errors), so this arm cannot produce a clean signal regardless of how many times it runs. Zero illegal-access hits across all jobs, but that is uninterpretable given the instrumentation-induced failures.

Also noted: TORCH_USE_HIP_DSA does not work as a runtime env var -- logs still print 'Device-side assertion tracking was not enabled by user'. It requires a torch built with it.

Closing as a dead end. #2598 reproduces the crash at startup without instrumentation, which is a better repro.

PR #2662, opened 2026-08-18
Fix AgentX KV offload metadata

cquil11:
## Summary

- annotate the three GB300 DSV4 Mooncake points with DRAM offload metadata
- annotate the four GB200 DSV4 Mooncake points with DRAM offload metadata
- annotate the six GB300 Qwen3.5 TensorRT-LLM points with native host-cache metadata
- size generated DRAM capacity to match the checked-in recipes: 180 GB, 140 GB, and 128 GiB respectively

## Scope

This changes master-config metadata only. It does not modify benchmark recipes or runtime behavior, so there is no performance changelog entry.

## Validation

- parsed `configs/nvidia-master.yaml`
- generated the three affected config keys successfully
- confirmed exactly 13 points emit `kv-offloading: dram`, the expected backend and version, and matching `total-cpu-dram-gb` values

### DIFF
--- a/configs/nvidia-master.yaml
+++ b/configs/nvidia-master.yaml
@@ -6699,9 +6699,12 @@ qwen3.5-fp4-gb300-dynamo-trt-agentic-disagg:
   disagg: true
   scenarios:
     agentic-coding:
-    - search-space:
+    - dram-utilization: 0.60953
+      search-space:
       # 1P1D: 1 prefill (TP1/EP1/dp-attn), 1 decode (TP2/EP2), conc=44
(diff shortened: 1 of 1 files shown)


cquil11:
Closing because the project scope is now limited to porting active, nondeprecated AMD multi-node configurations to srt-slurm. This PR is outside that scope.

cquil11:
Reopening: this PR predates the recent out-of-scope configuration-porting detour and was closed by mistake during an overly broad cleanup. Its prior state is being restored.

PR #2696, opened 2026-08-20
feat(bench): add --dp-rank-roundrobin for explicit DP-rank routing

rebel-joonhyunglee:
## What

Adds an opt-in `--dp-rank-roundrobin N` flag to `benchmark_serving.py`. When set, request `i` is pinned to data-parallel rank `i % N` via the `X-data-parallel-rank` HTTP header, bypassing vLLM's server-side DP load balancer.

## Why

Under `injection=inf` (request-rate unbounded), vLLM's internal DP balancer can dispatch requests unevenly across ranks — some engines sit idle while others queue — which depresses output TPS on high-DP configs (observed on DP8). This flag measures the **ideal round-robin routing** baseline as a clean A/B against the default balancer, with **no server or SDK change** (the `X-data-parallel-rank` header is already honored by vLLM v0.18+).

Judging signal for the A/B: with the flag on, `mean_ttft / median(p50) ttft` collapses toward ~1 and output TPS rises, indicating the default-balancer TPS gap was a routing artifact rather than a hardware limit.

## Changes

- `backend_request_func.py`: `RequestFuncInput` gains an `extra_headers` field; the OpenAI completions and chat-completions request funcs merge it into request headers.
- `benchmark_serving.py`: `--dp-rank-roundrobin` arg; `benchmark()` assigns the header per request when set. Default `None` → no behavior change.

## Compatibility

Off by default; no change to existing runs. Header is a no-op on servers that don't read it.

### DIFF
--- a/utils/bench_serving/backend_request_func.py
+++ b/utils/bench_serving/backend_request_func.py
@@ -30,6 +30,7 @@ class RequestFuncInput:
     extra_body: Optional[dict] = None
     multi_modal_content: Optional[dict] = None
     ignore_eos: bool = False
+    extra_headers: Optional[dict] = None
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


functionstackx:
We are phasing out single turn in favor of agentic bench so probably won't accept this contribution

PR #2703, opened 2026-08-21
[AgentX] DeepSeek-V4 B200 SGLang update

nvpohanh:
DeepSeek-V4 B200 SGLang update on Agentx


Ankur-singh:
- [x] Verified that this PR has passed PR validation. Please link to GitHub Action workflow that shows this. — https://github.com/SemiAnalysisAI/InferenceX/actions/runs/32454316647
- [x] Verified that this PR passes evals. Please link to GitHub Action workflow that shows this. — https://github.com/SemiAnalysisAI/InferenceX/actions/runs/32454316647
- [x] For agentic workloads: verified that speculative-decoding configs (EAGLE / MTP / draft models) run with simulated synthetic acceptance, with the acceptance-length value taken from the committed golden AL curve in [golden_al_distribution/](https://github.com/SemiAnalysisAI/InferenceX/tree/main/golden_al_distribution) for that model, thinking mode, and draft length.
- [x] Verified that the model architecture isn't changed with benchmark hacks like using --hf-overrides to skipping indexer for every x layers on models that don't natively support this.
- [x] If an company claims that they support vLLM/SGLang as first class LLM inference engines on their hardware, I have verified that the respective vLLM submission made using upstream https://hub.docker.com/u/vllm docker repo, upstream SGLang https://hub.docker.com/u/lmsysorg docker repo.
- [x] If an company claims that they support vLLM/SGLang as first class upstream in-tree LLM inference engines on their hardware, I have have verified that the respective vLLM/SGLang submission has been made before additional frameworks (TRT-LLM, ATOM, etc.).
  - [x] The matching recipe is already published upstream: https://docs.sglang.io/cookbook/autoregressive/DeepSeek/DeepSeek-V4 — linked and compared in the additional detail section below.
- [x] Verified that this PR does not patch the inference engine or serving stack — the pinned image must run as shipped.
- [x] If this PR uses `append-only: true` — N/A, the new changelog entry does not use `append-only`.

**Upstream recipe (single-node requirement).** The published SGLang cookbook page https://docs.sglang.io/cookbook/autoregressive/DeepSeek/DeepSeek-V4 covers this recipe and is already the cited upstream source for the sibling `dsv4-fp4-b200-sglang` key (`configs/nvidia-master.yaml:910`). Major server args match it exactly:
- `--moe-a2a-backend megamoe` — documented as Blackwell-only (B200/B300/GB200/GB300).
- `SGLANG_OPT_DEEPGEMM_MEGA_MOE_USE_FP4_ACTS=1` + `SGLANG_OPT_DEEPGEMM_MEGA_MOE_USE_MXF4_KIND=1` — the page's documented W4A4 pair.
- `SGLANG_OPT_DEEPGEMM_MEGA_MOE_NUM_MAX_TOKENS_PER_RANK=8320` — the page's recommended high-throughput value.
- No `--moe-runner-backend` on the DP path — the page explicitly says not to set it manually under MegaMoE.
- TP path `--moe-runner-backend flashinfer_mxfp4 --enable-deepseek-v4-fp4-indexer` — the page's FP4-indexer command verbatim.
- `--speculative-algorithm EAGLE --speculative-num-steps 3 --speculative-eagle-topk 1 --speculative-num-draft-tokens 4` — matches.

Two flags are not on the page and I am recording them as InferenceX-specific per-lane tuning rather than major deployment args: `--prefill-decode-interval 10` (the page documents the related `--prefill-delayer-max-delay-ms` instead; both are scheduler-cadence knobs of the same character as `--scheduler-recv-interval`) and `--disable-shared-experts-fusion` (a fusion toggle, already used by the merged siblings `benchmarks/single_node/agentic/dsv4_fp4_b300_sglang_mtp.sh:144` and `dsv4_fp4_mi355x_sglang_mtp.sh:216`). The usual harness knobs — concurrency lists, ports, HiCache ratio 8, `mem-fraction-static 0.88`, `swa-full-tokens-ratio 0.02`, image tag — likewise do not need to match.

**One deliberate deviation from the upstream recipe, driven by InferenceX policy.** The cookbook wires MegaMoE into its *high-throughput* preset, whose documented spec-decode setting is "MTP disabled — at saturation the verify step costs more than it saves". This recipe composes MegaMoE **with** EAGLE MTP because `MODELS.md:37` requires that agentic-coding recipes "are run and published with speculative decoding enabled only". That combination is therefore intentional and policy-mandated rather than an upstream mismatch.

**Validation run.** https://github.com/SemiAnalysisAI/InferenceX/actions/runs/32454316647 on commit `391d4c8c4`, which is in this PR. All 12 operating points that survive at head are green — TP8 c1/c2/c3/c4/c5, TP8 HiCache c8/c10/c16/c32, and TP8/EP8/DP-attention HiCache c32/c64/c96 — along with the `agentic eval /` job. To be explicit about the mixed state: that run's overall conclusion is `failure`, attributable solely to the DP-attention c128/c160/c196 jobs, which head `f36c82593` removes from the matrix. Head itself carries no benchmark check-runs because it changes only one line of `configs/nvidia-master.yaml` (the conc-list) and one line of this PR's own `perf-changelog.yaml` entry; the benchmark script is byte-identical between `391d4c8c4` and head, so every reused point's recipe is unchanged. `/reuse-sweep-run 32454316647` is pinned above, which is what makes this run reusable under `utils/find_reusable_sweep_run.py` (`allow_failed` is enabled for a pinned run id).

**Evals.** gsm8k `em_strict = 0.9636` (se 0.0052, n_eff 1319) against the `dsv4` override of `0.91` in `utils/evals/thresholds.yaml:9-10` — not the 0.90 default. Image match verified from the eval server log: `SGLANG_IMAGE_TAG=lmsysorg/sglang:dev-nightly-0820`, identical to `configs/nvidia-master.yaml:932`.

**Golden AL.** `benchmarks/single_node/agentic/dsv4_fp4_b200_sglang_mtp.sh:169-173` pins `SGLANG_SIMULATE_ACC_LEN=2.49` with `SGLANG_SIMULATE_ACC_METHOD=match-expected` and `SGLANG_SIMULATE_ACC_TOKEN_MODE=real-draft-token`, guarded by `if [ "${EVAL_ONLY}" != "true" ]`. With `--speculative-num-steps 3`, that matches `golden_al_distribution/dsv4_mtp.yaml` → `deepseek-v4-pro.thinking_on.3 = 2.49` exactly. `dsv4_mtp.yaml` is the correct curve because the config serves `deepseek-ai/DeepSeek-V4-Pro`, not the DSpark-0813 checkpoint. Thinking mode is pinned structurally: `--chat-template benchmarks/single_node/chat_templates/deepseek_v4_thinking.jinja` unconditionally emits the think prefix on `add_generation_prompt`, so `thinking_on` is the right column. I verified the eval/throughput split from the run artifacts rather than from the code alone — the throughput server log contains all three `SGLANG_SIMULATE_ACC_*` variables and the eval server log contains none, so evals ran real target verification.

**Image provenance.** `lmsysorg/sglang:dev-nightly-0820` from the upstream `lmsysorg` org, pushed 2026-08-20T22:56:51Z, built from sglang commit `92eeed41d7b501ca30a9a4cd45e61852c3ce035a`. It runs as shipped — no `.patch`, `git apply`, `sed -i` on engine sources, site-packages edit, or engine wheel install anywhere in the diff, so no waiver is required. The `python3 -m venv` in the script is the AIPerf client environment; `SGLANG_PYTHON` is captured at line 51 before the venv is prepended to `PATH` at line 55, so the server always launches on the image's own interpreter.

**Changelog.** New entry appended at the physical end of `perf-changelog.yaml`; the head commit's changelog edit touches only this PR's own entry's final description line. No historical entry was modified.

**Non-blocking, but @nvpohanh please read — I would rather see this hardened than merely green.** The c128/c160/c196 failures are a genuine HBM ceiling, not flakes: all three died with the same signature on three different runners, a CUDA OOM inside the DSv4 FP4 indexer during chunked prefill, followed by a DP-attention-wide wedge that burned the full four-hour AIPerf warmup grace (`errors=0`, `returned=94/1,422`). The proximate cause is this PR's change of `CUDA_GRAPH_MAX_BS` from `min(CONC, 64)` to an unbounded `2 * CONC`. The CUDA-graph private pool tracks it exactly — 463 MiB at max_bs 256, 628 MiB at 320, 752 MiB at 392 — while free HBM at the moment of failure was 5.25 / 5.31 / 6.54 GiB against asks of 5.46 / 5.56 / 6.55 GiB. Shortening the concurrency list removes the symptom but leaves the unbounded formula in the merged script, and the surviving c96 point logged 213 `CUDACachingAllocator ... OOM` retry warnings, so it passes only via allocator retry. A cap on `CUDA_GRAPH_MAX_BS`, or a lower `mem-fraction-static` on the DP-attention arm, would be a more durable fix than a shorter conc-list. I am not blocking on it, but it will resurface the moment anyone raises the concurrency list again.

Two smaller notes. `--cuda-graph-max-bs` is logged as deprecated by this image in favour of `--cuda-graph-max-bs-decode`, which the cookbook also uses; harmless today but it will break on a future image bump. And the `eval-only` job ran on the DP-attention arm at concurrency 196, a point head deletes — the recipe hash is identical across that whole arm (`recipe-1325ae161bac3e14` for c32 through c196), so the accuracy result still applies to the surviving points, but the eval row records `conc: 196`.

Signed: `ankur-singh`

Klaud-Cold:
## ✅✅✅ **Verdict: PASS** ✅✅✅

✅ Check 0 (CODEOWNER): PASS — `Ankur-singh` is a named owner of `configs/nvidia-master.yaml`; the other two paths are catch-all-only.
✅ Check 1 (sweep on in-PR commit): PASS — [run 32454316647](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/32454316647) on in-PR commit `391d4c8c4`: green executed `agentic /` check-runs for all 12 points in the head matrix plus green `agentic eval /`. The run's 3 failed jobs (DPA c128/c160/c196) are exactly the points head `f36c825` removes; head vs `391d4c8c4` changes only that conc-list and one changelog line (script byte-identical), and the pinned reuse path (`allow_failed` in `utils/find_reusable_sweep_run.py:494`) accepts this run.
✅ Check 2 (evals pass): PASS — gsm8k `em_strict = 0.9636` (n_eff 1319) vs the `dsv4` bar `0.91` in `utils/evals/thresholds.yaml`; eval job matrix pins `lmsysorg/sglang:dev-nightly-0820`, same image as this PR's config.
✅ Check 3 (recipe): PASS — published [SGLang cookbook DeepSeek-V4 page](https://docs.sglang.io/cookbook/autoregressive/DeepSeek/DeepSeek-V4) linked; verified against its source (`docs/cookbook/.../DeepSeek-V4.mdx`): `--moe-a2a-backend megamoe` + W4A4 env pair, tokens-per-rank 8320, no manual `--moe-runner-backend` under MegaMoE, TP path `flashinfer_mxfp4` + `--enable-deepseek-v4-fp4-indexer`, EAGLE steps 3 / topk 1 / draft 4, DP-attention with prefill delayer — all match. Informational only: `--prefill-decode-interval 10` (page uses the sibling `--prefill-delayer-max-delay-ms`) and `--disable-shared-experts-fusion` are InferenceX-specific tuning; MegaMoE+MTP composition is mandated by MODELS.md's spec-decode-only agentic policy.
✅ Check 4 (reuse command): PASS — `/reuse-sweep-run 32454316647` posted by `Ankur-singh` (COLLABORATOR).
✅ Check 5 (latest checklist): PASS — every current-template item present and checked.
✅ Check 6 (upstream image / ordering): PASS — `lmsysorg/sglang:dev-nightly-0820` from the upstream `lmsysorg` org on B200; framework is SGLang itself, so engine-first ordering is trivially satisfied.
✅ Check 7 (deprecations): PASS — dsv4 agentic-coding MTP is the active replacement arm per MODELS.md as of 2026-08-21.
✅ Check 8 (no architecture hacks): PASS — no `--hf-overrides`/config edits; the FP4 indexer is an upstream-documented precision path and evals pass.
✅ Check 9 (spec-decode chat template): PASS — server pins `--chat-template deepseek_v4_thinking.jinja`; AgentX replay drives chat completions.
✅ Check 10 (no engine patches): PASS — no patching; the venv is AIPerf-client-only and the server launches on the image's own interpreter (`SGLANG_PYTHON` captured before the PATH prepend).
✅ Check 11 (golden AL): PASS — `SGLANG_SIMULATE_ACC_LEN=2.49` with `match-expected`/`real-draft-token`, throughput-only (`EVAL_ONLY` guard); equals `golden_al_distribution/dsv4_mtp.yaml` `deepseek-v4-pro.thinking_on.3 = 2.49`, and the thinking template pins thinking_on.
➖ Check 12 (append-only): N/A — the new `perf-changelog.yaml` entry does not use `append-only: true`.

Ankur-singh (approved):
Approving — CODEOWNER sign-off posted above and verified PASS by codeowner-signoff-verify (12 PASS / 1 N/A). Reuse is pinned to run 32454316647, so no re-run is needed.

Nice work on the golden-AL handling in particular: the `EVAL_ONLY` guard around `SGLANG_SIMULATE_ACC_*` is verifiably correct in the artifacts, not just in the code — throughput server logs carry all three variables and the eval logs carry none.

The one thing I would still like to see followed up, as noted in the sign-off and not blocking this merge: `CUDA_GRAPH_MAX_BS = 2 * CONC` is now unbounded, which is what pushed c128/c160/c196 past the HBM ceiling, and the surviving c96 point logged 213 allocator OOM retries. A cap (or a lower `mem-fraction-static` on the DP-attention arm) would be more durable than a shorter concurrency list.

cquil11 (approved):

PR #2707, opened 2026-08-21
[nv] dsv4 b200 tp8 agentic full_and_piecewise cudagraphs

xinli-sw:
> [!NOTE]
> **Low Risk**
> Benchmark recipe and published concurrency points only; no product, auth, or data-path code.
> 
> **Overview**
> Tunes the B200 DSv4 vLLM agentic MTP recipe so **TP8** captures mixed prefill/decode batches with `FULL_AND_PIECEWISE` CUDA graphs (sizes 100–500 plus the existing decode multiples), while **DEP8** stays on `FULL_DECODE_ONLY` with `mode:0`.
> 
> Also replaces the published TP8 concurrency sweep `[1, 4, 8, 16, 24]` with `[1, 4, 8, 14]`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cb1f60b9f1a4ca7705aab274fe370885fc328736. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/benchmarks/single_node/agentic/dsv4_fp4_b200_vllm_mtp.sh
+++ b/benchmarks/single_node/agentic/dsv4_fp4_b200_vllm_mtp.sh
@@ -257,14 +257,24 @@ if [ "${EVAL_ONLY:-false}" = "true" ]; then
 else
     SPEC_CONFIG="{\"method\": \"mtp\", \"num_speculative_tokens\": $NUM_SPEC_TOKENS, \"rejection_sample_method\": \"synthetic\", \"synthetic_acceptance_length\": 2.49}"
 fi
-CUDA_GRAPH_CAPTURE_SIZES=""
+CAPTURE_SIZE_LIST=()
 for ((num_seqs = 1; num_seqs <= MAX_NUM_SEQS; num_seqs++)); do
-    if [ -n "$CUDA_GRAPH_CAPTURE_SIZES" ]; then
-        CUDA_GRAPH_CAPTURE_SIZES+=","
-    fi
-    CUDA_GRAPH_CAPTURE_SIZES+="$((num_seqs * TOKENS_PER_SEQ))"
+    CAPTURE_SIZE_LIST+=("$((num_seqs * TOKENS_PER_SEQ))")
 done
-COMPILATION_CONFIG="{\"cudagraph_mode\":\"FULL_DECODE_ONLY\",\"cudagraph_capture_sizes\":[${CUDA_GRAPH_CAPTURE_SIZES}],\"mode\":0}"
+# TP additionally captures piecewise graphs for the mixed prefill/decode batches,
+# on top of the full graphs for the uniform decode batches above. Piecewise
+# capture requires the compiled graph, so no "mode":0 on that path. The DEP arm
+# keeps decode-only capture. The decode multiples and the piecewise sizes overlap
+# once MAX_NUM_SEQS*(1+N) passes 100, so sort and de-duplicate.
+if [ "$DP_ATTENTION" != "true" ]; then
+    CAPTURE_SIZE_LIST+=(100 200 300 400 500)
+fi
+CUDA_GRAPH_CAPTURE_SIZES=$(printf '%s\n' "${CAPTURE_SIZE_LIST[@]}" | sort -n -u | paste -sd, -)
+if [ "$DP_ATTENTION" != "true" ]; then
+    COMPILATION_CONFIG="{\"cudagraph_mode\":\"FULL_AND_PIECEWISE\",\"cudagraph_capture_sizes\":[${CUDA_GRAPH_CAPTURE_SIZES}]}"
+else
+    COMPILATION_CONFIG="{\"cudagraph_mode\":\"FULL_DECODE_ONLY\",\"cudagraph_capture_sizes\":[${CUDA_GRAPH_CAPTURE_SIZES}],\"mode\":0}"
+fi
 
 echo "Starting vllm server..."
 export TORCH_CUDA_ARCH_LIST="10.0"
--- a/configs/nvidia-master.yaml
+++ b/configs/nvidia-master.yaml
@@ -974,7 +974,7 @@ dsv4-fp4-b200-vllm-agentic-mtp:
     - dram-utilization: 0.95
       search-space:
       # TP8 SimpleCPU + MTP
-      - { tp: 8, kv-offloading: dram, kv-offload-backend: { name: vllm-simple }, spec-decoding: mtp, conc-list: [1, 4, 8, 16, 24] }
+      - { tp: 8, kv-offloading: dram, kv-offload-backend: { name: vllm-simple }, spec-decoding: mtp, conc-list: [1, 4, 8, 14] }
       # DEP8 SimpleCPU + MTP
       - { tp: 8, ep: 8, dp-attn: true, kv-offloading: dram, kv-offload-backend: { name: vllm-simple }, spec-decoding: mtp, conc-list: [32, 64, 96, 128, 160, 196], router: { name: vllm-router, version: "0.1.14" } }
 
--- a/perf-changelog.yaml
+++ b/perf-changelog.yaml
@@ -6350,3 +6350,11 @@
     - "Use the concurrency-1 aggregated Pareto point and disable TensorRT-LLM performance-metrics responses for all seven points."
   pr-link: https://github.com/SemiAnalysisAI/InferenceX/pull/2657
   
+
+- config-keys:
+    - dsv4-fp4-b200-vllm-agentic-mtp
+  scenario-type:
+    - agentic-coding
+  description:
+    - "Serve the TP8 arm with FULL_AND_PIECEWISE cudagraphs, capturing the mixed prefill/decode batches as well as the uniform decode batches. The DEP8 arm is unchanged."
+  pr-link: https://github.com/SemiAnalysisAI/InferenceX/pull/2707


Oseltamivir (approved):
lgtm

Ankur-singh:
- [x] Verified that this PR has passed PR validation. Please link to GitHub Action workflow that shows this. — https://github.com/SemiAnalysisAI/InferenceX/actions/runs/32526153588
- [x] Verified that this PR passes evals. Please link to GitHub Action workflow that shows this. — https://github.com/SemiAnalysisAI/InferenceX/actions/runs/32526153588
- [x] For agentic workloads: verified that speculative-decoding configs (EAGLE / MTP / draft models) run with simulated synthetic acceptance, with the acceptance-length value taken from the committed golden AL curve in [golden_al_distribution/](https://github.com/SemiAnalysisAI/InferenceX/tree/main/golden_al_distribution) for that model, thinking mode, and draft length.
- [x] Verified that the model architecture isn't changed with benchmark hacks like using --hf-overrides to skipping indexer for every x layers on models that don't natively support this.
- [x] If an company claims that they support vLLM/SGLang as first class LLM inference engines on their hardware, I have verified that the respective vLLM submission made using upstream https://hub.docker.com/u/vllm docker repo, upstream SGLang https://hub.docker.com/u/lmsysorg docker repo.
- [x] If an company claims that they support vLLM/SGLang as first class upstream in-tree LLM inference engines on their hardware, I have have verified that the respective vLLM/SGLang submission has been made before additional frameworks (TRT-LLM, ATOM, etc.).
  - [x] The matching recipe is already published upstream: https://recipes.vllm.ai/deepseek-ai/DeepSeek-V4-Pro — linked and compared arg-by-arg in the additional detail section below.
- [x] Verified that this PR does not patch the inference engine or serving stack — the pinned image must run as shipped.
- [x] If this PR uses `append-only: true` — N/A, the new changelog entry does not use `append-only`.

**Disclosure.** I pushed `cb1f60b9f` to this branch to resolve the `perf-changelog.yaml` conflict with `main` (restored byte-for-byte from `origin/main`, this PR's entry appended at the tail — a strict `+9/-0` append, no historical entry touched) and to add a missing description line, see the changelog note below. No benchmark script or master-config content was altered.

**Upstream recipe.** https://recipes.vllm.ai/deepseek-ai/DeepSeek-V4-Pro (published; source `vllm-project/recipes@main:models/deepseek-ai/DeepSeek-V4-Pro.yaml` plus `strategies/single_node_tp.yaml`). The recipe declares `hardware: b200: verified` and lists `single_node_tp` among its compatible strategies, which is exactly the arm this PR changes. Major args match: TP8 / DP1 / no EP, `--kv-cache-dtype fp8`, `--block-size 256`, 1,048,576 max-model-len, the Blackwell `use_fp4_indexer_cache` override, `--no-enable-flashinfer-autotune`, the `deepseek_v4` tokenizer / tool-call / reasoning parsers, and MTP speculative decoding.

Notably the CUDA-graph change **moves this arm onto the recipe rather than away from it**: the recipe specifies `{"cudagraph_mode":"FULL_AND_PIECEWISE", "custom_ops":["all"]}` for the DSv4 TEP/DEP strategies and reserves `FULL_DECODE_ONLY` for hopper, AMD, and the pd-cluster decode worker. The pre-PR `FULL_DECODE_ONLY` + `"mode":0` on the TP8 arm was the deviation. `custom_ops: ["all"]` is satisfied without being passed — the resolved server config reports `'custom_ops': ['+quant_fp8', 'all', '+quant_fp8']`.

Deltas I am recording as InferenceX-specific rather than recipe mismatches: `num_speculative_tokens: 3` versus the recipe's 2 (the checklist explicitly permits any supported draft length), the synthetic-acceptance knobs (AgentX fairness harness, mandated by `golden_al_distribution/README.md`, not a recipe arg), and the usual harness tuning — `cudagraph_capture_sizes`, `--max-num-seqs`, `--numa-bind`, `--enable-cumem-allocator`, concurrency lists, ports, image tag. One informational note: the DEP8 arm still runs `FULL_DECODE_ONLY` while the recipe's `single_node_dep` says `FULL_AND_PIECEWISE`. That is pre-existing and untouched here, but it is now the arm out of step with upstream and is worth a follow-up.

**Golden AL.** `benchmarks/single_node/agentic/dsv4_fp4_b200_vllm_mtp.sh:258` pins `"rejection_sample_method": "synthetic", "synthetic_acceptance_length": 2.49` with `NUM_SPEC_TOKENS=3`, which is exactly `golden_al_distribution/dsv4_mtp.yaml` → `deepseek-v4-pro.thinking_on.3 = 2.49`. Eval runs keep real target verification: the `EVAL_ONLY` branch at lines 255-259 drops the synthetic knobs, and the eval server log confirms the resolved `SpeculativeConfig(method='mtp', num_spec_tokens=3)` carries no `rejection_sample_method` or `synthetic_acceptance_length`.

**Validation.** https://github.com/SemiAnalysisAI/InferenceX/actions/runs/32526153588 — 10 executed `agentic /` points all `success` (TP8 c1/c4/c8/c14 and TP8/EP8/DPA c32/c64/c96/c128/c160/c196) plus `agentic eval /` `success`. gsm8k `em_strict = 0.9613` (n_eff 1319) against the `dsv4` override of `0.91` in `utils/evals/thresholds.yaml:9-10`, not the 0.90 default. Image match verified two ways: the eval check-run name carries `vllm/vllm-openai:nightly-dev-x86_64-cu13.0.1-426e59f`, matching `configs/nvidia-master.yaml:965`, and both the eval and c14 server logs report engine build `v0.17.2rc1.dev3682+g426e59f62`.

**CUDA-graph memory — I checked this specifically, because a sibling PR hit an HBM ceiling doing something superficially similar.** There is no unbounded growth here; the effective bound actually tightens. `MAX_NUM_SEQS=$((2 * CONC))` is untouched, so lowering the TP8 ceiling from concurrency 24 to 14 takes `MAX_NUM_SEQS` from 48 to 28 and the largest decode capture from 192 to 112 tokens. The five added piecewise sizes (100, 200, 300, 400, 500) are fixed constants that do not scale with concurrency. Measured from the sweep's server logs: CUDA-graph memory is 0.88 GiB at c1, 2.05 GiB at c8 and 3.03 GiB at c14, against roughly 12 GiB of absolute slack, with KV cache steady at 23.56 GiB across all three. A grep for out-of-memory, allocator-retry, and CUDA-error strings across the c1/c8/c14 server logs returns nothing, and all ten points passed on the first non-cancelled attempt.

**Changelog.** Appended at the physical end of `perf-changelog.yaml`; no historical entry edited. I added one description line that the original entry was missing: the TP8 concurrency list changes from `[1, 4, 8, 16, 24]` to `[1, 4, 8, 14]`, which drops the previously published concurrency 16 and 24 points and adds concurrency 14. That is a real edit to an already-published curve and should be on the record alongside the cudagraph change, as comparable PRs have done. @xinli-sw please confirm that dropping 16 and 24 is intentional Pareto pruning rather than a consequence of the new capture configuration.

**Two things that are not blockers but should be fixed or noted:**

1. **The eval never exercised the code path this PR changes.** The single `agentic eval /` job ran on the DEP8 c196 arm, whose server log resolves to `cudagraph_mode: FULL_DECODE_ONLY`. The TP8 `FULL_AND_PIECEWISE` arm has no accuracy coverage in this sweep. CUDA-graph mode does not change the math, so the risk is low, but the coverage gap is real.
2. **The inline rationale for dropping `"mode":0` is inaccurate in this image.** The comment says piecewise capture "requires the compiled graph, so no `\"mode\":0` on that path", but this build auto-enables `VLLM_USE_BREAKABLE_CUDAGRAPH=1` and logs `disabling vLLM's torch.compile pipeline. Equivalent to -cc.mode=none`, resolving to `compilation_config={'mode': CompilationMode.NONE}` either way. Piecewise capture still succeeded (32 graphs at c14) via the breakable-cudagraph path. So removing `"mode":0` is a no-op today; the latent risk is that a future image bump where breakable-cudagraph is not auto-enabled would silently engage inductor compilation on the TP8 arm.

Also worth correcting for the record: the `claude[bot]` inline comment flagging `pr-link: .../pull/TBD` is a false positive — it cites a SHA that is not this PR's head, and the committed file has always carried the real link. Its separate note about the stale file-header comment at lines 14-17 (which still describes `FULL_DECODE_ONLY` for the throughput sweep) is correct and worth a follow-up.

Signed: `ankur-singh`

Ankur-singh (approved):

Klaud-Cold:
## ✅✅✅ **Verdict: PASS** ✅✅✅

✅ Check 0 (CODEOWNER): PASS — `@Ankur-singh` is a named owner of `configs/nvidia-master.yaml`; the other two changed paths are catch-all-only.
✅ Check 1 (sweep on in-PR commit): PASS — commit `4552491` (still in this PR) has 10 executed `agentic /` points and `agentic eval /` all `success`, from [run 32526153588](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/32526153588).
✅ Check 2 (evals pass): PASS — gsm8k `em_strict` 0.9613 (n_eff 1319) vs the dsv4 bar of 0.91 in `utils/evals/thresholds.yaml`; eval ran on image `vllm/vllm-openai:nightly-dev-x86_64-cu13.0.1-426e59f`, matching this PR's `configs/nvidia-master.yaml` entry.
✅ Check 3 (recipe): PASS — published recipe https://recipes.vllm.ai/deepseek-ai/DeepSeek-V4-Pro (`b200: verified`, `single_node_tp` compatible) matches the changed TP8 arm's major args: TP8/no-EP, `--kv-cache-dtype fp8`, `--block-size 256`, 1M `--max-model-len`, FP4 indexer cache, MTP spec decoding, `--no-enable-flashinfer-autotune`, `deepseek_v4` parsers; the recipe's DSv4 strategies themselves use `FULL_AND_PIECEWISE`. Informational (not blockers): `num_speculative_tokens` 3 vs the recipe's 2 (any supported draft length is allowed), and the untouched DEP8 arm keeps `FULL_DECODE_ONLY` and the FP4-expert `deep_gemm_amxf4_mega_moe` form of the recipe's Blackwell mega-MoE kernel.
✅ Check 4 (reuse command): PASS — `/reuse-sweep-run` by `xinli-sw` and `/reuse-sweep-run 32526153588` by `Ankur-singh`, both `COLLABORATOR`.
✅ Check 5 (latest checklist): PASS — every item of the current `docs/PR_REVIEW_CHECKLIST.md` template is present and checked in the sign-off.
✅ Check 6 (upstream image / engine-first): PASS — `vllm/vllm-openai:nightly-dev-x86_64-cu13.0.1-426e59f` is from the upstream vLLM Docker Hub org; framework is vLLM, so engine-first ordering does not apply.
✅ Check 7 (deprecations): PASS — dsv4 agentic-coding with MTP is the published arm per `MODELS.md` as of 2026-08-22.
✅ Check 8 (architecture hacks): PASS — the diff only changes cudagraph mode/capture sizes and a concurrency list; no FLOPs-reducing overrides.
✅ Check 9 (chat template): PASS — the agentic replay drives `/v1/chat/completions` with `--endpoint-type chat` (`benchmark_lib.sh`), and evals use `local-chat-completions --apply_chat_template`.
✅ Check 10 (engine patches): PASS — no `.patch`/heredoc/site-packages edits or forked engine wheels; the pinned image runs as shipped.
✅ Check 11 (golden AL): PASS — throughput pins `"rejection_sample_method": "synthetic", "synthetic_acceptance_length": 2.49` with `num_speculative_tokens: 3`, exactly `golden_al_distribution/dsv4_mtp.yaml` → `deepseek-v4-pro.thinking_on.3`; the `EVAL_ONLY` branch correctly reverts to real target verification.
➖ Check 12 (append-only): N/A — the new `perf-changelog.yaml` entry does not use `append-only: true`.

PR #2710, opened 2026-08-23
Move the DSv4 MI355X SGLang AgentX arm to sgl-dev v0.5.18 and split TP4/TP8 by concurrency

karverma-amd:
## Summary

Updates `dsv4-fp4-mi355x-sglang-agentic-mtp` to the image and serving flags validated on the AMD fork, and splits the search space so TP4 covers the low-concurrency end.

**Image and serving flags** (`a3feffec`)
- Image `lmsysorg/sglang-rocm:v0.5.17-rocm720-mi35x-20260813` → `rocm/sgl-dev:v0.5.18-rocm720-mi35x-20260822`
- Enable the three DSv4 kernel fusions this image carries, all of which default off in `environ.py`: `SGLANG_OPT_USE_AITER_BATCHED_GEMM`, `SGLANG_OPT_NATIVE_BPRESHUFFLE_SCALE`, `SGLANG_OPT_FUSE_COMPRESS_NORM_ROPE`
- `mem-fraction-static` 0.85 → 0.89, enlarging the full-attention KV pool. The DSv4 compressor state pools are sized from that pool and allocated *after* it, outside this budget, so the remainder still has to cover them — which is why this stops short of 0.90.
- `--disable-shared-experts-fusion` → `--enforce-shared-experts-fusion`. The tuned MoE tables in this image cover the resulting `(inter_dim 384, expert 385, topk 7)` shape at the token sizes this workload hits; on the previous image that shape fell back to a heuristic FlyDSL kernel.

**Search space** (`5823400`)
- Add a TP4 arm at concurrency `[1, 2, 4, 8, 10]` with no host KV tier, where halving the GPUs per replica raises throughput per GPU
- Reduce the TP8 no-offload arm from `[1, 2, 4, 8, 16]` to `[16]`, so TP8 picks up where TP4 leaves off rather than duplicating those points
- Net effect on published data: drops TP8 concurrency 1, 2, 4, 8 and adds TP4 concurrency 1, 2, 4, 8, 10. The hicache arm is unchanged at `[16, 32, 48]`, and concurrency 16 still appears on both TP8 arms to isolate the host KV tier's gain.

TP4 doubles per-rank weights and leaves far less room for KV, which is why it stops at concurrency 10 and ships no hicache variant.

The `runner` (`cluster:mi355x-amds`) is unchanged, and no directory structure changed.

[...]

PR #2713, opened 2026-08-24
Refresh the DSv4 MI355X SGLang AgentX arm to v0.5.18 and split TP4/TP8 by concurrency

karverma-amd:
## Summary

Refreshes `dsv4-fp4-mi355x-sglang-agentic-mtp` onto the v0.5.18 MI355X image with the serving flags validated on the AMD side, and splits the search space so TP4 covers the low-concurrency end instead of TP8 duplicating it.

Replaces #2710, which was opened from a fork.

**Image and serving flags**
- `lmsysorg/sglang-rocm:v0.5.17-rocm720-mi35x-20260813` → `lmsysorg/sglang-rocm:v0.5.18-rocm720-mi35x-20260822`
- Enable `SGLANG_OPT_USE_AITER_BATCHED_GEMM=1`, the aiter batched GEMM for the absorbed MLA projections, which defaults off in `environ.py`
- `mem-fraction-static` 0.85 → 0.89, enlarging the full-attention KV pool. The DSv4 compressor state pools are sized from that pool and allocated *after* it, outside this budget, so the remainder still has to cover them — which is why this stops short of 0.90.
- `--disable-shared-experts-fusion` → `--enforce-shared-experts-fusion`. The tuned MoE tables in this image cover the resulting `(inter_dim 384, expert 385, topk 7)` shape at the token sizes this workload hits.

**Search space**
- Add a TP4 arm at concurrency `[1, 2, 4, 8, 10]` with no host KV tier, where halving the GPUs per replica raises throughput per GPU
- Reduce the TP8 no-offload arm from `[1, 2, 4, 8, 16]` to `[16]`, so TP8 picks up where TP4 leaves off
- Net effect on published data: drops TP8 concurrency 1, 2, 4 and 8; adds TP4 concurrency 1, 2, 4, 8 and 10. The hicache arm is unchanged at `[16, 32, 48]`, and concurrency 16 still appears on both TP8 arms to isolate the host KV tier's gain.

TP4 doubles per-rank weights and leaves far less room for KV, which is why it stops at concurrency 10 and ships no hicache variant.

`runner` (`cluster:mi355x-amds`) is unchanged and no directory structure changed. The key's comment is also corrected: it no longer claims the image matches the disagg AgentX entry (that key is still on `v0.5.17-...-20260817`), and "pure TP8 only" becomes "tensor-parallel only", which is what it meant — the exclusion is DP-attention, not TP4.

## Test plan

- [x] `bash -n benchmarks/single_node/agentic/dsv4_fp4_mi355x_sglang_mtp.sh`
- [x] `configs/amd-master.yaml` and `perf-changelog.yaml` both parse
- [x] Config generation yields 9 cells on the new image: TP4 at conc 1/2/4/8/10 no-offload, TP8 at conc 16 no-offload plus 16/32/48 hicache
- [x] `process_changelog.py --base-ref main --head-ref HEAD` triggers all 9 cells plus the conc-48 agentic eval
- [ ] Sweep run on MI355X — this image and flag set has not yet been run on this key

Made with [Cursor](https://cursor.com)


1am9trash (approved):
- [x] Verified that this PR has passed PR validation. Please link to GitHub Action workflow that shows this. Link: https://inferencex.semianalysis.com/inference?unofficialRun=32724927772
- [x] Verified that this PR passes evals.  Please link to GitHub Action workflow that shows this. Link: https://inferencex.semianalysis.com/evaluation?unofficialRun=32724927772

- Recipe at https://docs.sglang.io/cookbook/autoregressive/DeepSeek/DeepSeek-V4#hw=mi355x&variant=pro&quant=fp4&strategy=high-throughput&nodes=single

Signed: @1am9trash

Klaud-Cold:
## ✅✅✅ **Verdict: PASS** ✅✅✅

✅ Check 0 (CODEOWNER): PASS — `configs/amd-master.yaml` is specifically owned by @1am9trash; the other changed paths carry only the catch-all, which a recognized CODEOWNER satisfies.
✅ Check 1 (passing sweep on in-PR commit): PASS — run [32724927772](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/32724927772) on head `58bbb80` executed all 8 `agentic /` benchmark cells plus `agentic eval /` with conclusion `success` (agentic-only PR, so the `single-node */` / `eval /` lanes were legitimately not selected).
✅ Check 2 (evals pass): PASS — gsm8k `em_strict` 0.9674 (n_eff 1319) ≥ the 0.91 dsv4 threshold, run on this PR's image `lmsysorg/sglang-rocm:v0.5.18-rocm720-mi35x-20260822`.
✅ Check 3 (recipe merged & args match): PASS — the linked published [SGLang cookbook page](https://docs.sglang.io/cookbook/autoregressive/DeepSeek/DeepSeek-V4#hw=mi355x&variant=pro&quant=fp4&strategy=high-throughput&nodes=single) documents MI355X Pro FP4 single-node with the same major args: `--attention-backend dsv4`, `--page-size 256`, `--kv-cache-dtype fp8_e4m3`, `--enforce-shared-experts-fusion`, EAGLE 3-1-4, `SGLANG_OPT_USE_AITER_BATCHED_GEMM`; the TP8 arm matches the verified TP-only cell, TP4 is a documented parallelism option for this variant, and the hicache arm matches the cookbook's AMD hicache spec (`page_first_direct` + `direct`). Informational InferenceX tuning diffs only: mem-fraction 0.89 vs 0.90, swa-full-tokens-ratio 0.10 vs 0.15, hicache ratio 1.5, image tag v0.5.18 vs the page's v0.5.17.
✅ Check 4 (reuse command): PASS — `/reuse-sweep-run 32724927772` posted by @1am9trash (COLLABORATOR).
✅ Check 5 (latest checklist): PASS — every item of the current `docs/PR_REVIEW_CHECKLIST.md` template is present and checked.
✅ Check 6 (upstream image / engine-first): PASS — `lmsysorg/sglang-rocm:v0.5.18-rocm720-mi35x-20260822` is the upstream lmsysorg org, and the entry is `framework: sglang` so engine-first ordering is inherently satisfied.
✅ Check 7 (deprecated models): PASS — dsv4 agentic-coding MTP is the active, published arm per MODELS.md as of 2026-08-24.
✅ Check 8 (no architecture hacks): PASS — no `--hf-overrides` or model-config edits; `--enforce-shared-experts-fusion` fuses the same computation without removing FLOPs.
✅ Check 9 (chat templates): PASS — AgentX trace replay drives chat-completions with the model's native chat template (no truncating template override).
✅ Check 10 (no engine patches): PASS — no patching of the serving stack; the script installs only client-side harness deps.
✅ Check 11 (golden AL): PASS — throughput runs pin `SGLANG_SIMULATE_ACC_LEN=2.49` (`match-expected`, `real-draft-token`), equal to `golden_al_distribution/dsv4_mtp.yaml` thinking_on / 3 speculative steps; eval-only runs keep real target verification.
➖ Check 12 (append-only): N/A — the new perf-changelog entry does not use `append-only: true`.

1am9trash:
Hi, @functionstackx 
May you help to review this? Thanks.

functionstackx:
@1am9trash plz ping @cquil11 or @adibarra for review plz

cquil11 (approved):

PR #2761, opened 2026-08-27
[Klaud] [AMD][Power] fix: wait for AMD telemetry to cover the benchmark window end before stopping the monitor / 修复 AMD 遥测在基准窗口结束前被截断导致功耗校验失败的问题

edwingao28:
Closes gap G10 (PLAN-05). Fixes the `benchmark_window_not_bracketed` power-validation failures on AMD single-node AgentX runs (e.g. run 32433563482, `kimik3-fp4-mi355x-atom-agentic-mtp`, 3 of 4 points failed).

## Problem

On run 32433563482 conc1 the amd-smi stream's last CSV tick is `1787277605` (integer epoch second) while the aiperf window end is `1787277609.157` — telemetry stops ~4 s before the formal window end. Three interacting mechanisms, none bounded by the old fixed `sleep interval+2` before killing the pipeline:

1. **Consumer-side data loss at kill** — `GPU_MONITOR_PID` is the `awk` consumer, not `amd-smi`; rows sitting in the OS pipe (awk's per-row `fflush()` can block on NFS-backed `$result_dir`) are discarded at kill.
2. **Producer staleness** — one `amd-smi metric -p -c -t -u` iteration over 8 GPUs takes non-trivial time, so the newest emitted tick can trail wall clock by seconds.
3. **Degenerate teardown rows** — SMI can emit N/A or 0 W power cells near termination; an N/A row in the ±3 s ingest band flipped validity via `invalid_power_sample`, and a 0 W row past the window end could silently fake end bracketing while corrupting the boundary interpolation.

## Solution

**`benchmarks/benchmark_lib.sh` — deterministic poll-until-covered stop.**
`stop_gpu_monitor`'s AMD branch now polls `$GPU_METRICS_CSV` until every observed GPU has a usable tick (numeric epoch timestamp, numeric power > 0) stamped ≥ floor(stop-entry epoch)+1, then kills. The window end is always ≤ the stop-entry wall clock and amd-smi stamps integer seconds, so that tick strictly covers any fractional window end — no aiperf knowledge needed in the shell. Details:

- Bounded by `AMD_MONITOR_STOP_TIMEOUT_S` (default 30 s, env-overridable, `0` skips the wait; a non-integer value warns and falls back to 30 instead of aborting the stop). On timeout or early monitor death: warn and proceed — aggregation then flags the point exactly as today (fail-safe, never fail-silent).
- Non-epoch timestamps (older amd-smi emitting ISO strings) fall back to the legacy `interval+2` fixed tail once; millisecond epochs (>1e12) are normalized to seconds, mirroring `_parse_timestamp` in `utils/aggregate_power.py`.
- The awk coverage helper is POSIX-only (no interval regex, no gawk extensions; works under mawk/busybox awk) and neutralizes amd-smi's quoted list cells (embedded commas), CR line endings, and blank lines between tick groups — all verified against the real run-32433563482 CSV.
- AgentX `INT`/`TERM`/`EXIT` traps stop in **abort** mode (`AMD_MONITOR_STOP_TIMEOUT_S=0`): a cancelled run's power validity is moot and teardown stays fast. The explicit post-replay stop performs the full wait. Idempotency via `agentx_monitor_stopped` preserved.
- **NVIDIA branch is behaviorally unchanged.** This also fixes every AMD single-node consumer of `stop_gpu_monitor` (all `benchmarks/single_node/fixed_seq_len/*` recipes), not just AgentX.

**`utils/aggregate_power.py` — explicit boundary-degenerate handling.**
`integrate_power` now skips-and-counts rows inside the ±`max_sample_gap_s` ingest band but **outside** the formal window whose power is unparseable, non-finite, or ≤ 0, into a new per-GPU sidecar field `boundary_degenerate_rows` (`"unknown"` buckets rows without a GPU identity). They are never used for bracketing and never poison validity. **In-window semantics are byte-identical**: in-window N/A still ⇒ `invalid_power_sample`; in-window 0 W still integrates.

### Contract compatibility (PLAN-06 / PLAN-07)

- No new top-level reason codes; `benchmark_window_not_bracketed` keeps meaning "a real usable sample does not bracket the boundary".
- Aggregate-row schema untouched (`power_valid` still numeric; no `power_invalid_reasons` on aggregates).
- Sidecar `schema_version` stays 1; `boundary_degenerate_rows` is additive and readers tolerate its absence.

### Validity flips on re-aggregated legacy artifacts (intentional)

Only when explicitly re-running aggregation on old raw artifacts (e.g. the recover-failed-ingest path):
- streams previously failed **only** by an outside-window N/A row now validate (strict improvement);
- streams whose only end bracket was a degenerate 0 W row now fail `benchmark_window_not_bracketed` (they were silently corrupt before).
Already-ingested aggregate rows are untouched.

### Cost

Normal case ~1–2 s at stop (first whole second after stop entry). Pathological hang: up to 30 s (was 3 s), bounded, logged, env-tunable.

## Tests

- `bash -n benchmarks/benchmark_lib.sh` clean.
- `utils/test_aggregate_power.py` + `utils/agentic/aggregation/test_power_lifecycle.py`: 74 passed (57 + 17).
  - 5 new aggregation tests — N/A-outside-window skipped-and-counted; 0 W tail no longer fakes bracketing (flips a silently-corrupt legacy case to explicit-invalid); in-window 0 W semantics frozen; sidecar key present-and-empty on clean streams; regression fixture encoding the run-32433563482 conc1 shape (real 21-column MI355X watch header, integer ticks ending 4 s before fractional end `...609.157497`, real power values, quoted list cells, CR endings, blank inter-tick lines, plus the documented synthetic N/A / 0 W teardown shapes) asserting `benchmark_window_not_bracketed` attribution.
  - 6 new shell-contract tests running the **real** `stop_gpu_monitor` against a scripted producer (coverage wait for all GPUs; degenerate rows never satisfy coverage → timeout warning; per-GPU min semantics; ISO-timestamp legacy fallback; non-integer timeout survival; millisecond-epoch normalization), plus the signal test now pins abort mode (`AMD_MONITOR_STOP_TIMEOUT_S=0` at trap-driven stop) and the new trap strings.
- `utils/test_process_result.py`: the AMD stop test now pins the new contract (covered tick ⇒ no legacy sleep, energy snapshot still written); truncated-row repair test runs with the wait skipped.
- Broader power suites (`test_aggregate_power*`, `test_process_result`, `utils/agentic/aggregation/`): 224 passed.
- Full `utils/` suite: 804 passed; the only 2 failures (`utils/evals/test_run_eval_dispatch.py`) reproduce identically on clean `origin/main` in this local environment (macOS temp-path issue), unrelated to power.
- Shell-contract tests re-run 3× locally, no flakes (0.2 s producer cadence vs ≥1 s coverage threshold ⇒ ≥5× margin).

## Review

Two independent review passes; every finding verified and resolved:

1. **CONFIRMED (minor):** a non-integer `AMD_MONITOR_STOP_TIMEOUT_S` (e.g. `"30s"`) aborted the whole `stop_gpu_monitor` call via a bash arithmetic error under `set -e` — leaking the monitor process and skipping tail repair and the energy sidecar. **Fixed:** the timeout is sanitized once before any arithmetic (warn + fall back to 30); the timeout warning now always prints a real number. Pinned by `test_amd_stop_survives_non_integer_timeout`.
2. **PLAUSIBLE (nit, raised by both reviewers):** a hypothetical amd-smi build stamping millisecond epochs would trivially satisfy coverage (~1.8e12 ≥ any second-scale target) and skip the tail wait entirely. No such build is known, and the failure mode was fail-safe, but the hardening is cheap. **Fixed:** the awk helper normalizes >1e12 timestamps to seconds, mirroring `_parse_timestamp`. Pinned by `test_amd_stop_normalizes_millisecond_epoch_timestamps`.
3. **CONFIRMED (minor):** the PR-opening step was deferred out of the implement phase per workflow rules. **Resolved:** this PR.

## Evidence note

The retrieved run-32433563482 conc1 artifact shows the trailing rows all carry **valid** socket_power (254–264 W) with N/A activity/voltage cells — the observed failure is purely mechanisms (1)+(2) (the stream stops 4 s early). The degenerate-row handling hardens against mechanism (3) using the plan's documented synthetic shapes.

## Hardware smoke (for PLAN-14; not a merge gate for this PR, but the gap-closure gate)

Re-sweep `kimik3-fp4-mi355x-atom-agentic-mtp` (`configs/amd-master.yaml:656`) from this PR branch on the MI355X ATOM cluster. For each point (conc 1, 4, 8, 10) in the raw-results artifact:

- `agg_*.json` has `power_valid == 1` and finite `avg_power_w` / `total_gpu_energy_j`;
- `results/**/power_validation.json` has `reasons == []`, and for every GPU the newest usable `gpu_metrics.csv` tick ≥ `benchmark_window.end_time_unix`;
- `boundary_degenerate_rows` (if non-empty) shows teardown rows were counted, not integrated;
- `accumulator_check.within_tolerance == true` where the sidecar snapshots exist;
- the job log shows no `never covered the stop request` warning, and per-point wall time increased by ≲ a few seconds versus the failing run.

If any point still fails with `benchmark_window_not_bracketed`, attach its CSV tail + `power_validation.json` to this PR before iterating (per-GPU `sampling_gap_exceeded` would instead indicate the separate cadence issue listed under PLAN-05 follow-ups).

---

> [!NOTE]
> **Medium Risk**
> Changes benchmark teardown timing and power-validation semantics for AMD telemetry and boundary rows; affects AgentX power metrics but is bounded, logged, and covered by contract tests.
> 
> **Overview**
> Fixes AMD AgentX **`benchmark_window_not_bracketed`** failures when amd-smi integer-second ticks trail the fractional benchmark window end or pipe-buffered rows are lost at kill.
> 
> **`stop_gpu_monitor` (AMD)** no longer uses a fixed `sleep interval+2` before killing the watch pipeline. It polls **`gpu_metrics.csv`** until every GPU has a usable sample (numeric epoch timestamp, power &gt; 0) at or after the first whole second past stop entry, bounded by **`AMD_MONITOR_STOP_TIMEOUT_S`** (default 30s; `0` skips the wait). ISO timestamps fall back to the legacy tail sleep; ms epochs are normalized like **`aggregate_power`**. AgentX **INT/TERM/EXIT** traps call stop in **abort** mode (`AMD_MONITOR_STOP_TIMEOUT_S=0`); the normal post-replay stop keeps the full wait.
> 
> **`integrate_power`** skips N/A, non-finite, or ≤0 W rows that sit in the ingest band but outside the formal window, counting them in sidecar **`boundary_degenerate_rows`** instead of invalidating power or faking end bracketing (0 W tails now fail bracketing explicitly). Tests cover the shell poll, signal abort path, and aggregation regressions including the MI355X run shape.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3c088145271d35c0e6e988fd6de3e00c8dacf34d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/benchmarks/benchmark_lib.sh
+++ b/benchmarks/benchmark_lib.sh
@@ -266,9 +266,14 @@ _amd_monitor_min_covered_tick() {
             if (cell[power_col] !~ /^[0-9]+(\.[0-9]+)?$/) next
             if (cell[power_col] + 0 <= 0) next
             if (cell[gpu_col] == "") next
+            ts = cell[1] + 0
+            # Mirror _parse_timestamp in utils/aggregate_power.py: normalize
+            # millisecond epochs so a ms-stamping amd-smi build cannot
+            # trivially satisfy any second-scale stop target.
+            if (ts > 1e12) ts /= 1000
             gpu = cell[gpu_col]
-            if (!(gpu in newest) || cell[1] + 0 > newest[gpu])
-                newest[gpu] = cell[1] + 0
+            if (!(gpu in newest) || ts > newest[gpu])
+                newest[gpu] = ts
         }
         END {
             have = 0
@@ -286,12 +291,20 @@ _amd_monitor_min_covered_tick() {
 # on timeout or early monitor death it warns and lets aggregation attribute
 # the missing coverage (fail-safe, never fail-silent).
 _wait_for_amd_stop_coverage() {
-    local target deadline covered
-    if [[ "${AMD_MONITOR_STOP_TIMEOUT_S:-30}" -le 0 ]]; then
+    local target deadline covered timeout_s
+    # A non-integer timeout (e.g. "30s") would abort the whole stop_gpu_monitor
+    # call under `set -e` at the arithmetic below, leaking the monitor process
+    # and skipping tail repair + the energy sidecar; fall back to the default.
+    timeout_s="${AMD_MONITOR_STOP_TIMEOUT_S:-30}"
+    if [[ ! "$timeout_s" =~ ^-?[0-9]+$ ]]; then
+        echo "[GPU Monitor] Warning: ignoring non-integer AMD_MONITOR_STOP_TIMEOUT_S='$timeout_s', using 30" >&2
+        timeout_s=30
+    fi
+    if [[ "$timeout_s" -le 0 ]]; then
         return 0
     fi
     target=$(( $(date +%s) + 1 ))
-    deadline=$(( target + ${AMD_MONITOR_STOP_TIMEOUT_S:-30} ))
+    deadline=$(( target + timeout_s ))
     while :; do
         covered=$(_amd_monitor_min_covered_tick)
         if [[ -z "$covered" ]]; then
@@ -308,7 +321,7 @@ _wait_for_amd_stop_coverage() {
             return 0
         fi
         if [[ "$(date +%s)" -ge "$deadline" ]]; then
-            echo "[GPU Monitor] Warning: AMD telemetry never covered the stop request within ${AMD_MONITOR_STOP_TIMEOUT_S}s (covered=$covered target=$target)" >&2
+            echo "[GPU Monitor] Warning: AMD telemetry never covered the stop request within ${timeout_s}s (covered=$covered target=$target)" >&2
             return 0
         fi
         sleep 1
--- a/benchmarks/benchmark_lib.sh
+++ b/benchmarks/benchmark_lib.sh
@@ -111,6 +111,8 @@ unset _benchmark_caller
 GPU_MONITOR_PID=""
 GPU_MONITOR_VENDOR=""
 GPU_MONITOR_INTERVAL=1
+# Bounded wait for AMD telemetry to cover a stop request; 0 skips the wait.
+AMD_MONITOR_STOP_TIMEOUT_S="${AMD_MONITOR_STOP_TIMEOUT_S:-30}"
 GPU_METRICS_CSV="${GPU_METRICS_CSV:-gpu_metrics.csv}"
 NVIDIA_GPU_MONITOR_QUERY="timestamp,index,power.draw,temperature.gpu,clocks.current.sm,clocks.current.memory,utilization.gpu,utilization.memory"
 export GPU_METRICS_CSV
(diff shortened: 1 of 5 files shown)


edwingao28:
Superseded by #2767 (branch renamed to fix/amd-agentx-window-bracketing); identical diff.

PR #2792, opened 2026-09-01
[Klaud Cold] [AMD] Bump DSV4 MI355X vLLM 8k/1k to the 2026-09-01 nightly / [Klaud Cold] [AMD] 将 DSV4 MI355X vLLM 8k/1k 更新至 2026-09-01 nightly

jiacao-amd:
## Summary

Update both fixed-sequence DeepSeek-V4-Pro FP4 MI355X vLLM configurations for the 8k/1k workload:

- `dsv4-fp4-mi355x-vllm` (STP)
- `dsv4-fp4-mi355x-vllm-mtp` (MTP)

Image update:

- From: `vllm/vllm-openai-rocm:nightly-09663abde0f50944a8d5ea30120666024b503faa`
- To: `vllm/vllm-openai-rocm:nightly-7c5dc571cbd1064ecc8a9b1045637ff647aa22cb`
- Digest: `sha256:f0bdaf5217a09949842b45c1ea1f12260d3205ec81f143b320dfc2eb3ec95e55`
- Docker Hub push time: `2026-09-01T05:31:02Z`
- vLLM commit: [`7c5dc571`](https://github.com/vllm-project/vllm/commit/7c5dc571cbd1064ecc8a9b1045637ff647aa22cb)

The 8k/1k TP8 concurrency range remains unchanged at 4-512 for both STP and MTP.

## vLLM DSV4 optimization review

The new nightly includes the following relevant changes merged after the 2026-08-17 nightly used by #2590:

- [vLLM #52737](https://github.com/vllm-project/vllm/pull/52737): fuse DeepSeek-V4 mHC post/pre and RMSNorm with AITER.
- [vLLM #52882](https://github.com/vllm-project/vllm/pull/52882): optimize DeepSeek-V4 C4A top-k with an automatic AITER/native hybrid selector on gfx950.
- [vLLM #53838](https://github.com/vllm-project/vllm/pull/53838): fuse the DeepSeek-V4 C4 compressor GEMMs.
- [vLLM #53540](https://github.com/vllm-project/vllm/pull/53540): fuse SWA q/kv RMSNorm and q FP8 group quant on the ROCm/AITER path.
- [vLLM #53382](https://github.com/vllm-project/vllm/pull/53382): tune cooperative top-k for medium batch sizes.

[...]

PR #2850, opened 2026-09-07
[AMD][DSV4] Repin MI355X vLLM entries to the nightly shipping FHMoE

Fangzhou-Ai:
## What

Repins the three DSv4 MI355X vLLM entries — `dsv4-fp4-mi355x-vllm` (8K/1K STP), `dsv4-fp4-mi355x-vllm-mtp` (8K/1K MTP), and `dsv4-fp4-mi355x-vllm-agentic-mtp` (AgentX) — to the first ROCm nightly that ships [vllm-project/vllm#53161](https://github.com/vllm-project/vllm/pull/53161) (merged as `de69e821`), which fuses the DeepSeek-V4 native-FP8 shared expert into the MXFP4 routed-expert AITER kernel (FHMoE).

## Why this is not a duplicate

- **[#2792](https://github.com/SemiAnalysisAI/InferenceX/pull/2792)** adds `VLLM_ROCM_USE_AITER_FUSION_SHARED_EXPERTS=1` to the two 8K/1K scripts and pins the Sep-1 `nightly-7c5dc571`. This PR deliberately does **not** touch those scripts — the export belongs there. It is materially different: it moves the image to a build where that export actually does something.
- **[#2841](https://github.com/SemiAnalysisAI/InferenceX/pull/2841)** bumps only the AgentX image, to `nightly-e962733e`, which also predates `de69e821`.

The env var alone was never sufficient. On every nightly before `de69e821`, vLLM's eligibility check rejected this mixed FP4+FP8 checkpoint and the fusion self-disabled at startup, which is exactly what #2792's changelog and the `vllm-project/recipes` DSv4-Pro caveat currently describe. #53161 is the piece that makes the flag take effect.

## Eligibility

vLLM gates the fused path on gfx950, TP8, DP1, no expert parallelism, no EPLB, `--moe-backend aiter`, BF16 dtype, and the DeepSeek-V4 block-128 FP8 shared-expert quant config. All three entries satisfy this. The AgentX DP-attention arm (TP1/DP8) does not and keeps the unfused path — no change in behaviour there.

## Blocking on

**The image tag is intentionally a `nightly-TBD` placeholder.** `de69e821` was the tip of vLLM `main` when this was written; the newest published nightly is `nightly-1970f3ed` (2026-09-06 05:31 UTC), which predates it. Before this leaves draft:

- [ ] Confirm the next nightly contains `de69e821` and replace `nightly-TBD` with the published sha (3 places in `configs/amd-master.yaml`, 1 note in `perf-changelog.yaml`)
- [ ] Land #2792 first (or absorb its two export lines here if it stalls)
- [ ] Run the 8K/1K STP + MTP sweeps and the AgentX sweep on the new image
- [ ] Confirm from the server log that the fused path is actually selected, not silently self-disabled
- [ ] Fill in `pr-link: .../PENDING` in `perf-changelog.yaml`

## Testing

No benchmark has been run on this image because it does not exist yet. Local validation only:

```
python3 utils/validate_perf_changelog.py --base-ref origin/main --head-ref HEAD
# Validated perf-changelog.yaml: final newline present and matrix generated
```

Supporting local measurements on a source build of the FHMoE branch (TP8 AgentX, MI355X, pure TP, synthetic AL 2.49) are promising but are **not** a validation of this image: c1 1,417, c4 2,628, c8 4,699, c32 14,178 tok/s/GPU, versus the reference run [32082871496](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/32082871496) at 1,143 / 2,251 / 4,112 / 12,939. That delta also carries three weeks of unrelated vLLM changes and should not be read as the FHMoE effect on its own.

> AI assistance was used to prepare this PR. Numbers, eligibility conditions, and duplicate checks were verified against the linked runs and source before submission.

### DIFF
--- a/configs/amd-master.yaml
+++ b/configs/amd-master.yaml
@@ -1202,7 +1202,7 @@ dsv4-fp4-mi355x-sglang-mtp:
 # gpu-mem-util=0.6. TP8 sweeps conc 4-64; DEP8 has a single conc=64
 # probe to validate the ROCm DP+EP path.
 dsv4-fp4-mi355x-vllm:
-  image: vllm/vllm-openai-rocm:nightly-09663abde0f50944a8d5ea30120666024b503faa
+  image: vllm/vllm-openai-rocm:nightly-de69e821b7c834bcf1ec96325ec8c410ac183680
   model: deepseek-ai/DeepSeek-V4-Pro
   model-prefix: dsv4
   runner: mi355x
@@ -1225,7 +1225,7 @@ dsv4-fp4-mi355x-vllm:
 # above ~conc32 (-37% @ conc32). Image reuses the base entry's v0.22.0 ROCm
 # build, which already contains the MTP commit.
 dsv4-fp4-mi355x-vllm-mtp:
-  image: vllm/vllm-openai-rocm:nightly-09663abde0f50944a8d5ea30120666024b503faa
+  image: vllm/vllm-openai-rocm:nightly-de69e821b7c834bcf1ec96325ec8c410ac183680
   model: deepseek-ai/DeepSeek-V4-Pro
   model-prefix: dsv4
   runner: mi355x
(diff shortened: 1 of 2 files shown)


Fangzhou-Ai:
Closing until the image exists. `de69e821` (vllm-project/vllm#53161) landed after the most recent ROCm nightly was cut, so the tag pinned here is not yet published and this repin cannot be validated.

Reopening once the next nightly builds and the real sha is known. Branch `fangzhou/dsv4-mi355x-fhmoe-nightly` is preserved; the only change needed is swapping the placeholder tag for the published one.

The documentation half of this work is tracked separately in vllm-project/recipes#932.

PR #3002, opened 2026-09-11
Revert "fix: size the GLM-5.2 MI355X eval HiCache pool absolutely (#2983)"

functionstackx:
Reverts #2983 (merge commit a5f23f4).

The HiCache pool sizing change was a workaround for a symptom, not the cause. The node runs one job at a time under `--exclusive`, so the shared-DRAM-budget explanation in #2983 does not hold, and it does not explain why only the eval path was crashing. The GSM8K crash is believed to be the SGLang EAGLE bug fixed upstream in sgl-project/sglang#38318, so the recipe change should not carry forward.

Discussed in #ext-amd-sw-engineering-semianalysis; Bryan Shan agreed to revert.

---

> [!NOTE]
> **Low Risk**
> Benchmark script-only change; restores prior HiCache CLI flags with no production runtime impact.
> 
> **Overview**
> Reverts the **eval-only HiCache pool workaround** from #2983 in the GLM-5.2 MI355X agentic benchmark launcher.
> 
> When `agentic_kv_offload_enabled` is on, **both** `hicache` and `mooncake` paths again always pass **`--hicache-ratio`** (with DP vs TP defaults unchanged) instead of switching to a fixed **`--hicache-size`** (16 GB default) when `EVAL_ONLY=true`. The Mooncake arm again sets **`--hicache-size 0`** alongside the ratio. Logging no longer uses `HICACHE_POOL_DESC` and always reports `ratio=$HICACHE_RATIO`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7c849424f2cb800f5f611d0aabb92ad42516715a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/benchmarks/single_node/agentic/glm5.2_fp4_mi355x_sglang_mtp.sh
+++ b/benchmarks/single_node/agentic/glm5.2_fp4_mi355x_sglang_mtp.sh
@@ -91,15 +91,6 @@ if agentic_kv_offload_enabled; then
         # env-var override for maximum throughput on nodes with >4 TB DRAM.
         HICACHE_RATIO="${HICACHE_RATIO:-1.5}"
     fi
-    # GSM8K never fills the agentic host pool; ratio 1.5 OOMs the TP4 DRAM share.
-    if [ "${EVAL_ONLY:-false}" = "true" ]; then
-        HICACHE_EVAL_SIZE_GB="${HICACHE_EVAL_SIZE_GB:-16}"
-        HICACHE_POOL_ARGS=(--hicache-size "$HICACHE_EVAL_SIZE_GB")
-        HICACHE_POOL_DESC="size=${HICACHE_EVAL_SIZE_GB} GB/rank (eval-only)"
-    else
-        HICACHE_POOL_ARGS=(--hicache-ratio "$HICACHE_RATIO" --hicache-size 0)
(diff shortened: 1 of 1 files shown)


jiejingzhangamd:
Agree on the scheduling point: launch_mi355x-amds.sh uses salloc ... --exclusive, so there is no co-tenancy here. The sentence in #2983 about another job owning the other 4 GPUs was wrong and I'll remove it. TOTAL_CPU_DRAM_GB=1199 is only a sizing budget, not evidence of co-tenancy.

I still think the failure itself is host OOM. EVAL_ONLY inherited --hicache-ratio 1.5; with a 127.90 GB/rank device pool that became 169.73 GB/rank pinned host memory, ~679 GB across TP4. The pool is registered up front but faulted in lazily, which matches the log: server ready → first lm_eval batch → SIGKILL.

Also, --exclusive is not the same as --mem=0; the MI355X launcher does not request --mem=0.

I don't think sgl-project/sglang#38318 explains this one: there was no kv_index_translator traceback, and both the failing and green runs used the same pre-#38318 image. The only relevant change was reducing the host pool; with --hicache-size 16, GSM8K completed at 0.9682 flexible / 0.9666 strict: https://github.com/SemiAnalysisAI/InferenceX/actions/runs/34535781749

PR #3058, opened 2026-09-12
perf: MI355X DSv4.1 Flash on the ROCm nightly image with the CK a8w4 MoE experts / MI355X DSv4.1 Flash 切换 ROCm nightly 镜像并选择 CK a8w4 MoE 专家内核

Fangzhou-Ai:
Supersedes #3035 and targets `main` directly, so the MI355X DSv4.1 Flash recipe lands in one self-consistent state. #3035 is frozen and its five commits are carried here unchanged (rebased onto current `main`); this PR adds the MoE backend change on top. #3035 can be closed once this merges.

Carried from #3035:

- `configs/amd-master.yaml`: image `vllm/vllm-openai-rocm:deepseekv41-flash-0909` → `nightly-eed1f3d0c6043bd494424a22443ee198dd56f657`, matching merged vllm-project/recipes#968
- `AITER_TRITON_LOG_LEVEL=ERROR`, which removed 98% of the server log (411k of 417k lines) with no throughput effect
- `MAX_NUM_SEQS` from `2 * CONC` to an explicit reproducibility cap of 128 (the effective `vllm serve` default on MI355X is 1024), which took queue times at CONC=1 from 16.2 s to zero

New here: the MoE backend change described below.

## Validation

- Full exact-image sweep and eval: https://github.com/SemiAnalysisAI/InferenceX/actions/runs/34710937012
- Corresponding upstream recipe merged: https://github.com/vllm-project/recipes/pull/968

## Summary

The recipe pinned `--moe-backend aiter_triton_mxfp4_bf16`. That name maps to exactly one entry, the Triton W4A16 kernel, so decode ran `_moe_gemm_a16w4`. Passing plain `aiter` opens vLLM's full priority list, whose head is the Composable Kernel backend, and the experts become the a8w4 family the DSV4-Pro MI355X recipe already uses:

```
mfma_moe1_silu_mul_afp8_wfp4_bf16_t32x128x256_pm1_async_gui_v33.kd
mfma_moe2_afp8_wfp4_bf16_cshuffle_t32x128x128_vscale_fix3_fp4opt_v1_pm1.kd
```

Two things made this non-obvious, so the flag carries a comment: the backend is named `..._BF16`, and this checkpoint declares `activation_scheme: dynamic` with no `input_scale` tensors. vLLM's explicit W4A8 path does require static FP8 activation scales and correctly refuses, but CK quantizes activations to FP8 inside the kernel and needs no checkpoint scales.

## Decode traces

[...]

PR #3035, opened 2026-09-12
config: switch MI355X DSv4.1 Flash to the ROCm nightly image / MI355X DSv4.1 Flash 切换到 ROCm nightly 镜像

Fangzhou-Ai:
Switch the MI355X DeepSeek-V4.1-Flash recipe from the `deepseekv41-flash-0909` release tag to the upstream ROCm nightly, so it picks up [vllm-project/vllm#56503](https://github.com/vllm-project/vllm/pull/56503) (merged 2026-09-12, commit `46d2b23`). That PR moves the mHC delayed pre block off the eager Torch reference and onto AITER; on this model the mHC block is 85% of the decoder's kernel launches, so the `0909` tag leaves most of the decode cost unaddressed.

**The nightly is now published**, so the earlier `nightly-TBD` placeholder is replaced with the real pin:

| | |
|---|---|
| tag | `vllm/vllm-openai-rocm:nightly-eed1f3d0c6043bd494424a22443ee198dd56f657` |
| digest | `sha256:960228cfcb5de9f4cd22d28998d1125be62b546c3d220a884f570343e99ffcee` |
| published | 2026-09-12T05:27:48Z |

`eed1f3d` is 11 commits ahead of, and 0 behind, #56503's merge commit, so the change is included. The upstream vLLM recipe is being moved to the same image in **[vllm-project/recipes#962](https://github.com/vllm-project/recipes/pull/962)**, which keeps `recipes.vllm.ai` and `amd-master.yaml` pointing at one tag.

**Still draft pending GPU validation on this exact image.** Rebased onto `main` now that #2962 has merged.

Measured on MI355X (8x gfx950, TP4, fp4, five-token DSpark) against the 2026-09-11 ROCm nightly with #56214 and #56503 applied, which is what the new image contains:

| conc 1 | `deepseekv41-flash-0909` | nightly + #56503 |
|---|---|---|
| inter-token latency p50 | 25.58 ms | **3.85 ms** |
| output tokens/s per user | 39.1 | **260.0** |
| aggregate output tokens/s | 35.9 | **132.9** |
| median cold-prefix TTFT | 10,236 ms | **1,440 ms** |
| gsm8k strict-match | 0.9704 | 0.9719 ± 0.0045 |

[...]

PR #946, opened 2026-03-25
[AMD] [Don't Merge] Update cli args qwen

zhentaocc:

PR #3068, opened 2026-09-13
perf: MI355X DSv4.1 Flash graph capture sized to concurrency, sweep extended to 128 / MI355X DSv4.1 Flash 按并发设定 graph capture 规模并将并发扫描扩展至 128

Fangzhou-Ai:
Supersedes #3058 and targets `main` directly, so the MI355X DSv4.1 Flash recipe lands in one self-consistent state. All of #3058's commits are carried here unchanged; this PR adds two on top. #3058 (and #3035 behind it) can be closed once this merges.

Carried from #3058:

- `configs/amd-master.yaml`: image `deepseekv41-flash-0909` → `nightly-eed1f3d0c6043bd494424a22443ee198dd56f657`, matching vllm-project/recipes#962
- `AITER_TRITON_LOG_LEVEL=ERROR`, which removed 98% of the server log with no throughput effect
- `MAX_NUM_SEQS` from `2 * CONC` to the upstream default of 128, which took queue times at CONC=1 from 16.2 s to zero
- `--moe-backend aiter` rather than `aiter_triton_mxfp4_bf16`, selecting the CK a8w4 experts

## 1. Size CUDA graph capture to the point being measured

`CAPTURE_SIZE` was derived from `MAX_NUM_SEQS`, which is a constant, so `--max-cudagraph-capture-size` was 1024 at every point in the sweep regardless of what was being measured. At CONC=1 the engine never sees a decode batch larger than 6, yet it still captured graphs out to 1024: 18.71 GiB of PIECEWISE plus 19.40 GiB of FULL capture memory per GPU, roughly 30% of what is left after weights, spent on shapes the workload cannot reach.

**This is hygiene, not a throughput fix, and it is not expected to move any published metric.** Reclaiming that memory grows the KV pool from 33.6M tokens, but KV was never the constraint at these points ([run 34710937012](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/34710937012) server logs):

| conc | 1 | 2 | 4 | 8 | 16 | 32 |
|---|---|---|---|---|---|---|
| peak GPU KV cache used | 1.6% | 2.2% | 2.7% | 3.3% | 5.9% | 13.0% |
| peak prefix cache hit rate | 100% | 100% | 100% | 100% | 99.9% | 99.7% |

Nothing was evicted and nothing refilled. Extrapolating the observed doubling, conc 128 would sit near half of even the old pool, so the ladder could have been extended without this change. The reasons to make it are that capture should describe the point being measured, that 38.1 GiB per GPU plus the startup time to capture 1024-wide shapes buys nothing, and that the headroom at the new points is then not something a reader has to reason about.

The derivation comes from the workload: each of the `CONC` lanes submits one real token plus `NUM_SPEC_TOKENS` drafts per step, so the largest decode batch is `CONC * (1 + NUM_SPEC_TOKENS)`, rounded up to a power of two.

| conc | 1 | 2 | 4 | 8 | 16 | 32 | 64 | 128 |
|---|---|---|---|---|---|---|---|---|
| capture size | 8 | 16 | 32 | 64 | 128 | 256 | 512 | 1024 |

This matches the derivation the other DSv4.1 Flash agentic recipes already use, so nothing here is novel to MI355X — the AMD arm was simply deriving it from the wrong variable.

## 2. Extend the ladder to 64 and 128

`conc-list` was `[1, 2, 4, 8, 16, 32]`, which stops before the curve does anything interesting. From the published sweep:

| conc | interactivity (tok/s/user) | output tok/s/GPU | total tok/s/GPU | mean ITL (ms) | mean TTFT (s) | J / output token |
|---|---|---|---|---|---|---|
| 1 | 256.5 | 33.4 | 4,542.9 | 3.90 | 0.75 | 11.91 |
| 2 | 239.8 | 39.9 | 4,892.6 | 4.17 | 0.57 | 10.53 |
| 4 | 220.9 | 52.9 | 7,339.4 | 4.53 | 0.58 | 8.60 |
| 8 | 191.0 | 94.4 | 12,584.0 | 5.24 | 0.57 | 5.60 |
| 16 | 127.1 | 159.6 | 23,392.8 | 7.87 | 0.70 | 3.81 |
| 32 | 72.4 | 259.1 | 40,919.4 | 13.80 | 0.90 | 2.60 |

At the top of the ladder throughput is still climbing steeply — output throughput per GPU rises 62% from conc 16 to 32 with no sign of a knee — and energy per output token is still improving, 3.81 → 2.60 J. TTFT is flat near 0.9 s and KV occupancy peaked at 13%, so nothing indicates the engine is near saturation.

There is a sharper reason than the shape of the curve. AgentX is a closed-loop harness: each of the `CONC` lanes issues a request, waits for it, then spends time on the client side of the agent turn, so aggregate throughput is the product of per-request output rate and how many requests are actually in flight. Computing average in-flight requests as `sum(e2el) / window` shows how much of the cap the sweep is using:

| conc | 1 | 2 | 4 | 8 | 16 | 32 |
|---|---|---|---|---|---|---|
| avg in-flight requests | 0.57 | 0.69 | 0.99 | 2.02 | 4.94 | 14.00 |
| share of the concurrency cap | 57% | 35% | 25% | 25% | 31% | 44% |

Up to conc 8 the engine is served by a quarter of the lanes, so the aggregate figure is set by trace demand rather than by the engine, and per-request latency is absorbed by lanes sitting idle between turns. Occupancy only starts climbing at 16 and 32. The points where lane occupancy approaches the cap — and where aggregate throughput therefore becomes a statement about the engine rather than about the harness — are above the current top of the ladder, which is exactly the region that determines serving cost.

`conc-list` becomes `[1, 2, 4, 8, 16, 32, 64, 128]`, the same ladder the other DSv4.1 Flash agentic recipes run.

## 3. Engram note corrected

The previous wording ("upstream currently rejects it on ROCm") was accurate but incomplete, and invited the reading that MI355X therefore keeps the Engram tables in HBM. It does not.

`EngramConfig.verify_model_config` gates on `current_platform.is_cuda()`, which is strictly `PlatformEnum.CUDA`, so passing `--engram-config` on ROCm raises `ValueError` and the engine refuses to start. But `VllmConfig._resolve_and_verify_engram_config` returns *before* validation when the flag is absent, and `models/deepseek_v4_1/common/engram.py` then defaults to `cpu_offload=True`. Every MI355X run already logs the result:

```
[engram.py:690] Engram table offloaded to pinned host memory: 96,002,514 rows x 256, 23.60 GiB per rank
```

Four ranks × two Engram layers, so ~188 GiB in pinned host memory. Omitting the flag is therefore not a downgrade, and the capture-memory figures in section 1 have nothing to do with Engram placement.

For the record, the underlying machinery is not CUDA-only: `is_uva_available()` requires only pinned memory and returns True here, and `get_accelerator_view_from_cpu_tensor()` already dispatches on `is_cuda_alike()`, which includes ROCm, into `torch.ops._C.get_cuda_view_from_cpu_tensor` — an op that is registered in this ROCm build. The upstream gate looks like a conservative whitelist rather than a capability limit.

## Not changed

`MAX_NUM_SEQS` stays pinned at 128, which is `SchedulerConfig.DEFAULT_MAX_NUM_SEQS`, so the scheduler cap is unchanged. `--gpu-memory-utilization` stays at 0.9 per upstream recipe #946. `dram-utilization` stays at 0.60 since it only takes effect for `kv-offloading: dram`.

## Validation

- `python -m pytest utils/matrix_logic/ -v`: 319 passed.
- `python -m infx.workflows.validate_perf_changelog` passes locally against `origin/main`.
- `bash -n` on the recipe passes; the capture derivation was executed for all eight concurrencies and produces the table in section 1.
- `configs/amd-master.yaml` parses and the resolved search space is `tp: 4, kv-offloading: none, spec-decoding: mtp` with the new `conc-list`.

Runtime evidence for the two new points is pending. Note that the MI355X runners have been failing independently of this recipe: legs die with a worker killed by signal during PIECEWISE capture (`Worker proc VllmWorker-N died unexpectedly (exit code: None)`) on both MoE backends and on four different nodes, while the identical serve command completes on a dedicated MI355X box. Capturing far fewer graph shapes shortens that window but is not a fix for it.

---

> [!NOTE]
> **Medium Risk**
> Changes vLLM serving flags, memory gates, and sweep concurrency on MI355X AgentX benchmarks; misconfiguration could cause init failures or non-comparable reruns, though scope is limited to one recipe key.
> 
> **Overview**
> Consolidates the MI355X **DeepSeek-V4.1-Flash** AgentX DSpark recipe (`dsv41flash-fp4-mi355x-vllm-agentic-dspark`) into one state: bumps the vLLM ROCm image to pinned **nightly** (for AITER mHC), switches **`--moe-backend`** from `aiter_triton_mxfp4_bf16` to **`aiter`**, sets **`MAX_NUM_SEQS=128`**, suppresses noisy AITER Triton logs, and calls **`wait_for_amd_gpu_clean 1`** before serve so prior-job VRAM does not abort init or skew KV pool sizing.
> 
> The main new behavior in this PR is **CUDA graph capture sized per sweep point**: `--max-cudagraph-capture-size` now scales as **`CONC * (1 + NUM_SPEC_TOKENS)`** (power of two) instead of staying at 1024 from `MAX_NUM_SEQS`—reclaiming capture memory for headroom without changing the intended throughput story at already-swept concurrencies. **`conc-list`** in `amd-master.yaml` extends to **64 and 128** so the AgentX ladder can reach engine-saturated serving cost. Docs (EN/ZH) and **`perf-changelog.yaml`** entries document MoE, capture, concurrency, and GPU-clean gates.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b10c107cf2babc692f272458a61e7b4745d005cf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chunfangamd:
Upstream recipe PR for the public serving side of this change: **https://github.com/vllm-project/recipes/pull/966**

It is an independent one-file change authored on current `vllm-project/recipes` main, aligned to this PR's head `b10c107cf2babc692f272458a61e7b4745d005cf`. It carries the parts of this PR that describe how a user serves the model:

- ROCm image pinned to `vllm/vllm-openai-rocm:nightly-eed1f3d0c6043bd494424a22443ee198dd56f657` (NVIDIA keeps the `deepseekv41-flash-0909` release tag)
- `--moe-backend aiter` in place of `--moe-backend aiter_triton_mxfp4_bf16`
- `AITER_TRITON_LOG_LEVEL=ERROR`
- CUDA graph capture sizing under DSpark documented as `next_power_of_two(6 x peak active decode sequences)`, with the 1–128 table, rather than a hard-coded global value — the recipe site has no runtime `CONC` input, so the derivation is what transfers
- the ROCm Engram note: `--engram-config` is CUDA-gated and must be omitted, and omitting it leaves the tables in pinned host memory rather than HBM

The harness-only mechanics stay here and are deliberately not upstreamed: trace-corpus pinning, result staging, the pre-launch GPU-memory gate, container mounts, the `conc-list` schema, synthetic acceptance settings, and the changelog entry.

I re-verified the vLLM-side claims against `vllm-project/vllm` at the pinned commit itself rather than restating them: the `aiter` candidate list and CK-first ordering in `fused_moe/oracle/mxfp4.py`, `EngramConfig.verify_model_config`'s `is_cuda()` gate, the early return in `_resolve_and_verify_engram_config` when the flag is absent, and `cpu_offload=True` defaulting in `ParallelEngramEmbedding` with the `engram.py:690` offload log. They all hold at `eed1f3d`.

**Runtime evidence is marked pending on the recipe PR.** The exact-head sweep here (run [34768550843](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/34768550843)) is still queued, so the recipe PR makes no `mi355x: verified` claim and does not add the badge. Once that run completes I will attach the run URL, head SHA, and concurrency coverage to the recipe PR.

Feel free to add the link to the description if you want it there — I did not want to edit your PR body.

Fangzhou-Ai:
Closing this PR because the per-concurrency CUDA graph bound is invalid for AgentX. `CONC` counts top-level lanes, while subagent fan-out produced up to four simultaneous model requests. With five DSpark draft tokens, the c1/c2 limits of 8/16 tokens sent larger decode batches through `CUDAGraphMode.NONE`, causing the measured tail-latency regression.

I will replace this with an append-only PR based on #3058 that keeps the 1024-token capture ceiling and only adds c64/c128 to the sweep.

<details><summary>中文</summary>

关闭此 PR，因为按并发数计算 CUDA graph 上限并不适用于 AgentX。`CONC` 表示顶层 lane 数量，但子代理扇出使同时进行的模型请求最多达到 4 个。DSpark 使用 5 个 draft token，因此 c1/c2 的 8/16-token 上限会让较大的 decode batch 进入 `CUDAGraphMode.NONE`，从而造成已测得的尾延迟回退。

后续将提交一个基于 #3058 的 append-only PR，保留 1024-token capture 上限，只把 c64/c128 加入 sweep。

</details>

PR #3120, opened 2026-09-15
[AMD][MI35X] 0915 DSV4 sglang mtp agentic

1am9trash:
Successful run:
- https://inferencex.semianalysis.com/inference?unofficialRun=35014953307
- https://inferencex.semianalysis.com/evaluation?unofficialRun=35014953307

Summary:
- Bump the image to `lmsysorg/sglang-rocm:v0.5.19-rocm720-mi35x-20260914`.
- Raise `--prefill-decode-interval` from 10 to 20 on both the TP-only and DP-attention serving paths, so the scheduler runs more decode steps between prefill admissions.
- Switch the sglang-router policy for the DP-attention arms from `consistent_hashing` to `cache_aware`.


1am9trash:
image

yctseng0211 (approved):
- [x] Verified that this PR has passed PR validation. Please link to GitHub Action workflow that shows this. https://inferencex.semianalysis.com/inference?unofficialRun=35014953307
- [x] Verified that this PR passes evals.  Please link to GitHub Action workflow that shows this. https://inferencex.semianalysis.com/evaluation?unofficialRun=35014953307

- insert any additional info here
recipe: https://github.com/sgl-project/sglang/pull/39702
Signed: @yctseng0211

1am9trash:
Hi @cquil11 @adibarra, 
May you help review this update for deepseek-v4 mxfp4 agent mode on MI355? 
Many thanks.

adibarra (approved):

PR #3167, opened 2026-09-16
fix(amd): collect logs from every Slurm node / 修复 AMD Slurm 多节点日志收集

nvpohanh:
[by Codex]

## Summary

- collect server logs from every AMD Slurm node after the primary server step exits
- merge node-local prefill, decode, router, client, and wrapper logs into the shared log directory before artifact packaging
- restore runner ownership after the fan-in so later jobs can clean the workspace
- add behavioral coverage for merging separate prefill/decode node log trees
- document the AMD multinode diagnostic-artifact contract in English and Chinese

## Root cause

In run 34926284365, job 104504413449, the uploaded multinode server-log archive contained the rank-0 prefill/frontend logs but no decode log. `/run_logs` is backed by each compute node's local `/tmp`, while the engine launcher copied only rank 0's tree into shared storage. The workflow correctly uploaded the resulting tarball, but the decode node's tree had never reached it.

This change performs a second one-task-per-node Slurm staging pass after every server container and outer `tee` has exited. Each node merges its completed `/tmp/slurm_job-${SLURM_JOB_ID}` tree into the shared job directory, covering SGLang, vLLM, and Atom through the common AMD launcher.

## Validation

- `bash -n benchmarks/multi_node/amd_utils/job.slurm benchmarks/multi_node/amd_utils/stage_node_logs.sh`
- `python -m pytest utils/test_amd_node_log_staging.py -q` — 2 passed
- `git diff --check`

No `perf-changelog.yaml` entry is included because the new staging pass runs only after the measured server step has completed and changes diagnostic artifact collection, not benchmark execution or performance.

Fixes the missing decode-log artifact observed in https://github.com/SemiAnalysisAI/InferenceX/actions/runs/34926284365/job/104504413449.

<details><summary>中文</summary>

## 摘要

- 在主要服务器步骤结束后，从每个 AMD Slurm 节点收集服务器日志
- 在打包工件之前，将各节点本地的 Prefill、Decode、Router、Client 和包装器日志合并到共享日志目录
- 汇总完成后恢复 Runner 用户所有权，确保后续作业可以正常清理工作区
- 添加行为测试，覆盖独立 Prefill/Decode 节点日志树的合并
- 同步更新中英文 AMD 多节点诊断工件契约文档

## 根因

在 Run 34926284365 的 Job 104504413449 中，上传的多节点服务器日志工件包含 Rank 0 的 Prefill/Frontend 日志，但缺少 Decode 日志。`/run_logs` 实际映射到每个计算节点本地的 `/tmp`，而引擎启动器只把 Rank 0 的日志树复制到共享存储。工作流正确上传了生成的压缩包，但 Decode 节点的日志树从未进入共享目录。

本变更会在所有服务器容器及其外层 `tee` 退出后，再执行一次每节点一个任务的 Slurm 日志汇总。每个节点都会把完整的 `/tmp/slurm_job-${SLURM_JOB_ID}` 日志树合并到共享作业目录。该修复位于 AMD 公共启动路径，因此同时覆盖 SGLang、vLLM 和 Atom。

## 验证

- `bash -n benchmarks/multi_node/amd_utils/job.slurm benchmarks/multi_node/amd_utils/stage_node_logs.sh`
- `python -m pytest utils/test_amd_node_log_staging.py -q` — 2 passed
- `git diff --check`

本 PR 未添加 `perf-changelog.yaml` 条目，因为新增汇总步骤只在被测服务器步骤完成后运行，仅改变诊断工件收集，不影响基准执行或性能。

</details>


nvpohanh:
[by Codex]

The analogous per-node staging change is not needed for the current NVIDIA GPU pipelines. Their logging paths already write each role directly into storage shared with the runner:

- The standard `srt-slurm` path assigns every prefill/decode worker a node-qualified file under the common `runtime.log_dir` (`<node>_<mode>_w<index>.out`), and the launcher archives that entire log directory.
- The NVIDIA llm-d path writes each rank’s vLLM/sidecar logs directly under the shared `/benchmark_logs` mount.
- Native TileRT writes both `tilert_prefill.log` and `tilert_decode.log` directly to `BENCHMARK_LOGS_DIR`.

The AMD path fixed by this PR was different: every node wrote under its own node-local `/tmp` (mounted as `/run_logs`), but only rank 0 copied logs back to shared `BENCHMARK_LOGS_DIR`. Therefore decode-node logs never reached the artifact. The fan-in helper in this PR should remain AMD-specific rather than being applied to the NVIDIA launchers.

PR #3169, opened 2026-09-16
feat(agentx): retune Kimi-K3 FP4 MI355X ATOM DSpark recipe

gbyu-amd:
Track the checked-in ATOM recipe [`recipes/Agentic-Kimi-K3.md`](https://github.com/ROCm/ATOM/blob/k3_agentic_recipe/recipes/Agentic-Kimi-K3.md) at ROCm/ATOM#2131, retuned per concurrency on image `kimi_k3_agentic_0911`.

Publish the concurrency set **[1, 2, 4, 8, 16, 32, 40, 48, 56, 64, 72]** in place of [1, 2, 4, 8, 12, 14, 16, 32, 40, 56, 64]. Concurrency 48 and 72 are new points, 12 and 14 are dropped, and every surviving point is retuned.

- `gpu-memory-utilization` is 0.90 and `max-num-batched-tokens` is 8192 on every point, so the launcher sets both once instead of per band.
- `max-num-seqs` is 32 through concurrency 16 and `2 * CONC` from 32 up.
- `AITER_REUSE_IDENTICAL_COMM_GROUPS` becomes a threshold: 1 at concurrency 56, 64, and 72, 0 from 1 through 48.
- The two paged-KV `dram-utilization` blocks carry onto the new set, still divided by TP for the per-rank `LMCACHE_MAX_LOCAL_CPU_SIZE`: 0.343 lands 128 GB/rank for concurrency 8 through 48, 0.513 lands 192 GB/rank for 56 through 72.

The concurrency-14 CUDA-graph pin goes with the point, and was already dead code: it set `CUDAGRAPH_MAX_NUM_SEQS=32` inside the `case` block, which the capture-width derivation overwrote with `2 * CONC` two statements later, so it never reached the server. Capture behaviour is unchanged — `--level 3` with `--cudagraph-mode FULL` over `[2 .. 2 * CONC * (1 + draft tokens)]` — and reproduces the recipe's `graph_max` column at all eleven points:

| CONC | 1 | 2 | 4 | 8 | 16 | 32 | 40 | 48 | 56 | 64 | 72 |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| `max-num-seqs` | 32 | 32 | 32 | 32 | 32 | 64 | 80 | 96 | 112 | 128 | 144 |
| draft tokens | 7 | 7 | 7 | 3 | 3 | 0 | 0 | 0 | 0 | 0 | 0 |
| `graph_max` | 16 | 32 | 64 | 64 | 128 | 64 | 80 | 96 | 112 | 128 | 144 |

> [!NOTE]
> Opened from a fork (`gbyu-amd/InferenceX`) because `gbyu-amd` has read-only access here, which also blocks adding the `AMD`, `agentx`, and `full-sweep-enabled` labels. If the sweep dispatch does not reach fork branches, a maintainer can re-push the branch upstream and close this PR in favour of that one.


functionstackx:
there is 2 duplicate ATOM Kimi K3 prs at the same time, can we just have 1 reopenned plz?

PR #2823, opened 2026-09-04
[AMD][DSV4][Disagg] Enable UMBP, DP-attention tuning, FP4 sparse-attention indexer optimizing

ichbinblau:
## Summary

Tunes the `dsv4-fp4-mi355x-sglang-disagg-agentic-hicache-mtp` (DeepSeek-V4-Pro-AgentX, MI355X, disaggregated, HiCache DRAM offload) recipe:

### DP-attention rework
- Gate DP-attention-only SGLang knobs (`SGLANG_SHARED_EXPERT_TP1`, `SGLANG_DP_SHARED_EXPERT_LOCAL`, `SGLANG_DP_USE_GATHERV`, `SGLANG_DP_USE_REDUCE_SCATTER`, wider `GPU_MAX_HW_QUEUES`) behind `PREFILL_ENABLE_DP`/`DECODE_ENABLE_DP` in `env.sh`, instead of exporting them unconditionally, so TP-only arms keep `GPU_MAX_HW_QUEUES=2`. Add `SGLANG_MORI_RECV_BOUND=1` for the disagg MoRI path.
- Round out `dp_flags` in `models.yaml` (`--tokenizer-worker-num 8`, `--enable-dp-attention-local-control-broadcast`, `--stream-interval 20`, , `enable_two_batch_overlap`) to mirror the single-node DP-attention recipe.
- Switch the active DP-attention search-space arm's decode side from `ep:8` to `ep:1` (DP-attention over TP-sharded MoE instead of expert-parallel dispatch).
- Widen `decode.dp.cuda_graph_bs_range` from `1-128` to `1-256` and the DP-attention arm's `conc-list` back to the full `[64, 96, 128, 192, 256]` sweep.

### FP4 sparse-attention indexer
- Bump the image from `lmsysorg/sglang-rocm:v0.5.18-rocm720-mi35x-20260831` to `...202609011`, the first tag carrying gfx95 (MI350X/MI355X) support for DeepSeek-V4's FP4 sparse-attention indexer via AITER kernels ([sglang PR #37353](https://github.com/sgl-project/sglang/pull/37353), merged 2026-09-02). Previously the flag was gated to NVIDIA SM100/SM120 only and hard-failed on AMD.
- Add `--enable-deepseek-v4-fp4-indexer` to `DeepSeek-V4-Pro-AgentX` `base_flags`. 
- Lower prefill/decode `mem_fraction_static` 0.90 -> 0.86 in DP case.

[...]

PR #3170, opened 2026-09-16
fix(amd): restore native MTP checkpoint and collect all node logs

cquil11:
The MI355X disaggregated recipe pairs Pro-0813 with the legacy EAGLE/MTP draft architecture, which logs missing required draft weights. Switch its model back to `deepseek-ai/DeepSeek-V4-Pro` so the existing EAGLE path uses the original checkpoint’s native MTP weights. The existing selector restores thinking-on golden AL **2.49** for three draft tokens; evaluations retain real acceptance.

Also collect server logs from every Slurm node. `/run_logs` is node-local, and the old launcher copied only node 0’s tree, leaving decode logs out of GitHub artifacts. After the server step and container cleanup, run one collection task per node into the shared directory used by the existing artifact packager. Preserve server failures, fail otherwise-successful jobs if collection fails, and restore runner ownership. This adapts the fix proposed in #3167.

Explicitly pass the existing SGLang router port into evaluation. The shared library now requires `PORT` before parsing `--port`; without this caller input, evaluation exits after successful server startup. The endpoint remains port 30000. Forward the workflow-owned `AIPERF_EXPERIMENTAL_FAST` and `REQUIRE_POWER` values through the existing container environment allowlist; both are explicitly `0` for this sweep. This restores the intended client inputs without introducing defaults or changing validation.

The serving image and seven-point sweep remain unchanged. A perf changelog entry and `full-sweep-enabled` request fresh results.

Validation:
- All seven expected matrix points use the original checkpoint, MTP, and two nodes.
- Cluster inspection confirms all 64 original checkpoint shards and the required native MTP tensor names are present.
- Behavioral node-log staging tests: 2 passed, including separate prefill/decode trees and a missing-source failure.
- Real eval-library reproduction confirms the port fix uses the existing router endpoint.
- Bash syntax, Ruff lint/format, perf changelog/matrix validation, and `git diff --check` pass; historical changelog bytes are preserved.

Runtime checks confirmed complete native MTP loading on both roles, golden AL 2.49 in the throughput command, and both prefill/decode logs in an uploaded server artifact. GPU qualification is still in progress after fixing the missing client inputs. Confirm complete draft weight loading, real acceptance during evaluation, and both roles in uploaded server-log archives before treating the rerun as validated.

Follow-up to #2823.


cquil11:
Closing as superseded per maintainer direction: use #3188 (Pro-0813 with DSPARK) as the serving baseline for the srt-slurm port in #2555. This does not claim successful E2E validation.

PR #3179, opened 2026-09-16
Fail closed when collecting workflow statistics / 运行统计收集异常时明确失败

adibarra:
Workflow statistics currently catch authentication failures, exit successfully, and write `null` to `run_stats.json`. Collect job pages through the existing `infx.github` client instead, and fail before writing statistics if a request or response is invalid.

- Preserve successful JSON output, hardware matching, zero-count buckets, and counts across every retry (`filter=all`). Skipped and unrelated jobs remain excluded; cancelled and unfinished hardware jobs still count as attempts.
- Reject malformed pagination responses instead of treating them as empty results. Pagination owns `page` and `per_page`, while retaining caller filters.
- Load configuration when collecting statistics rather than during import, so installed hardware-matching helpers work outside a checkout.
- Remove PyGithub and four unused transitive dependencies. The CLI/module and legacy stub remain available. Redundant identity/repository/run preflight calls and their diagnostic messages are removed; the statistics table is unchanged.

Validation: 154 focused behavioral tests passed in 1.51s, including HTTP failures, multi-page retry counts, preservation of existing output after a later-page failure, installed-wheel behavior, reuse and sign-off consumers. A temporary comparison of the actual previous and new implementations covered 729 combinations of job conclusions with identical successful outputs; its scratch files were removed. Ruff check and format passed. CI passed all 1,558 tests in 28.97s, lint, MCP compatibility, and sign-off scope.

Follow-up to #3168. No benchmark, recipe, workflow YAML, or permission-policy changes.

<details>
<summary>中文</summary>

当前运行统计在身份验证失败后仍以成功状态退出，并将 `null` 写入 `run_stats.json`。本 PR 改为通过现有的 `infx.github` 客户端读取 Job 分页；请求失败或响应无效时，在写入统计文件之前报错。

- 保留成功时的 JSON 输出、硬件匹配、零计数分组，以及所有重试的统计（`filter=all`）。跳过和无关的 Job 仍不计入；取消和未完成的硬件 Job 仍计为一次尝试。
- 拒绝格式错误的分页响应，避免将其当作空结果。分页函数自行控制 `page` 和 `per_page`，同时保留调用方的筛选条件。
- 将配置读取推迟到收集统计时，使安装后的硬件匹配函数可以在仓库目录之外使用。
- 移除 PyGithub 及四个不再使用的间接依赖。保留模块 CLI 和旧路径入口，移除冗余的身份、仓库和运行预检请求及其诊断消息；统计表格不变。

验证：154 项针对性行为测试在 1.51 秒内通过，覆盖 HTTP 错误、多页重试计数、后续分页失败时保留原有输出、安装包行为，以及复用和签核调用方。临时对照运行了新旧真实实现的 729 种 Job 结论组合，成功输出完全一致，临时文件已清理。Ruff 检查和格式检查通过，CI 的 1,558 项测试在 28.97 秒内全部通过，Lint、MCP 兼容性和签核范围检查也通过。

这是 #3168 的后续改进，不修改基准测试、配方、工作流 YAML 或权限策略。

</details>


adibarra:
@claude Please review this follow-up, especially successful-output compatibility, retry pagination, failure handling, and test quality. Review only; do not modify the branch or dispatch benchmarks.

<details>
<summary>中文</summary>

请审阅此后续 PR，重点检查成功输出的兼容性、重试分页、失败处理和测试质量。仅进行审阅，不修改分支，也不触发基准测试。

</details>

Klaud-Cold:
**Claude finished @adibarra's task in 3m 16s** —— [View job](https://github.com/SemiAnalysisAI/InferenceX/actions/runs/35074543283)

---
### Review of PR #3179

- [x] Gather context (diff vs `origin/main`, changed files, AGENTS.md)
- [x] Review successful-output compatibility
- [x] Review retry pagination (`filter=all`, page/per_page handling)
- [x] Review failure handling (fail-closed behavior)
- [x] Review test quality
- [x] Run tests and ruff locally
- [x] Post review

**Verdict: looks good to merge.** No blocking findings. Two optional nits below.

Local verification on this branch:

| Check | Result |
|---|---|
| `pytest` on the four touched test files | 49 passed |
| `uvx ruff check infx` | All checks passed |
| `uvx ruff format --check infx` | 86 files already formatted |
| Imports of `github`, `requests`, `urllib3`, `nacl`, `charset_normalizer` in any `.py` | none |

### Successful-output compatibility

The JSON contract is preserved. Both old and new code emit one entry per label from `load_hardware_labels()`, which is already sorted, so key order in the `indent=2` dump is unchanged. Zero-count buckets are still emitted. The counting rule is identical: `skipped` is excluded, `success` increments both counters, and every other conclusion including `null` for unfinished jobs increments only `total` (`infx/workflows/calc_success_rate.py:73-78`). The table printer is untouched.

One behavior change worth naming as a positive: the old path called `GET /user` first, which fine-grained PATs without user scope reject. That preflight was the only error the old code swallowed into `null`. The new path needs only repository `actions:read`, so a token that used to produce `null` will now either succeed or fail loudly.

### Retry pagination

- `filter=all` is forwarded through `params` and `paginate` now owns `page` and `per_page` by placing them after the spread (`infx/github.py:57`). The updated test in `utils/test_github.py:71-80` proves a caller-supplied `page=9, per_page=1` cannot shrink or skip pages.
- Termination on `len(items) < 100` is correct for the jobs endpoint, whose maximum page size is 100. An exact multiple of 100 costs one extra empty request, which is fine.
- The stricter shape check now rejects a dict without the item key, a non-list value, and non-dict items. Previously a dict lacking `jobs` was silently treated as an empty final page, which would have produced all-zero stats with exit code 0. This is the most important correctness fix in the PR.
- The reactions caller passes `item_key=""` and receives a top-level list, so it is unaffected by the stricter dict handling.

### Failure handling

Every failure path raises before the output file is opened in `main()`, so a stale `run_stats.json` is never overwritten and a fresh one is never created:

- HTTP errors become `RuntimeError` in `api()`. Network-level `URLError` and timeouts propagate uncaught, which also exits non-zero.
- Missing env vars raise `KeyError`; a non-integer run id raises `ValueError` before any request. Casting to `int` also keeps the run id out of the URL path unsanitized.
- Malformed job objects raise `KeyError` on `name` or `conclusion`.

Workflow impact: the `calc-success-rate` job in `run-sweep.yml:901` and `e2e-tests.yml:608` runs with `if: always()` and no `continue-on-error`, so these errors turn the job red and skip the `run-stats` upload. Downstream, `validate_run_stats` in `infx/workflows/validate_reusable_sweep_artifacts.py:584` requires that artifact when fixed-sequence rows exist. That is the intended fail-closed contract and I agree with it. Note that transient 5xx mid-pagination already failed the old PyGithub iteration, so this PR does not make the job flakier than before. It only stops the auth-failure case from masquerading as success.

### Test quality

All new tests exercise the real `main()` or `paginate()` with only `urllib.request.urlopen` stubbed, and assert on written artifacts, stdout, or raised errors with hand-computed expectations. They satisfy the AGENTS.md test rules.

- `test_success_rates_include_all_pages_and_retries` has teeth on three axes at once: the fake returns one fewer job unless `filter=all` is present, 101 jobs force a second page, and the request path is asserted inside the stub. Expected values (1/2 and 1/3) check out by hand.
- `test_failed_stats_do_not_publish_an_artifact` covers four distinct error branches without duplication.
- `test_later_page_failure_preserves_previous_artifact` is the key fail-closed test, since it distinguishes "write after all pages" from "write as you go."
- `test_installed_hardware_matching_works_without_repository_files` would fail on `main`, where import-time `load_hardware_labels()` hits `FileNotFoundError` from a tmp cwd. Good regression guard for the deferred config load.
- `test_run_statistics_reject_invalid_run_id_without_publishing` runs both the module and legacy entrypoints via subprocess and checks stderr plus artifact absence.

Minor observation, not blocking: `test_empty_job_list_still_writes_zero_counts` overlaps somewhat with the `unused` zero bucket already asserted in the first test. It does cover the distinct "empty first page terminates and still writes" path, so I would keep it.

### Nits (optional)

1. `print_success_rates` still accepts `None` and has a "No data to display" branch (`infx/workflows/calc_success_rate.py:85-89`). Since `calculate_hardware_success_rates()` can no longer return `None`, that branch is dead and the `| None` in the signature overstates the contract. [Fix this →](https://claude.ai/code?q=On%20branch%20fix%2Frun-stats-fail-closed%2C%20in%20infx%2Fworkflows%2Fcalc_success_rate.py%2C%20tighten%20print_success_rates%20to%20accept%20dict%5Bstr%2C%20dict%5Bstr%2C%20int%5D%5D%20only%20and%20remove%20the%20dead%20%22if%20success_rates%20is%20None%22%20branch%2C%20since%20calculate_hardware_success_rates%20can%20no%20longer%20return%20None.%20Run%20uvx%20ruff%20check%20infx%20and%20pytest%20utils%2Ftest_calc_success_rate.py.&repo=SemiAnalysisAI/InferenceX)
2. In `calculate_hardware_success_rates`, `load_hardware_labels()` runs before the env-var reads, so a missing `GITHUB_RUN_ID` outside a checkout surfaces as `FileNotFoundError` rather than `KeyError`. Harmless in CI. Reading env first would give a clearer error for local misuse.

Dependency removal checks out: the lockfile drops `pygithub` plus its transitive `pynacl`, `requests`, `urllib3`, and `charset-normalizer`, and nothing in the repo imports any of them. `uv run --locked` in the workflows will catch any lock drift. English and Chinese docs were updated together.

<details>
<summary>中文</summary>

**结论：可以合并。** 无阻塞问题，两条可选的小建议。

- **成功输出兼容性**：JSON 结构、键顺序、零计数分组和计数规则与旧实现一致。旧实现只有 `GET /user` 预检失败会被吞掉并写入 `null`，新实现仅需 `actions:read` 权限，因此以前输出 `null` 的令牌现在要么成功，要么明确报错。
- **重试分页**：`filter=all` 正确传递；分页函数强制控制 `page` 和 `per_page`，测试证明调用方无法覆盖。响应形状检查更严格，避免了缺少 `jobs` 键时静默写出全零统计。
- **失败处理**：所有错误都在打开输出文件之前抛出，不会覆盖或创建 `run_stats.json`。`calc-success-rate` Job 没有 `continue-on-error`，因此会变红并跳过上传，这是预期的 fail-closed 行为。
- **测试质量**：所有新测试都运行真实实现、仅模拟 `urlopen`、断言写入的文件或抛出的错误，符合 AGENTS.md 要求。
- **可选建议**：`print_success_rates` 的 `None` 分支已成死代码；`load_hardware_labels()` 先于环境变量读取执行，本地误用时错误信息不够直观。

本地验证：49 项测试通过，`ruff check` 和 `ruff format --check` 通过，仓库中无任何文件导入已移除的依赖。

</details>
 · Branch: [`fix/run-stats-fail-closed`](https://github.com/SemiAnalysisAI/InferenceX/tree/fix/run-stats-fail-closed)

PR #3196, opened 2026-09-16
[MI355x][Kimi-K3] config change to get the ROCM_AITER_MLA on the Draft model and other optimizations

ppalanga:
We are updating the KK3 performance to (1) reflect the optimizations in the VLLM nightly and (2) update the config for 

HW: MI355x
Platform: VLLM
Model: Kimi-K3
Benchmark: Single Node AgentX

Thanks


ppalanga:
not needed anymore

PR #3369, opened 2026-09-22
perf(agentx): Kimi K3 MI355X -- VLLM Image update

ppalanga:
Updating the Kimi-K3 MI355x VLLM AgentX image to the latest nightly. This contains additional optimizations that we merged to VLLM and AITER upgrade


functionstackx:
@chunfangamd @ppalanga can u plz open it in upstream branch plz

